PR #517, opened 2018-03-05
Add account initialization - Closes #284

### LINKED ISSUE #284: Implement account initialization
### Expected behaviour
It should be something like the designs bellow, but inside of the narrow transfer module, no in the wide ones.

https://drive.google.com/drive/folders/1yb9uwukDrmWzvw1pyATO1hevH5v_2WR6
08_address_initialization-start-dashboard-account-initialization-1-lisk-l
08_address_initialization-wallet-account-initialization-1-l

ginacontrino:
### What was the problem?
We needed an account initialisation

### How did I fix it?
we had none

### How to test it?
Log into an account with only incoming transactions -> you should see the account init option in the wallet and dashboard, then init your account -> the notice should be gone after

### What it looks like
screen shot 2018-03-15 at 10 47 12

screen shot 2018-03-13 at 10 47 54
screen shot 2018-03-13 at 10 48 21

### Review checklist
- The PR solves #284


src/components/accountInitialization/index.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  render() {
+    const { account, t, nextStep } = this.props;
+
+    return (<Box className={`${styles.wrapper} ${grid.row}`}>
+      <header>
+        <h2>{t('Initialize Lisk ID')}</h2>
+        <p>{t('It is recommended that you initialize your Lisk ID.')}</p>
+        <p>{t('The easiest way to do this is to send LSK to yourself by clicking this button. ' +
```
slaweet: I would delete `by clicking this button` because unlike in the original design, there are two buttons now.

src/components/accountInitialization/index.js
```diff
@@ -0,0 +1,66 @@
...
+            'It will cost you only the usual 0.1 LSK transaction fee.')}</p>
```
slaweet: Splitting `{t('...')}` into two lines is not supported by the parser script that generates `i18n/locales/en/common.json`

src/components/accountInitialization/index.js
```diff
@@ -0,0 +1,66 @@
...
+        <p>
+          <a target='_blank' href='https://help.lisk.io/account-security/should-i-initialize-my-lisk-account' rel='noopener noreferrer'>
+            {t('Why should I initialize my Lisk ID?')} <FontIcon>arrow-right</FontIcon>
```
slaweet: The question mark next to the arrow-right looks weird. It could be solved by reformulating: `Learn more about Lisk ID initialization`

src/components/accountInitialization/index.js
```diff
@@ -0,0 +1,66 @@
...
+          </a>
+        </p>
+      </header>
+      <div></div>
```
slaweet: The middle element should IMO contain all `<p>` elements with text that are currently in `<header>`, so that the text is in the middle.

src/components/accountInitialization/accountInit.css
```diff
@@ -0,0 +1,99 @@
...
+    text-transform: none;
+    min-width: 0;
+    width: 100%;
+  }
+}
+
+@media (--xLarge-viewport) {
+  header {
+    & h2 {
+      font-size: var(--font-size-h2-xl);
```
slaweet: Redefining `h2` `font-size`s shouldn't be necessary, these should be global and same in all components

test/integration/wallet.test.js
```diff
@@ -192,6 +193,35 @@ describe('@integration: Wallet', () => {
       step('When I click "send button"', () => helper.clickOnElement('.send-button button'));
       step(`Then I should see text ${successMessage} in "result box message" element`, () => helper.haveTextOf('.result-box-message', successMessage));
     });
+
+    describe('Scenario: should show account initialisation option if no public key and balance is greater than 0', () => {
+      step('Given I\'m on "wallet" as "genesis" account', () => setupStep('genesis', false, false));
```
slaweet: This would be more readable as `setupStep('genesis', { isLocked = false, withPublicKey = true })` 

Or even better, instead of `withPublicKey = false`, create a new  `uninitialized account` in `/test/constants/accounts.js` for this purpose.

test/integration/wallet.test.js
```diff
@@ -192,6 +193,35 @@ describe('@integration: Wallet', () => {
       step('When I click "send button"', () => helper.clickOnElement('.send-button button'));
       step(`Then I should see text ${successMessage} in "result box message" element`, () => helper.haveTextOf('.result-box-message', successMessage));
     });
+
+    describe('Scenario: should show account initialisation option if no public key and balance is greater than 0', () => {
```
slaweet: At least one test should go through the whole process of sending the "Initialization" transaction  (but maybe it can be e2e test)

test/utils/genericStepDefinition.js
```diff
@@ -40,7 +40,15 @@ export default class GenericStepDefinition {
...
-    expect(this.wrapper.find(query)).to.have.text(text);
+    expect(this.wrapper.find(query).first()).to.have.text(text);
+  }
+  /**
+   *
+   * @param {String} query - dom query that we need to check text of that
+   * @param {String} text - expect text of the dom query entry
+   */
+  haveInputValueOf(query, text) {
+    expect(this.wrapper.find(query).first().props().value).to.equal(text);
```
slaweet: https://github.com/producthunt/chai-enzyme#valuestr can be used here to have better error messages when the expectation is not met.

src/components/accountInitialization/accountInit.css
ginacontrino: I agree, it should be like that, but it's not right now, so I will create a ticket to clean up that mess

test/integration/wallet.test.js
ginacontrino: 👍

test/utils/genericStepDefinition.js
ginacontrino: Cool, didn't know this works

src/components/accountInitialization/index.js
ginacontrino: 👍

PR #522, opened 2018-03-06
Generate Index.html with webpack - Closes #514

### LINKED ISSUE #514: Generate index.html with webpack
### Expected behaviour
Webpack should generate index.html to be able to:
- Add a comment with Lisk Hub version number (So that DevOps can check they deployed the right thing and the old one is not still cached)
- Use hash in bundle names, so that we can cache forever.
- inject splash screen css that is currently hardcoded in there.

### Actual behaviour
index.html is static

faival:
### What was the problem?
- dev-ops requires to check built version deployed within the html of the application
- bundles at the moment do not contain hash strings (cached forever) 
- automate injection of inline css on head of document

### How did I fix it?

- added meta tag in head of document displaying current package.json version:
- bundles now append the hash used to generate them (only for production build)
- head css has been moved to a separate file, (now it's stylelinted, and fixed)

### How to test it?
- verify head of document contains current version in package.json
```<meta data-lisk-hub-version="0.2.0-rc.1">```
- verify scripts inclusion in body contain the hash in the referenced bundle
```
<link rel="stylesheet" href="./styles.c99be3303f67c17c6c39.css">
<script src="./bundle.vendor.c99be3303f67c17c6c39.js"></script>
<script src="./bundle.app.c99be3303f67c17c6c39.js"></script>
```
- verify there's a style tag in head containing the styles of src/assets/css/styles.head.css and those are minified
```
<style type="text/css">
          body{background-color:#013165}.splash-screen{
...
</style>
```

### Review checklist
- The PR solves #514  


faival:
`dev` build goal should also inject styles in head.

PR #537, opened 2018-03-08
Post Slack notifications to lisk-ci-hub channel

ghost:
### What was the problem?

Slack notifications were being posted to the old channel instead of `lisk-ci-hub`

### How did I fix it?

The logic to generate the channel name was changed: now the job base name name is taken and `lisk-` is replaced with `lisk-ci-`.

### How to test it?

Build this PR in Jenkins.


slaweet (approved):
👍

PR #542, opened 2018-03-09
Refactor route definition - Closes #499

### LINKED ISSUE #499: Refactor route definitions into one place
### Expected behaviour
All routes should be defined in https://github.com/LiskHQ/lisk-hub/blob/development/src/constants/routes.js and other code should refer to those.

### Actual behaviour
Routes are scattered across different places:
https://github.com/LiskHQ/lisk-hub/blob/development/src/constants/routes.js
https://github.com/LiskHQ/lisk-hub/blob/development/src/utils/routes.js
https://github.com/LiskHQ/lisk-hub/blob/development/src/components/app/index.js
https://github.com/LiskHQ/lisk-hub/blob/development/src/components/login/login.js#L103
and more

michaeltomasik:
### What was the problem?
Some routes definitions were in the string
### How did I fix it?
Created constants for route
### How to test it?
Search for `main` and `explorer` to check if route definition comes from one file
### Review checklist
- The PR solves #499 


src/constants/routes.js
```diff
@@ -1,10 +1,16 @@
...
-  search: { long: '/explorer/search', short: '/search' },
-  searchResult: { long: '/explorer/result', short: '/result' },
-  transaction: { long: '/explorer/transactions', short: '/transactions' },
+  wallet: { url: '/transactions' },
+  account: { url: '/accounts' },
+  accountVisualDemo: { url: '/account-visual-demo' },
+  dashboard: { url: '/dashboard' },
+  search: { url: '/search' },
+  searchResult: { url: '/result' },
+  transaction: { url: '/transactions' },
```
yasharAyari: We have already had a transaction route, So the second one is extra.

src/components/login/login.test.js
```diff
@@ -85,18 +86,18 @@ describe('Login', () => {
   });
 
   describe('History management', () => {
-    it('calls this.props.history.replace(\'/main/dashboard\')', () => {
+    it(`calls this.props.history.replace('${routes.main}${routes.dashboard.url}')`, () => {
```
ginacontrino: Please update the `it` blocks with descriptions that are more readable (as you were already changing them anyway). We want to avoid exactly the situation that you were in here, where when we change some implementation, the describe blocks have to be updated too. Instead, by having a general readable description of what we are expecting, there will be no need to always update the blocks. So one possible example could be, instead of 
`calls this.props.history.replace('${routes.main}${routes.dashboard.url}')` -> `redirects to dashboard`

src/utils/routes.js
```diff
@@ -3,56 +3,56 @@ import routes from './../constants/routes';
...
   }, {
     regex: /\/explorer\/result\/([0-9]+|[a-z]+)(?:\/[^/]*)?$/,
-    path: new RegExp(`${routes.searchResult.long}/([0-9]+|[a-z]+)/`),
+    path: new RegExp(`/explorer${routes.searchResult.url}/([0-9]+|[a-z]+)/`),
     params: 'query',
     name: 'result',
   }, {
     regex: /\/explorer\/search(?:\/[^/]*)?$/,
-    path: `${routes.search.long}/`,
+    path: `/explorer${routes.search.url}/`,
```
ginacontrino: Why not `${routes.explorer}` instead of `/explorer`?

PR #578, opened 2018-03-20
Fix filters persist - Closes #531

### LINKED ISSUE #531: Delegates page filters persist in confirm step
### Expected behaviour
Delegates vote confirm page should always display all changed delegates.

### Actual behaviour
Delegates vote confirm page shows only filtered delegates if filter was applied.

### Steps to reproduce
- Go to delegates page
- Click on "Not voted" and select some delegates
- Click on "Voted" and select some delegates
- Click on "Next" and only Voted delegates are displayed

michaeltomasik:
### What was the problem?
Delegates vote confirm page didn't display all changed delegates.
### How did I fix it?
Added show all filters when you are on confirm page.
### How to test it ?
- Go to delegates page
- Click on "Not voted" and select some delegates
- Click on "Voted" and select some delegates
- Click on "Next"
- Check if all changed delegates are displayed.

### Review checklist
- The PR solves #531

### DIFF
--- a/src/components/delegateList/votingRow.js
+++ b/src/components/delegateList/votingRow.js
@@ -11,7 +11,7 @@ const setRowClass = (voteStatus) => {
   if (pending) {
     return styles.pendingRow;
   } else if (confirmed !== unconfirmed) {
-    return confirmed ? `${styles.downVoteRow} votedRow` : `${styles.upVoteRow} votedRow`;
+    return confirmed ? `${styles.downVoteRow} selected-row` : `${styles.upVoteRow} selected-row`;
   }
   return confirmed ? styles.votedRow : '';
 };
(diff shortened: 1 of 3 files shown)


src/components/delegateList/votingRow.js
```diff
@@ -11,7 +11,7 @@ const setRowClass = (voteStatus) => {
   if (pending) {
     return styles.pendingRow;
   } else if (confirmed !== unconfirmed) {
-    return confirmed ? styles.downVoteRow : styles.upVoteRow;
+    return confirmed ? `${styles.downVoteRow} votedRow` : `${styles.upVoteRow} votedRow`;
```
slaweet: We use dash-separated class names for tests - `voted-row`. 
But the class should IMO be named `vote-row` and be set directly on the `ul` in https://github.com/LiskHQ/lisk-hub/blob/bc21d45ef1456e57edde098aa19f2e2c52813ca1/src/components/delegateList/votingRow.js#L32

michaeltomasik: It is possible but I will need to write same conditions directly on `ul` because I need that class specifically for rows that were selected, otherwise I would use `delegate-row`.
I can't use `delegate-row` to count it because in HTML it shows all rows and hides on that shouldn't be displayed.

PR #579, opened 2018-03-20
Remove hash in urls - Closes #327

### LINKED ISSUE #327: Remove hash in urls
### Expected behaviour
The URLs should be without `#` sign: http://lisk.io/hub/main/dashboard

In the electron wrapper `#` is used becase it loads `.../index.html`, but it might be worth to switch it to a simple node server inside the electron app.

### Actual behaviour

The URLs currently would be http://lisk.io/hub/#/main/dashboard

faival:
### What was the problem?
We're using hashHistory and we should try to use browserHistory

### Review checklist
- The PR solves #327

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -6,5 +6,6 @@
   "author": {
     "name": "Lisk Foundation",
     "email": "admin@lisk.io"
-  }
+  },
+  "homepage": "."
 }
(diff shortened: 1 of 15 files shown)


faival:
Suspended, there is dev-ops work involved and 1 failing test externalLinks.test.js, which we cannot test immediately and therefore merge will be delayed.

slaweet:
Closing as this PR is not relevant without the hosted version of Hub

PR #582, opened 2018-03-20
Create a guide for writing tests - Closes #516

### LINKED ISSUE #516: Create a guide for writing tests
### Expected behaviour

Create a `docs/tests_guide.md` that covers:
- how are unit tests organized
- how are integration tests organized
- how are e2e tests organized
- which of them should I use when

Links to some relevant articles would be a useful part of the documentation to reduce copy-paste.

The file should be linked from PR template

slaweet:
### What was the problem?
We didn't have a test guide
### How did I fix it?
I wrote a test guide
### How to test it?
Read the guide, look for errors, inacuracies, and missing stuff.
### Review checklist
- The PR solves #516

### DIFF
--- a/docs/TEST_GUIDE.md
+++ b/docs/TEST_GUIDE.md
@@ -15,6 +15,24 @@ But then we also need several bigger (integration) tests to make sure that the u
 [this](https://giphy.com/gifs/fail-technology-i5RWkVZzVScmY/fullscreen).
 The same argument holds true for unit vs. integration tests as well as for integration vs. end-to-end (E2E) tests. 
 
+### Which level should I use for a given feature?
+With the previous paragraph in mind, you should use the lowest possible level that covers the whole feature. 
+
+This section illustrates the problem on the "Send LSK" feature. Feature description: If user enters valid address and amount smaller than their balance, then they should be able to send the transaction to Lisk network.
+
+Test split:
+- **Unit test**: If the whole feature is inside one component. E.g. Address validation, because validity of the entered address depends only on the entered value and a regexp inside the component. 
+- **Integration test**: If the feature relies on multiple components or Redux store. E.g. Balance validation, because the validity of the balance depends on the entered value and the account balance in Redux store.
+- **E2E test**: If the feature involves multiple pages and/or API calls. E.g. Sending a LSK transaction with valid inputs.
+
+### Where is each level mocking?
+All the tests make assumptions about how things work outside of their scope. These assumptions are made by mocking the environment. Each level mocks something else:
+- **Unit test** mock called functions and methods.
+- **Integration test** mock server API, router, localStorage.
+- **E2E test**: use a local devnet node of Lisk which is a mock of Lisk Mainnet network.
+
+The problem of these assumptions is that we cannot be 100% sure that the mocks behave exactly as the things they're mocking. If the mock is not accurate, then we can have a passing test and broken feature. Therefore, every feature should have at least one test that doesn't mock.
+
 
 ## Unit tests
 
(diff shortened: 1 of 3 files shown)


docs/TEST_GUIDE.md
```diff
@@ -0,0 +1,66 @@
...
+Configuration is in [karma.conf.js](/LiskHQ/lisk-hub/blob/development/karma.conf.js) and [config/webpack.config.test.js](/LiskHQ/lisk-hub/blob/development/config/webpack.config.test.js).
+
+### What tools are used?
+- **Assertions** use `expect` syntax of [chai](http://www.chaijs.com/), e.g.:
+https://github.com/LiskHQ/lisk-hub/blob/80afda8289b75cb70bf345d14d36117fde2bdd90/src/utils/passphrase.test.js#L136
+
+- **Spies, stubs, ands mocks** use [sinon](http://sinonjs.org/), e.g.:
+https://github.com/LiskHQ/lisk-hub/blob/19a2a613fa08a34fe3088b0e40c11c7fa37a645d/src/store/middlewares/savedSettings.test.js#L20
+
+- To test **React compoments** we use [enzyme](http://airbnb.io/enzyme/), e.g.:
```
faival: Should we mention here the helper `chai-enzyme` that can be used also to simplify tests?

faival: Overall, I miss some brief/general hints on exclusion of tests between the different layers. 

Scenario: sumComponent - a components that sums two numbers. 
- unit --> test that the sum is correct
- integration 
-- test that the sum is displayed, whatever it is, does not need to assert that is actually correct. 
-- test that when the sum is displayed, no other components are affected
- e2e
-- we can assert here that the sum contains the actual production value expected, the rest is already covered?

reyraa: Stateless components will need unit tests only if they're not fully covered by their stateful parents.

docs/TEST_GUIDE.md
```diff
@@ -0,0 +1,75 @@
...
+https://github.com/LiskHQ/lisk-hub/blob/79165170a326a7f98efee098732e55be37d31223/src/components/toolbox/sliderCheckbox/index.test.js#L21
+
+
+## Integration tests
+
+### What do they test?
```
reyraa: I believe it is vital to note here to not test the same functionality which has been tested in unit tests and/or e2e tests. we have to test logics which may result in different behaviour/response if the `state` was different from our mock in unit tests. 
Take [transactionID](https://github.com/LiskHQ/lisk-hub/blob/0.4.0/test/integration/transactionID.test.js) for instance.
The [SingleTransaction](https://github.com/LiskHQ/lisk-hub/blob/0.4.0/src/components/singleTransaction/index.js) component contains a simple JSX which receives `transaction` and `peers` through `redux-connect`. We test the corresponsing action and reducer using unit tests, we "should" have an e2e test scenario starting from login page, through transaction list and ending in transaction detail page. So I can not see a  remarkable value added by this integration test.

docs/TEST_GUIDE.md
```diff
@@ -0,0 +1,75 @@
...
+
+## End-to-end (E2E) tests
+
+### What do they test?
+Full user scenarios in the application as a whole, including the communication with Lisk Core. 
+
+### How to run them?
+See [relevant sections of README](/LiskHQ/lisk-hub#run-end-to-end-tests)
+
+### How are they organized?
```
reyraa: Just like the other 2, I think we need to define boundaries of e2e tests scenarios so we don't end up having a scenario like:
  

-   Given I'm on login page
-   When I fill in passphrase of "genesis" to "passphrase" field
-   And I click "login button"
-   Then I should be logged in as "genesis" account

slaweet:
I added more sections to clarify your comments @reyraa @faival. Please let me know if you think more is needed.

slaweet:
@michaeltomasik I had the gif in there at first, but they are so distracting that I couldn't focus on reading the text ;-)
I'm using Grammarly in the browser, just didn't check it as I was typing in Vim :-)

PR #591, opened 2018-03-22
Show loader until peer is set - Closes #510

### LINKED ISSUE #510: Show splash screen until peer is set
### Expected behaviour
The loader should show, when the peer is not set yet

### Actual behaviour
Currently, the loader doesn't consider if the peer (`props.peers.data` ) is set

### Steps to reproduce
checkout `src/components/loadingBar/` and e.g. `src/components/transactions/transactionList`

ginacontrino:
### What was the problem?
Loader didn't consider if the peer is already set

### How did I fix it?
By checking for peer

### How to test it?
Slow your network down and refresh the page (and compare to before) - unfortunately the difference is not that big

### Review checklist
- The PR solves #510

### DIFF
--- a/src/components/accountTransactions/index.test.js
+++ b/src/components/accountTransactions/index.test.js
@@ -36,7 +36,7 @@ describe('AccountTransaction Component', () => {
   });
 
   it('updates transactions on address update', () => {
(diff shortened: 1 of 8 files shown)


ginacontrino:
There is a new PR for this: #673

Issue #585, opened 2018-03-21
Move route definitions from app component to constants/routes.js


slaweet:
### Expected behaviour
Routes should be defined in https://github.com/LiskHQ/lisk-hub/blob/development/src/constants/routes.js 
```

export default {
  dashboard: { 
    path: '/main/dashboard',
    component: Dashboard,
    isPrivate: true,
  },
  ...
```
And `app` component should just loop over it, something like:
```
<Switch>
{Object.keys(routes).map(key => (
  <{routes[key].isPrivate ? PrivateRoutes : Route}  render={ ({ match }) => (
     <main className={routes[key].isPrivate ? offlineStyle.disableWhenOffline : null}>
      <Route path={routes[key].path}} component={routes[key].Component} />
    </main>
   )} />
)}
<Switch>
```

https://github.com/LiskHQ/lisk-hub/issues/584 can be solved in here too

### Actual behaviour
App component lists all routes
https://github.com/LiskHQ/lisk-hub/blob/1d734bdb0cd669696d56726a9dd206570226c863/src/components/app/index.js#L40-L65

michaeltomasik:
Done in https://github.com/LiskHQ/lisk-hub/pull/625

PR #625, opened 2018-03-28
Improve url scheme - Closes #608

### LINKED ISSUE #608: Improve url scheme
### Expected behaviour
- Url schema should not have `main`
- voting should be changed to delegates
- transactions to wallet

### Actual behaviour
- Url schema have `main`
- Voting path on delegates view
- transactions path on a wallet view
### Steps to reproduce
Search project for `main` path
check delegates and wallet url

michaeltomasik:
### What was the problem?
We are still using `main` route and old naming
### How did I fix it?
refactor code so it will use new terminology and no `main` route
### How to test it?
Check wallet and delegates urls if they match. Also make sure there is no `main` in any route.
Linked to https://github.com/LiskHQ/lisk-hub/issues/585
### Review checklist
- The PR solves #608


src/components/app/index.js
```diff
@@ -37,6 +24,13 @@ class App extends React.Component {
   }
 
   render() {
+    const allRoutes = Object.entries(routes).map(x => x[1]);
```
pitu7dg: Use Object.values instead

michaeltomasik: Good Catch 👍

slaweet:
Some of these `main/transactions`  need to be updated, especially in e2e tests:
```
$ git grep 'main/transactions'
src/store/middlewares/login.js:  // redirect to main/transactions
src/utils/externalLinks.js:          .replace('/main/transactions/send', '/wallet')
test/e2e/explorer.feature:    Then I should be on url "/?referrer=/main/transactions%3Faddress%3D15610359283786884938L"
test/e2e/savedAccounts.feature:    Then I go to "main/transactions"
test/e2e/transactions.feature:    Then I should be on url "/main/transactions"
```
I suspect that there are other changed paths used in e2e test.

PR #652, opened 2018-03-30
Fix random index calculation in resetForm() - Closes #651

### LINKED ISSUE #651: Random index does not return random index in Confirm
### Expected behaviour

From naming and code comments, [this function](https://github.com/LiskHQ/lisk-hub/blob/f06d412/src/components/passphrase/confirm/index.js#L104) should return a random index:

```
const indexByRand = num => Math.floor(num * (words.length - 1));
indexByRand(Math.random());
```

### Actual behaviour

It does not include the last index. This is because Math.random() is strictly < 1.0 and thus `num * (words.length - 1)` <  (words.length - 1) and which makes `words.length - 1` impossible.

Even if Math.random() would include 1.0, the last index would be chosen significantly less often then the other indices because only the value 1.0 leads to the last index where the other indices have a whole interval.

### Steps to reproduce

Simulate using this `node rnd_test.js`:

```
const words_length = 12;
const indexByRand = num => Math.floor(num * (words_length - 1));
const randomIndex = () => indexByRand(Math.random());

let statistics = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // number of occurrence for index 0 to 11.

for (let i = 0; i < 10000000; i++) {
    statistics[randomIndex()] += 1;
}

console.log(statistics);
```

prints

```
[ 907928,
  910979,
  908350,
  910128,
  909866,
  909620,
  907623,
  908070,
  908512,
  908911,
  910013,
  0 ]
```

webmaster128:
### What was the problem?

random index did not return random index. See #651.

### How did I fix it?

Apply correct formula, see #651

### How to test it?

Hmm, I din't know how to build this project. But I guess account creation should now include the last word in the password confirmation step.

### Review checklist
- The PR solves #651
- All new code is covered with unit tests?
- All new features are covered with e2e tests?
- All new code follows best practices?

### DIFF
--- a/src/components/passphrase/confirm/index.js
+++ b/src/components/passphrase/confirm/index.js
@@ -90,7 +90,6 @@ class Confirm extends React.Component {
 
   resetForm() {
     const words = this.props.passphrase.match(/\w+/g);
-    const indexByRand = num => Math.floor(num * (words.length - 1));
 
     /**
      * Returns a random index which doesn't exist in list
@@ -101,7 +100,7 @@ class Confirm extends React.Component {
     const randomIndex = (list) => {
       let index;
       do {
-        index = indexByRand(Math.random());
+        index = Math.floor(Math.random() * words.length);
       }
       while (list.includes(index));
       return index;


webmaster128:
CI is failing with "TimeoutError: waiting for element '.joyride-tooltip__header'". I have no clue what that means. Can this be related to my change?

reyraa:
I'm so glad to see you dove so deep into code to find this issue. My presumption - mistakenly - was that 1 is inclusive in the returned value of `Math.random`. Thanks for the descriptive ticket and clean PR.

webmaster128:
My pleasure :) All 3 passphrase related PRs can be ported to Lisk Nano. What is the strategy here? Are those changes needed in nano? Do you want me to port them? Or do you do that on your own?

PR #676, opened 2018-04-04
Update balance in account switcher - Closes #595

### LINKED ISSUE #595: Update balances in account switcher without entering accounts
### Expected behaviour
The balances in the account switcher should update

### Actual behaviour
Currently when you open the account switcher, the balances are not updated

ginacontrino:
### What was the problem?
The balances in the account switcher didn't update when changed

### How did I fix it?
Made an api call to get the account information, if the saved accounts are in a new block

### How to test it?
Make a transaction, open the account switcher -> the balance should be updated

### Review checklist
- The PR solves #595

### DIFF
--- a/src/store/middlewares/savedAccounts.js
+++ b/src/store/middlewares/savedAccounts.js
@@ -2,6 +2,8 @@ import actionTypes from '../../constants/actions';
 import { accountLoading, accountLoggedOut } from '../../actions/account';
 import { accountsRetrieved, accountSaved } from '../../actions/savedAccounts';
 import { activePeerSet } from '../../actions/peers';
+import { extractAddress, getAccount } from '../../utils/api/account';
+import { getLastActiveAccount } from '../../utils/savedAccounts';
 import getNetwork from '../../utils/getNetwork';
 import networks from '../../constants/networks';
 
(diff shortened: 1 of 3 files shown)


src/store/middlewares/savedAccounts.js
```diff
@@ -27,10 +29,43 @@ const savedAccountsMiddleware = (store) => {
...
+      });
+    });
+  };
+
   return next => (action) => {
     next(action);
     const { peers, account, savedAccounts } = store.getState();
     switch (action.type) {
+      case actionTypes.newBlockCreated:
+        if (action.data.windowIsFocused) {
```
slaweet: If the window is focused afterward, there is again the problem of outdated balances. I would remove this condition.

src/store/middlewares/savedAccounts.js
```diff
@@ -27,10 +29,43 @@ const savedAccountsMiddleware = (store) => {
     }
   });
 
+  const updateSavedAccounts = (peers, tx, savedAccounts) => {
+    const { accounts } = savedAccounts;
+    tx.forEach((transaction) => {
```
slaweet: if there are 10 transactions of an account in one block, then the account is fetched 10 times. I know it's unlikely, but still.

PR #677, opened 2018-04-04
Fix vote protocol twice in row - Closes #661

### LINKED ISSUE #661: Fix use of vote launch protocol twice in a row
### Expected behaviour
You should always be able to call launch protocol twice in a row

### Actual behaviour
When We using protocol after being on delegates page it is not re-rendering

michaeltomasik:
### What was the problem?
After calling voting protocol 2 times in a row electron app didn't update
### How did I fix it?
Added refresh function after calling launch voting protocol
### How to test it?
 - run `npm run pack` to compile electron app
 - install electron app from dist
 - go to https://earnlisk.com/ and click `vote 3`
- click `vote 4` and check if votes updated
### Review checklist
- The PR solves #661

### DIFF
--- a/src/utils/externalLinks.test.js
+++ b/src/utils/externalLinks.test.js
@@ -5,7 +5,7 @@ import history from '../history';
 import routes from '../constants/routes';
 
 describe('externalLinks', () => {
-  const historyPush = spy(history, 'push');
+  const historyReplace = spy(history, 'replace');
   const ipc = {
     on: spy(),
   };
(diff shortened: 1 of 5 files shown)


src/utils/externalLinks.test.js
```diff
@@ -5,7 +5,7 @@ import history from '../history';
 import routes from '../constants/routes';
 
 describe('externalLinks', () => {
-  const historyPush = spy(history, 'push');
+  const historyReplace = spy(history, 'replace');
```
slaweet: You should use `mock` instead of `spy` to prevent `HeadlessChrome 65.0.3325 (Linux 0.0.0) ERROR Some of your tests did a full page reload!` in https://jenkins.lisk.io/job/lisk-hub/job/PR-677/1/console

Issue #149, opened 2018-01-03
Review and reduce amount of e2e tests


slaweet:
### Expected behaviour
There should be fever e2e test scenarios. Some will be replaced with integration tests (as marked in #95) in other tasks. Other e2e tests should be merged to test the app in a more holistic and realistic way. In the end there should be 10 to 20 e2e test scenarios

### Actual behaviour
There are too many e2e tests - about 40 scenarios.

PR #680, opened 2018-04-04
Review and reduce amount of e2e tests - #149

yasharAyari:
### What was the problem?
There needed to have less e2e test scenarios. 

### How did I fix it?
Remove unnecessary ones and mege some of them.

### Review checklist
- The PR solves #149 


test/e2e/send.feature
```diff
@@ -8,19 +8,14 @@ Feature: Send dialog
...
-    Then I should see text "Transaction is being processed and will be confirmed. It may take up to 15 minutes to be secured in the blockchain." in "result box message" element
-
+    Then I click "dashboard" menu
+    And I should see 3 rows
+    When I click "seeAllLink"
+    Then I should be on url "/main/transactions"
+    And I should see 26 rows
+    When I scroll to the bottom of "transaction results"
+    Then I should see 51 rows
+  @pending
```
slaweet: `Scenario: should be able to init account if needed` should not be @pending.

test/e2e/send.feature
```diff
@@ -8,19 +8,14 @@ Feature: Send dialog
...
-    And I wait 1 seconds
-    Then I should see text "Transaction is being processed and will be confirmed. It may take up to 15 minutes to be secured in the blockchain." in "result box message" element
-
+    Then I click "dashboard" menu
+    And I should see 3 rows
+    When I click "seeAllLink"
+    Then I should be on url "/main/transactions"
+    And I should see 26 rows
+    When I scroll to the bottom of "transaction results"
+    Then I should see 51 rows
```
slaweet: Adding these 7 lines to `test/e2e/send.feature` doesn't make much sense. I know you want to have bigger scenarios, but it's making it harder to find what is covered by e2e tests and what is not.

test/e2e/voting.feature
```diff
@@ -1,47 +1,23 @@
...
-    When I go to "main/voting/"
     And I fill in "genesis_42" to "search" field
     Then I should see 1 instances of "delegate row"
-    And I clear "search" field
+    And I click "clean icon"
     And I fill in "doesntexist" to "search" field
     Then I should see 0 instances of "delegate row"
     And I should see text "No delegates found." in "empty message" element
-
-  @testnet
```
slaweet: There still should be a scenario that does a vote with `@testnet` because we had before the issue that all transactions worked with localhost but not with testntet/mainnet.

PR #697, opened 2018-04-06
Double click on "Yes it's safe" breaks whole app - Closes #642

### LINKED ISSUE #642: Double click on "Yes it's safe" breaks whole app
### Expected behaviour
The second click shouldn't preform any actions

### Actual behaviour
As title suggests if we double click on _Yes it's safe_ button in Registration process, the component throws error.

### Steps to reproduce
* Navigate to Register page
* Proceed until you get the passphrase and see the button titled : Yes it's safe.
* double click on it

yasharAyari:
### What was the problem?
As title suggests if we double clicked Yes it's safe button in Registration process, the component throwed error.

### How did I fix it?
Disable the button after first click.

### Review checklist
- The PR solves #642

### DIFF
--- a/src/components/passphrase/safekeeping/index.js
+++ b/src/components/passphrase/safekeeping/index.js
@@ -10,6 +10,7 @@ class SafeKeeping extends React.Component {
     super();
     this.state = {
       step: '',
+      isSafe: false,
     };
   }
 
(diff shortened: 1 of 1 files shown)


alepop:
@yasharAyari hello! I think, there are a lot of place in the app, then you need to prevent double click, maybe better is to make a HOC component for click elements?

reyraa:
@alepop Correct. thanks for your suggestion. we can create an HOC for all buttons presenting double click event. I'll create a ticket to fix it with next release.

yasharAyari:
@alepop I don't agree with you . I think we can handle all of them with 'disabled' property in button component like this one.

alepop:
@yasharAyari you can create a HOC that will set the `disabled` attribute automatically on the first click. In my opinion, it is better than add all this code in each component.

PR #706, opened 2018-04-09
Animate shape colors in registration - Closes #703

### LINKED ISSUE #703: Improve shapes in "create step" of registration
### Expected behaviour
Shapes in "create step" of registration should change colors when moving

### Actual behaviour
They are not changing colors

ginacontrino:
### What was the problem?
The shapes' colors were not animated

### How did I fix it?
By changing the gradients depending on the progress percentage + fading some of them out towards the end

### How to test it?
Try to register a new account, in the create step you should see the color gradient animation

used scheme:
screen shot 2018-04-09 at 18 18 39

screen shot 2018-04-09 at 18 21 07

### Review checklist
- The PR solves #703


slaweet:
The issue says:
> Shapes in "create step" of registration should change colors when moving

Now the colors are changing even when the shapes/mouse are not moving. It doesn't tell the story that the color depends on moving the mouse.

PR #712, opened 2018-04-11
Refactor and Increase the unit tests of SliderCheckbox component - Closes #242

### LINKED ISSUE #242: Improve test coverage of checkbox component
### Expected behaviour
checkbox component should have more that 80% test coverage 
### Actual behaviour
It is excludes in karma config file

reyraa:
### What was the problem?
 - Some if-else statements could be refactored for ease of use in unit testing.
 - The `textAsIcon` options could be removed in favour of having icons for languages, which helps us implement all language names and signs with out the need to install specific font package.
 - The was an issue with `onClick` event listener which caused the`change` method call even for drag length of less than 50px.
 - Unit test coverage was insufficient.

### How did I fix it?
 - All above issues are fixed.

### How to test it?
Make sure:
 - If you slide for less than 50px the slider checked state won't change.
 - Click event works
 - Unit tests pass and cover above 90%

### Review checklist
- The PR solves #242 

### DIFF
--- a/src/components/toolbox/sliderCheckbox/index.js
+++ b/src/components/toolbox/sliderCheckbox/index.js
@@ -62,9 +62,9 @@ class SliderCheckbox extends React.Component {
   stopTracking() {
     if (this.trackable) {
       this.trackable = false;
-
-      if (Math.abs(this.delta) > 50 && !this.props.disabled) {
-        this.change('swiped');
+      // delta === 0 covers clicking
+      if ((this.delta === 0 || Math.abs(this.delta) > 50) && !this.props.disabled) {
+        this.change();
       } else {
         this.revert();
       }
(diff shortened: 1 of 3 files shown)


src/components/toolbox/sliderCheckbox/index.js
```diff
@@ -38,21 +36,17 @@ class SliderCheckbox extends React.Component {
...
-      this.buttonWidth = buttonWidth;
-      this.maxMovement = sliderWidth - buttonWidth;
-    } else {
-      this.buttonWidth = this.props.buttonWidth;
-      this.maxMovement = this.props.maxMovement;
-    }
+    /* istanbul ignore next */
+    const sliderWidth = this.props.sliderWidth || this.parent.getBoundingClientRect().width;
+    /* istanbul ignore next */
+    const buttonWidth = this.props.buttonWidth || this.shape.getBoundingClientRect().width;
```
faival: do we need the extra const declaration? I think it can be safely assigned to this.

src/components/toolbox/sliderCheckbox/index.js
```diff
@@ -117,45 +109,38 @@ class SliderCheckbox extends React.Component {
...
-            }
+            <FontIcon className={theme.icon}>
+              {icons.checked || 'checkmark'}
+            </FontIcon>
           </span>
         </span>
         { label ?
           <div>
-            <span>{label}</span>
+            <span className='label'>{label}</span>
```
faival: I would suggest to use less generic class names for unit tests, there might be class collisions coming from 3rd party libs? 
Maybe better to scope it to the component name? `slider-check-label` , `slider-check-arrow`

src/components/toolbox/sliderCheckbox/index.test.js
```diff
@@ -32,40 +68,72 @@ describe('SliderCheckbox without HOC', () => {
...
+    const wrapper = mount(<SliderCheckbox {...props} />);
+    wrapper.find('input').instance().checked = true;
+    shouldBeChecked(wrapper);
+    drag(wrapper, -51);
+    isItTransformed(wrapper, '249px');
+    stopDragging(wrapper);
 
-    // Then the box should not be checked yet
-    expect(wrapper.find('input').instance().checked).to.equal(false);
+    // // The checkbox should be checked
```
faival: I think this comment does not match the following assertion statement.

src/components/toolbox/sliderCheckbox/index.js
reyraa: I've put it there to keep the other lines shorter and more readable. so I would say Yes.

src/components/toolbox/sliderCheckbox/index.js
reyraa: We don't use 3rd party libraries unless we isolate them with. css modules. this is the beauty of our CSS stack which allows us use more human readable names.

PR #766, opened 2018-04-25
Disable text marking

shilch:
### What was the problem?
Both, the browsers and the electron apps, allowed the user to select text.  
It affects the UX insofar that Lisk Hub less feels like an app but more like a website with text.  
Included are pictures before and after the fix.

bildschirmfoto 2018-04-25 um 23 46 30
bildschirmfoto 2018-04-25 um 23 48 07

### How did I fix it?
Globally applied the rule `user-select: none` in the app.css file.

### How to test it?
- Just try marking or press CTRL+A ;-)
- The change does **not** affect inputs. They still function like before (marking input contents, clicking buttons, ...)

### Review checklist
- The PR solves #INSERT_ISSUE_NUMBER

### DIFF
--- a/src/assets/css/styles.head.css
+++ b/src/assets/css/styles.head.css
@@ -1,7 +1,3 @@
-* {
-  user-select: none;
-}
-
 body {
   background-color: #013165;
 }
--- a/src/assets/css/styles.head.css
+++ b/src/assets/css/styles.head.css
@@ -1,3 +1,7 @@
+* {
+  user-select: none;
+}
+
 body {
   background-color: #013165;
 }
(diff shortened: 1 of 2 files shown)


reyraa:
We appreciate your contribution and apologize for the belated response. Although this is a good idea, using a wildcard causes some functionalities break. for example, Passphrase inputs use range selection to handle input events and they won't work after your change applied.

shilch:
Thanks for your response!  
Yeah, wildcards are not exactly brilliant - but applying the style to every .class and #id either.  
What about keeping the wildcard but disabling it for the elements that break? I.e. applying `user-select: all` to the passphrase inputs.

slaweet:
I suggest to disable the text selection just for the menu. Inside the main content, it can still be useful to be able to select text, e.g. if I get an error message and want to google what can I do about it.

PR #768, opened 2018-04-26
Fix a bug in showing vote summary - Closes #728

### LINKED ISSUE #728: Votes overview doesn't show the delegate if not loaded in the initial list of delegates
## Description
The confirmation step of voting doesn't show the delegates which are not loaded in delegates list. This issue happens because we filter the current list of delegates, and if a delegate is searched through search box and is not initially loaded into the list of delegates, it's technically impossible to show it the overview list of votes. 

## Is this a bug?
Yes, since the list of votes overview is different from what you will vote eventually.

## Suggested solution
One approach would be to inject the searched item into the list of delegates. but then we have to filter the list every time we load more items, to prevent duplicated items. But since we clean the list for filtering this method will not help.

My suggestion is to add Rank and Productivity to vote list items and then render the overview list using the vote list itself.

## In case this is a bug, is it unique to this new version or is it possible that it exists in older versions as well?
I can not find new changes that may caused this. This is likely to be an old bug.

## How to reproduce the situation?
Try voting for any delegates with rank above 101 before infinite scrolling.

reyraa:
### What was the problem?
Explained in #728 

### How did I fix it?
I used the vote list to render the vote summary view. to do so, I had to use multiStep component
and create dedicated components for vote item and delegate item and hence, their dedicated list components. 
I've also cleaned up our unit tests.

### How to test it?
Make sure if you search for a delegate outside the scope of loaded delegates, and click next, you'll see all the delegates you votes for, in vote summary list.

### Review checklist
- The PR solves #728 


src/components/votingListView/votingListView.js
```diff
@@ -144,54 +139,41 @@ class DelegateList extends React.Component {
...
           <div className={styles.table}>
-            <ul className={`${styles.tableHead} ${grid.row}`}>
-              <li className={`${grid['col-md-1']} ${grid['col-xs-2']} ${styles.leftText}`}>{this.props.t('Vote', { context: 'verb' })}</li>
-              <li className={`${grid['col-md-1']} ${grid['col-xs-2']}`}>{this.props.t('Rank')}</li>
-              <li className={`${grid['col-md-3']} ${grid['col-xs-5']}`}>{this.props.t('Name')}</li>
-              <li className={`${grid['col-md-5']}`}>{this.props.t('Lisk ID')}</li>
-              <li className={`${grid['col-md-2']} ${grid['col-xs-3']} ${styles.productivity}`}>{this.props.t('Productivity')}</li>
-            </ul>
-            { this.getList(filteredList) }
+            <ListLabels t={t} styles={styles} grid={grid} status={showChangeSummery} />
```
slaweet: Why pass `styles={styles} grid={grid}` instead of importing them in `ListLabels`? Or going one step further, `ListLabels` could have its own css file. Not necessarilly in this PR.

Issue #772, opened 2018-04-27
Create basic "Request LSK" component


ginacontrino:
### Expected behaviour
We want a basic component that looks like this:

![screen shot 2018-04-27 at 11 17 42](https://user-images.githubusercontent.com/9592216/39355240-f090e120-4a0c-11e8-91ac-6a70401f1fe2.png)

### Actual behaviour
Not implemented

PR #781, opened 2018-05-02
Implement basic request LSK tab

ginacontrino:
### What was the problem?
We had no way to request LSK

### How did I fix it?
By adding this , but without "request specific amount", so for now this is only one step
screen shot 2018-04-27 at 11 17 42

I put this component into the multistep anyway, since there will be more steps in future, so I had to change the multistep component to also work with only one child. This can be removed, once we have multiple children again.

### How to test it?
- Try clicking on "request"
- Make sure the QR code and everything else is working
- Then click again on `send`

### Review checklist
- The PR solves #772

### DIFF
--- a/src/components/sendWritable/send.js
+++ b/src/components/sendWritable/send.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import grid from 'flexboxgrid/dist/flexboxgrid.css';
 import { fromRawLsk } from '../../utils/lsk';
 import { Button } from './../toolbox/buttons/button';
 import { authStatePrefill } from '../../utils/form';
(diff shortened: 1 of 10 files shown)


test/integration/wallet.test.js
```diff
@@ -228,6 +228,14 @@ describe('@integration: Wallet', () => {
       step(`Then I should see text ${successMessage} in "result box message" element`, () => helper.haveTextOf('.result-box-message', successMessage));
     });
 
+    describe('Scenario: should display the request LSK component when clicking on the tab', () => {
+      step('Given I\'m on "wallet" as "genesis" account', () => setupStep('genesis'));
+      step('And I click "request tab"', () => { helper.clickOnElement('.request-tab'); });
+      step('Then I should see the QR code', () => helper.haveLengthOf('.request-qr-code', 1));
+      step('And I click "send tab"', () => { helper.clickOnElement('.send-tab'); });
```
slaweet: If we want to strictly follow Given-When-Then pattern, then both `And I click` (line 23 and 235) should be `When I click`. `And` is used only when there are two `When`s in a row.

src/components/sendWritable/send.js
```diff
@@ -82,6 +83,16 @@ class SendWritable extends React.Component {
...
           </header>
+
+          <div className={`${grid.row} ${styles.tab} `}>
+            <div className={`${grid['col-xs-6']} ${styles.tabActive} send-tab`}>
+              {this.props.t('Send')}
+            </div>
+            <div className={`${grid['col-xs-6']} ${styles.tabInactive} request-tab`} onClick={() => { this.props.setTabSend(false); }}>
+              {this.props.t('Request')}
+            </div>
+          </div>
```
slaweet: The tab switcher is duplicated in send and receive. Should be a separate component with own CSS.

PR #814, opened 2018-05-16
Add price converter e2e tests - Closes #796

### LINKED ISSUE #796: Cover USD/EUR conversion in an @advanced send e2e scenario

michaeltomasik:
### What was the problem?
No e2e test for price converter
### How did I fix it?
Added price converter
### How to test it?
Run test build
### Review checklist
- The PR solves #796

### DIFF
--- a/test/e2e/send.feature
+++ b/test/e2e/send.feature
@@ -4,8 +4,7 @@ Feature: Send dialog
     Given I'm logged in as "genesis"
     And I fill in "1" to "amount" field
     Then I should see "convertor" element
-    Given I wait 5 seconds
-    Then I should see "convertElem" element with text matching coverter price
+    Then I should see "converted price" element with text matching coverter price
     And I fill in "94495548317450502L" to "recipient" field
     And I click "send next button"
     And I click "send button"
--- a/test/e2e/send.feature
+++ b/test/e2e/send.feature
@@ -3,6 +3,9 @@ Feature: Send dialog
   Scenario: should allow to send when enough funds and correct address form
     Given I'm logged in as "genesis"
     And I fill in "1" to "amount" field
+    Then I should see "convertor" element
+    Given I wait 5 seconds
+    Then I should see "convertElem" element with text matching validator format
     And I fill in "94495548317450502L" to "recipient" field
     And I click "send next button"
     And I click "send button"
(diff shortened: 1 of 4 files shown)


test/e2e/step_definitions/send.step.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const { defineSupportCode } = require('cucumber');
+const { waitForElemAndMatchItsText } = require('../support/util.js');
+
+defineSupportCode(({ Then }) => {
+  Then('I should see "{elementName}" element with text matching coverter price', (elementName, callback) => {
+    const regexp = /^~ \d{0,100}(\.\d{0,2}){0,1}$/;
```
slaweet: The regexp could be a bit more restrictive `/^~ \d{1,100}(\.\d{1,2})?$/`. Also `?` is equal to and cleaner than `{0,1}`

Anyway, I really like the you're thinking about the future with making sure it will work even when LSK / USD is 1 / 10^100 😆 👍

test/e2e/send.feature
```diff
@@ -3,6 +3,9 @@ Feature: Send dialog
   Scenario: should allow to send when enough funds and correct address form
     Given I'm logged in as "genesis"
     And I fill in "1" to "amount" field
+    Then I should see "convertor" element
+    Given I wait 5 seconds
+    Then I should see "convertElem" element with text matching coverter price
```
slaweet: In e2e tests, we have a convention of kebap-case class names with dashes replaced from spaces. So here it would be "convert elem", in the step definition, you already have `replace(/ /g, '-')`, so in the converter component you can have `convert-elem` className. Of course, you'll need to update the className also in unit tests.

A more descriptive className would IMO be `converted-price`.

test/e2e/send.feature
```diff
@@ -3,6 +3,9 @@ Feature: Send dialog
   Scenario: should allow to send when enough funds and correct address form
     Given I'm logged in as "genesis"
     And I fill in "1" to "amount" field
+    Then I should see "convertor" element
+    Given I wait 5 seconds
```
slaweet: I think the wait is unneccesary, because the next step calls `waitForElemAndMatchItsText` and the converter component puts the element into DOM only when the value loaded, so it will wait.

PR #815, opened 2018-05-16
Fix for autoupdate error - Closes #808

### LINKED ISSUE #808: Error in auto updater
This happened when clicking "Update now" after clicking menu -> Help -> "Check for updates..."

screen shot 2018-05-09 at 16 23 20

michaeltomasik:
### What was the problem?
https://github.com/LiskHQ/lisk-hub/issues/808
### How did I fix it?
Added `window.close();` to UpdateApp
### How to test it?
Click "Update now" after clicking menu -> Help -> "Check for updates..."
### Review checklist
- The PR solves #808

### DIFF
--- a/app/src/modules/updateModal.js
+++ b/app/src/modules/updateModal.js
@@ -30,6 +30,8 @@ export default (electron, releaseNotes, updateCallBack, versions) => {
     });
     win.show();
   }, 1000);
+
+  ipcMain.removeAllListeners('update');
   ipcMain.on('update', () => {
     updateCallBack();
   });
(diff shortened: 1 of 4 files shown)


app/src/modules/updateModal.js
```diff
@@ -31,7 +31,6 @@ export default (electron, releaseNotes, updateCallBack, versions) => {
     win.show();
   }, 1000);
   ipcMain.on('update', () => {
-    win.close();
```
slaweet: I think you need to call
`ipcMain.removeAllListeners('update')` 
before 
`ipcMain.on('update', () => {	`
To prevent `updateCallBack();` being called twice in case I close the update window and open it again from the OS menu

https://electronjs.org/docs/api/ipc-renderer#ipcrendererremovealllistenerschannel

PR #821, opened 2018-05-18
Cache delegate names lookup in local storage - Closes #783

### LINKED ISSUE #783: Cache delegate names lookup in localStorage
To be used in
- vote launch protocol
- vote transaction detail
- deleagte detail

michaeltomasik:
### What was the problem?
Delegates in `Delegate vote` transaction type were fetching all delegate usernames by sending for each API request
### How did I fix it?
Added cache so it saves delegates data when user scrools delegate list
### How to test it?
Go to Delegate and scroll down a bit then Find `Delegate vote` transaction, open it and check if it makes additional requests.
### Review checklist
- The PR solves #783


karma.conf.js
```diff
@@ -66,6 +66,7 @@ module.exports = function (config) {
             'src/components/passphraseCreation/index.js',
             'src/utils/api/liskService.js',
             'src/components/errorBoundary/index.js',
+            'src/utils/delegates.js',
```
slaweet: Please put a comment into the file to explain why is this line there.

src/actions/voting.js
```diff
@@ -146,6 +147,7 @@ export const delegatesFetched = ({ activePeer, q, offset, refresh }) =>
         q,
       },
     ).then(({ delegates, totalCount }) => {
+      updateDelegateCache(delegates, activePeer.options.address || activePeer.options.name);
```
slaweet: IMO it would be a better-divided responsibility to define the localStorage key inside `updateDelegateCache`, so here it would be just `updateDelegateCache(delegates, activePeer)`

src/utils/delegates.js
```diff
@@ -0,0 +1,17 @@
...
+    delegate[address] = { publicKey, username };
+    return delegate;
+  }, {});
+  const updatedDelegates = { ...formatedDelegates, ...savedDelegates };
+
+  localJSONStorage.set(network, updatedDelegates);
+};
+
+export const loadDelegateCache = (network = networks.mainnet.name) =>
+  localJSONStorage.get(network, {});
```
slaweet: The key IMO shouldn't be just `network`. What if someone in the future will create another cache for something else that also depends on network. I would use something like `'delegateCache-${network}'`

PR #826, opened 2018-05-23
Better caching based on checksuming package.json - Closes #825

### LINKED ISSUE #825: Improve caching on Jenkins
### Expected behaviour

The cache should be used as often as possible to reduce build times.

### Actual behaviour

The current approach (cache `node_modules/` locally once a day and re-use for all other builds) does not work well any more since we have more than one Jenkins slave building lisk-hub.

### Steps to reproduce

Look at the build times of any given job: the "Install npm dependencies" step will take minutes or seconds depending on what machine it ends up running.

ghost:
### What was the problem?

The old caching behaviour was not working well with more than one Jenkins machines.

### How did I fix it?

Compute the sha1 checksum of package json and either
 - unpack the associated tarball if it exists or
 - create the tarball for future build to use
Tarballs older than 7 days are opportunistically deleted when a
new one gets created.

### How to test it?

Look at the build times of this PR: the "Install npm dependencies" step should consistently take only seconds after the first build on any given build machine.

### Review checklist
- The PR solves #825 

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,9 +52,16 @@ node('lisk-hub') {
     stage ('Install npm dependencies') {
       try {
         sh '''
-        cp -r ~/cache/development/node_modules ./ || true
+        cache_file="$HOME/cache/$( sha1sum package.json |awk '{ print $1 }' ).tar.gz"
+        if [ -f "$cache_file" ]; then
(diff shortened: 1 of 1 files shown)


slaweet (approved):
Looks good. Thank you @fchavant

PR #838, opened 2018-05-25
Vendor prefixes by stylelint - Closes

yasharAyari:
### What was the problem?
We shouldn't use browser prefix in css base on project style guide
### How did I fix it?
Add a new rule to stylelint config witch prevents this problem before happening

### Review checklist
- The PR solves #806

### DIFF
--- a/config/webpack.config.react.js
+++ b/config/webpack.config.react.js
@@ -123,7 +123,7 @@ module.exports = {
       template: './src/index.html',
       VERSION: bundleVersion,
       inject: false,
-      // inlineSource: '.(css)$',
+      inlineSource: '.(css)$',
     }),
     new HtmlWebpackInlineSourcePlugin(),
     new I18nScannerPlugin({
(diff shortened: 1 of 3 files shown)


config/webpack.config.react.js
```diff
@@ -122,7 +123,7 @@ module.exports = {
       template: './src/index.html',
       VERSION: bundleVersion,
       inject: false,
-      inlineSource: '.(css)$',
+      // inlineSource: '.(css)$',
```
faival: I think this will break the injection of css in head of document

src/components/votingListView/votingHeader.css
```diff
@@ -103,14 +103,6 @@
     border-bottom: solid 1px transparent;
     transition: all ease 250ms;
 
-    &:-moz-placeholder,
```
faival: I think the fix is to add an inline disable rule, as the input-placeholder are not supported by autoprefixer. 

https://github.com/postcss/autoprefixer/wiki/support-list

yasharAyari: it was a wrong rule. it didn't work at all.

faival: the webkit one is a valid one, not supported by autoprefixer.

https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-input-placeholder

PR #854, opened 2018-05-29
Currency set in converter is not remembered  - Closes #846

### LINKED ISSUE #846: Currency set in converter is not remembered

michaeltomasik:
### What was the problem?
The currency was not saved
### How did I fix it?
Added an active converter to localStorage
### How to test it?
Find currency converter, change currency to `EUR`, refresh page and check if currency is still `EUR`
### Review checklist
- The PR solves #846

### DIFF
--- a/src/components/converter/index.js
+++ b/src/components/converter/index.js
@@ -2,6 +2,8 @@ import React from 'react';
 import liskServiceApi from '../../utils/api/liskService';
 import Input from '../toolbox/inputs/input';
 import { fromRawLsk } from '../../utils/lsk';
+import localJSONStorage from '../../utils/localJSONStorage';
+
 import fees from './../../constants/fees';
 
 import styles from './converter.css';
@@ -17,6 +19,11 @@ class Converter extends React.Component {
       currencies: ['USD', 'EUR'],
     };
     this.fee = fees.send;
+
+    const localStorageCurrency = localJSONStorage.get('currency', '');
+    if (localStorageCurrency) {
+      this.selectActive(localStorageCurrency);
+    }
     this.updateData();
   }
 
@@ -37,6 +44,7 @@ class Converter extends React.Component {
 
       currencies.push(currencies.shift(currencies[currencyIndex]));
       this.setState({ currencies });
+      localJSONStorage.set('currency', currency);
     }
   }
 
--- a/src/components/converter/index.test.js
+++ b/src/components/converter/index.test.js
@@ -11,9 +11,12 @@ import Converter from './index';
 describe('Converter', () => {
   let explorereApiMock;
   let wrapper;
+  const storage = {};
 
   beforeEach(() => {
     explorereApiMock = sinon.stub(liskServiceApi, 'getPriceTicker').returnsPromise();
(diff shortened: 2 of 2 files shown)


karma.conf.js
```diff
@@ -66,6 +66,7 @@ module.exports = function (config) {
             'src/components/passphraseCreation/index.js',
             'src/utils/api/liskService.js',
             'src/components/errorBoundary/index.js',
+            'src/components/transactionDashboard/index.js',
```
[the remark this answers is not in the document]
slaweet: Not too sure, but since the transactionDashboard HOC is an integration point between redux and react, it looks like a missing integration test.

src/components/converter/index.js
```diff
@@ -12,14 +14,23 @@ class Converter extends React.Component {
...
+
     this.updateData();
   }
 
+  componentDidMount() {
+    const localStorageSettings = localJSONStorage.get('settings', {});
+    if (localStorageSettings.currency) {
+      this.selectActive(localStorageSettings.currency);
+    }
+  }
```
slaweet: The settings object is available in redux store, not need to read it from localStorage and store it in this component's state.

src/components/request/specifyRequest.test.js
```diff
@@ -14,6 +14,8 @@ describe('Specify Request', () => {
       address: '234l',
       nextStep: spy(),
       prevStep: spy(),
+      settings: {},
+      settingsUpdated: spy(),
```
faival: assert for spy missing

src/components/sendWritable/index.test.js
```diff
@@ -23,10 +25,21 @@ describe('SendWritableHOC', () => {
...
+    </Router>);
   });
 
   it('should render Send', () => {
     expect(wrapper.find('SendWritable')).to.have.lengthOf(1);
   });
+
+  it('should bind settingsUpdated action to SendWritable props.settingsUpdated', () => {
+    const actionsSpy = spy(settingsActions, 'settingsUpdated');
+    wrapper.find('SendWritable').props().settingsUpdated({});
```
faival: what about passing a valid settings object and asserting that the call contains the key of the object?

test/e2e/explorer.feature
```diff
@@ -8,9 +8,10 @@ Feature: Explorer page
     And I wait 2 seconds
     Then I should see text "No activity yet" in "empty message" element
     When I clear "search bar input" field
+    And I wait 2 seconds
```
faival: I think this is no needed for clearing an input field

src/components/sendWritable/index.js
```diff
@@ -9,4 +10,8 @@ const mapStateToProps = state => ({
   pendingTransactions: state.transactions.pending,
 });
 
-export default connect(mapStateToProps)(translate()(Send));
+const mapDispatchToProps = dispatch => ({
+  settingsUpdated: data => dispatch(settingsUpdated(data)),
```
slaweet: IMO it's not DRY that any component that uses the converter needs to have this. It would be DRYer if converter had it's own HOC.

PR #873, opened 2018-06-04
Fix sign message shouldn't ask for passphrase if account is unlocked - Closes #864

### LINKED ISSUE #864: Sign message shouldn't ask for passphrase if account is unlocked

michaeltomasik:
### What was the problem?
sign message shouldn't ask for a passphrase if an account is unlocked
### How did I fix it?
Added condition for checking if an account is unlocked
### How to test it?
Go to `http://localhost:8080/#/sign-message` and go through all steps while an account is unlocked.
It should not ask for password.
### Review checklist
- The PR solves #864 

### DIFF
--- a/src/components/signMessage/confirmMessage.js
+++ b/src/components/signMessage/confirmMessage.js
@@ -24,6 +24,12 @@ class ConfirmMessage extends React.Component {
     };
   }
 
+  componentWillMount() {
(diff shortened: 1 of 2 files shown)


slaweet:
Second passphrase is not necessary. It could be used, but IMO it would unnecessarily complicate both the signing and the verification process.

Issue #865, opened 2018-05-31
Integrate localStorage last searches with autosuggest component - Delegate Search (#745)


faival:
### Expected behaviour
- Currently localStorage saved searches should also appear when focusing on autosuggest searchBar and when not yet having typed a search
- Child of #745 
### Actual behaviour
- Not implemented yet

faival:
Closing as this ticket was referring to save searches used in autosuggest searches into localStorage, a separate issue is handling to show recent searches #895

Issue #835, opened 2018-05-24
Delegate search with autosuggest should be covered by unit/integration/e2e tests - Delegate Search (#745)


faival:
### Expected behaviour
- Add unit/integration/e2e tests for delegate search and autosuggest
- child of #745 

### Actual behaviour
- No implementation yet.

faival:
closing as testing is already in place:

https://github.com/LiskHQ/lisk-hub/blob/0.10.0/karma.conf.js#L56
https://github.com/LiskHQ/lisk-hub/blob/0.10.0/test/e2e/explorer.feature#L5

PR #878, opened 2018-06-04
Delegate search api integrated with autosuggest - Closes #834, #835 , #865 

### LINKED ISSUE #834: Integrate delegateSearch API with autosuggest component - Delegate Search (#745)
### Expected behaviour
- Autosuggest component should use data provided by delegate search api wrapper
- child of #745 
### Actual behaviour
- No implementation yet.

faival:
### What was the problem?
- #834
### How did I fix it?
- replaced search input in search bar with new autosuggest
- updated tests to account for change
- every time that a suggestion is used, the entity id, (account address, delegate address, transaction id) it's saved on latestSearches
- a separate ticket is created to account for suggestion input placeholder #900 
### How to test it?
- search for complete txIds, account addresses, or delegates names. try different mouse events.

### Review checklist
- The PR solves #834


faival:
Updated requirements and PR changes:
- Limit results length of delegates to an amount which makes sense for the height of the list shown
- Sort results by match of start of string 
- Autocomplete or rewrite input value to first suggestion
- Use blur and not mouse leave event to hide the dropdown

faival:
pr changes:
- filter transactions results by only having results matching start of search term
- redirect to /explorer/result on hit enter when no result was selected

faival:
seems this change has been done for explorer but not for wallet, as clicking in accountInfo still triggers transacionsFilterSet

PR #885, opened 2018-06-06
Limit linux builds to x86_64

abelboldu:
### What was the problem?

The jenkins job was creating artifacts of both 32 and 64 bits of Linux builds, taking too much disk space. Only 64bit ones are needed.

### How did I fix it?

Changing the linux builds to use `--x64` only

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -89,7 +89,7 @@ node('lisk-hub') {
         rsync -axl --delete --rsync-path="mkdir -p /var/www/test/${JOB_NAME%/*}/$BRANCH_NAME/ && rsync" $WORKSPACE/app/build/ jenkins@master-01:/var/www/test/${JOB_NAME%/*}/$BRANCH_NAME/
         npm run --silent bundlesize
         if [ -z $CHANGE_BRANCH ]; then
-          USE_SYSTEM_XORRISO=true npm run dist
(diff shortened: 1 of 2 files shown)


abelboldu:
I assumed that the 32bit build is part of the job (even if we don't need the file we need to test the build step), but @slaweet can confirm.

slaweet:
Good point @fchavant, it's not necessary to build 32 bit, we don't use it and nobody complained :-)
Can be just removed `--ia32` here:
https://github.com/LiskHQ/lisk-hub/blob/6595cdc2a7090547db8707eb2dca6add701fa2bc/package.json#L22

abelboldu:
What about the rest of the `--ia32` flags? and ARM? 
```
    "dist": "build --ia32 --x64 --publish onTag",
    "dist:win": "build --win --ia32 --x64 --publish onTag",
    "dist:mac": "build --mac",
    "dist:linux": "build --linux --ia32 --x64 --armv7l",
```

slaweet:
Good question @abelboldu 
Please keep it in `"dist:win"`, there it's all bundled into a single file.
You can remove `ia32` and `armv7l`  from `"dist:linux"`

abelboldu:
The tagged build will fail if we don't provide a GitHub token with `withCredentials`

abelboldu:
Credentials provided now @slaweet

PR #886, opened 2018-06-07
Add currency switcher to settings - Closes #845

### LINKED ISSUE #845: Add currency switcher to settings
### Expected behaviour
 - There should be an option in settings to select the preferred currency
### Actual behaviour
- not implemented yet
### Steps to reproduce

michaeltomasik:
### What was the problem?
Not implemented currency switcher on settings
### How did I fix it?
Implemented currency switcher
### How to test it?
Go to settings page, switch to ON then go to dashboard and check in transaction input if defaut currency cahnged.

### Review checklist
- The PR solves #845

### DIFF
--- a/src/components/setting/setting.js
+++ b/src/components/setting/setting.js
@@ -17,6 +17,7 @@ class Setting extends React.Component {
     super();
     this.state = {
       activeSlide: 0,
+      currencies: ['USD', 'EUR'],
     };
   }
 
@@ -56,6 +57,8 @@ class Setting extends React.Component {
       t, settings, settingsUpdated,
       hasSecondPassphrase,
     } = this.props;
+
+    const activeCurrency = settings.currency || 'USD';
 
     return (<Box className={styles.wrapper}>
       <aside>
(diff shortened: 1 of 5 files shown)


src/components/setting/setting.js
```diff
@@ -118,16 +121,21 @@ class Setting extends React.Component {
...
-        */}
+        {/* TODO: will be re-enabled when the functionality is updated */}
+        <h4>{t('Local')}</h4>
+        <div className={styles.item}>
+          <label>{t('Currency')}</label>
+          <ul className={styles.currencyList}>
+            {this.state.currencies.map(currency => (
+              <li
+                key={`currency-${currency}`}
+                className={`currency ${currency === activeCurrency ? styles.active : 'unactive'}`}
```
faival: if `unactive` is used for tests I would rename it to `currency-unactive`

src/components/setting/setting.js
```diff
@@ -57,6 +59,8 @@ class Setting extends React.Component {
       hasSecondPassphrase,
     } = this.props;
 
+    const activeCurrency = settings.currency || 'USD';
```
faival: USD value can be used from settingsConst.currencies

src/components/setting/setting.test.js
```diff
@@ -20,6 +20,7 @@ describe('Setting', () => {
   const settings = {
     autoLog: true,
     advancedMode: true,
+    currency: 'USD',
```
faival: let's use the constants also along the tests

PR #901, opened 2018-06-11
Update autosuggest input placeholder with suggested results - Closes #900

### LINKED ISSUE #900: Autosuggest should update input placeholder - Delegate Search #745
### Expected behaviour
- Input place holder should update it's value to first suggestion or every time the users moves selected result up/down in the list
- child of #745
### Actual behaviour
- No visual indicator placeholder with suggested result is shown.

faival:
### What was the problem?
- See #900 
### How did I fix it?
- Added an input with opacity which shows the full value of the suggestion when there are matches. 
- Updated the value every time the user interacts with the dropdown
### Review checklist
- The PR solves #900

### DIFF
--- a/src/components/app/variables.css
+++ b/src/components/app/variables.css
@@ -134,7 +134,8 @@ or "warn/action" ineastd of "red/green"
   --headerbar-index: 20;
   --autosuggest-index: 21;
   --searchbar-input-index: 22;
-  --accounts-index: 23;
+  --searchbar-placeholder-index: 23;
+  --accounts-index: 24;
 
   /*************************
         Margins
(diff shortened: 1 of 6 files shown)


slaweet:
Some ideas:
- Never use `overflow: scroll`, it just looks ugly on windows with the ugly window's scrollbar always visible, `overflow: auto` is better, but in this case not needed anymore for `autoSuggest` class, I think. It's already merged from a previous PR
- The placeholder suggestion should IMO turn to the input value after enter is hit
- Clicking the input field could select the text in it, so that I can easily clear it by just starting to type.
- `getValueFromCurrentIdx` looks like some refactoring of what is stored in the state is needed - maybe a follow-up refactoring ticket?

faival:
- leave autosuggest input filled with {address,transactions} id, when those suggestions are used. ✅

PR #908, opened 2018-06-12
Dont allow to call API when going on Delegate statistics page - Closes #879

### LINKED ISSUE #879: Delegate statistics sends request transactionFilterSet
### Expected behaviour
- Showing the active tab should be handled separately from setting a transactions filter
### Actual behaviour
- It triggers request with wrong filter set.

michaeltomasik:
### What was the problem?
Clicking on `Delegate Statistic` tab was triggering unnecessary API call with the wrong filter value
### How did I fix it?
Changed action for additional tabs
### How to test it?
Open `Network` console and change tab t o`Delegate Statistics` check if it triggers any API call
### Review checklist
- The PR solves #879

### DIFF
--- a/src/actions/filters.js
+++ b/src/actions/filters.js
@@ -1,7 +1,6 @@
 import actionTypes from '../constants/actions';
 // eslint-disable-next-line
-export const addFilter = ({ filterName, value }) => ({
+export const addFilter = data => ({
   type: actionTypes.addFilter,
-  filterName,
-  value,
+  data,
 });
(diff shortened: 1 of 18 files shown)


src/actions/search.js
```diff
@@ -57,6 +57,16 @@ export const searchAccount = ({ activePeer, address }) =>
     });
   };
 
+export const setSearchTranactions = searchObj => ({
```
faival: IMHO, this action is only used to set the filter, no need to pass address, transactions or count. Also the name of the action is not reflecting the change in the state.

src/actions/filters.js
```diff
@@ -0,0 +1,7 @@
+import actionTypes from '../constants/actions';
+// eslint-disable-next-line
+export const addFilter = ({ filterName, value }) => ({
```
faival: all actions use the key `data` to store the action data. Let's please use the same pattern here.

PR #910, opened 2018-06-12
Show recent searches in autosuggest - Closes #895

### LINKED ISSUE #895: Add "recent searches" to search autosuggest
### Expected behaviour
- Implement new design of recent searches in autosuggest
### Actual behaviour
- Not implemented
### Steps to reproduce

https://taikonauten.atlassian.net/browse/L20-459

faival:
### What was the problem?
- See #895 
### How did I fix it?
- Render search results based on last searches, reuse resultsList module
### How to test it?
- go to explorer or perform some valid searches with the autosuggest. 
- then focus on autosuggest but don't write any value.
### Review checklist
- The PR solves #895 

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -207,6 +207,7 @@
   "Rank": "Rank",
   "Rank / Status": "Rank / Status",
   "Receiver": "Receiver",
+  "Recent searches": "Recent searches",
   "Recipient": "Recipient",
   "Redo": "Redo",
   "Register": "Register",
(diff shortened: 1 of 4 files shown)


faival:
placeholder removal will be fixed once #900 merged to 0.10.0 and rebased.

slaweet:
The recent searches could also work with arrow down/up, but it would have the first item selected only after pressing arrow down for the first time.

But I wouldn't implement it now, we can have a follow-up after we have some feedback on how the features is actually used.

faival:
- provide keyboard navigation to recent searches, only fill placeholder on first arrowDown ✅

slaweet:
screen shot 2018-06-18 at 10 29 29
In the initial state:
- it should IMO not highlight the first item, only after the first arrow-down press.
- enter should not do anything if the input field is empty

PR #912, opened 2018-06-13
Add redux logic for followed accounts - Closes #902

### LINKED ISSUE #902: Create redux logic for followed accounts
### Expected behaviour
We should have redux logic for the followed accounts, similar to savedAccounts

### Actual behaviour
no logic existent

Parent #898

ginacontrino:
### What was the problem?
We had nothing set up for the followed accounts yet

### How did I fix it?
added actions, subscriber, middleware and reducer + tests
I might adjust some things in the following tickets

### How to test it?
read through everything  and make sure it makes sense

### Review checklist
- The PR solves #902


src/store/reducers/followedAccounts.js
```diff
@@ -0,0 +1,44 @@
...
+    case actionTypes.followedAccountsRetrieved:
+      return { accounts: Array.isArray(action.data) ? action.data : [action.data] };
+
+    case actionTypes.followedAccountAdded:
+      accounts.push(action.data);
+      return { accounts };
+
+    case actionTypes.followedAccountUpdated:
+      indexOfAccount = getIndexOfFollowedAccount(state.accounts, action.data);
+      changedAccount = action.data;
```
slaweet: This line has no effect. Can be removed

src/store/middlewares/followedAccounts.js
```diff
@@ -0,0 +1,61 @@
...
+    switch (action.type) {
+      case actionTypes.newBlockCreated:
+        checkTransactionsAndUpdateFollowedAccounts(
+          peers,
+          action.data.block.transactions,
+          followedAccounts,
+        );
+        break;
+      case actionTypes.accountLoggedIn:
+        updateFollowedAccounts(peers, followedAccounts.accounts);
```
slaweet: Currently, it is enough to `updateFollowedAccounts ` on `accountLoggedIn` but later I'd like to make the dashboard accessible without login, then it should happen on `activePeerSet `

src/store/reducers/followedAccounts.js
```diff
@@ -0,0 +1,44 @@
...
+  const accounts = [...state.accounts];
+  let indexOfAccount;
+  let changedAccount;
+
+  switch (action.type) {
+    case actionTypes.followedAccountsRetrieved:
+      return { accounts: Array.isArray(action.data) ? action.data : [action.data] };
+
+    case actionTypes.followedAccountAdded:
+      accounts.push(action.data);
```
slaweet: I was thinking about whether this should check if the account is already on the list, but I guess it is good as long as the "edit" component doesn't allow the same account twice. Then we would have a problem with using `publicKey` as a unique identifier.

Issue #831, opened 2018-05-24
Join two white boxes in account detail and delegates page


reyraa:
![screen shot 2018-05-17 at 10 19 55](https://user-images.githubusercontent.com/1254342/40165350-0e6c5cb0-59bc-11e8-8a38-56ef004a2019.png)

PR #935, opened 2018-06-21
Join two boxes in profile and delegate page

ginacontrino:
### What was the problem?
The two boxes in the profile and delegate page were not joined

### How did I fix it?
joined them:
screen shot 2018-06-21 at 13 40 55

screen shot 2018-06-21 at 14 04 47

### Review checklist
- The PR solves #831 

### DIFF
--- a/src/components/accountTransactions/accountTransactions.css
+++ b/src/components/accountTransactions/accountTransactions.css
@@ -7,7 +7,7 @@
 
 .wrapper {
   min-width: 100%;
-  box-shadow: 0px 22px 22px 0px rgba(0,0,0,.22);
+  box-shadow: var(--normal-shadow);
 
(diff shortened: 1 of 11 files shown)


src/components/accountTransactions/accountTransactions.css
```diff
@@ -7,9 +7,15 @@
 
 .wrapper {
   min-width: 100%;
+  box-shadow: 0px 22px 22px 0px rgba(0, 0, 0, 0.22);
```
slaweet: We have `var(--normal-shadow)` defined in `variables.css` that you can use here.

ginacontrino: oh thanks!

PR #936, opened 2018-06-21
Implement new design for network switcher - Closes #844

### LINKED ISSUE #844: Implement the new design of network switcher in login page
### Expected behaviour

- Network switcher should implement a design which matches rest of elements in UI

https://taikonauten.atlassian.net/browse/L20-414

Sign In
screen shot 2018-06-15 at 15 15 14

https://taikonauten.invisionapp.com/d/main/#/console/12665756/302905087/inspect
https://taikonauten.invisionapp.com/d/main/#/console/12665756/302905085/inspect
https://taikonauten.invisionapp.com/d/main/#/console/12665756/302905086/inspect
https://taikonauten.invisionapp.com/d/main/#/console/12665756/302905084/inspect
https://taikonauten.invisionapp.com/d/main/#/console/12665756/302905083/inspect
https://taikonauten.invisionapp.com/d/main/#/console/12665756/302905082/inspect

There shoud be a global network indicator - only if network switcher setting is enabled
screen shot 2018-06-15 at 15 14 38

https://taikonauten.invisionapp.com/d/main/#/console/12665756/302633917/inspect

updated iconfont @ https://drive.google.com/drive/folders/1h5HzWxbRXLlSfW-x3NnEHKgqp3R_keHR

### Actual behaviour

- Network switcher has a UI which is not aligned with current elements of hub

ginacontrino:
### What was the problem?
Network switcher still had old design

### What does it look like now?

screen shot 2018-06-22 at 16 11 49

### Review checklist
- The PR solves #844


src/components/login/login.js
```diff
@@ -146,44 +145,41 @@ class Login extends React.Component {
...
       <Box className={styles.wrapper}>
         <section className={`${styles.login} ${styles[this.state.passInputState]}`}>
           <section className={styles.table}>
             <header>
-              <a className={styles.forwardToLink} href='https://lisk.io' target='_blank' rel='noopener noreferrer'>
-                <span className={styles.label}>{this.props.t('Go to lisk.io')}</span>
-                <FontIcon className={styles.icon}>arrow-right</FontIcon>
-              </a>
+              {this.showNetworkOptions()
+                ? <div>
```
faival: I think the div is not needed if it's not adding extra styles

ginacontrino: The div is wrapping 2 components, they need some kind of wrapper

src/components/toolbox/dropdown/dropdown.css
```diff
@@ -0,0 +1,64 @@
+@import '../../app/variables.css';
+
+:root {
+  --dropdown-box-shadow: 0 8px 16px 0 rgba(29, 72, 119, 0.16);
+  --dropdown-list-hover-color: rgba(204, 217, 228, 0.2);
+}
+
+.dropdown ul {
```
faival: I think for this would be better to use the pattern:

```
.dropdown {

& ul {...}
& li, .selected {...}
& .values {...}
}
```
so that if ever we replace the name, then it's only in one place, and postcss will output the same.

src/components/toolbox/dropdown/dropdown.css
```diff
@@ -0,0 +1,64 @@
...
+.arrow {
+  position: absolute;
+  left: 95px;
+  top: 35px;
+  color: var(--color-primary-standard);
+  cursor: pointer;
+}
+
+input:-webkit-autofill {
+  -webkit-box-shadow: inset 0 0 0px 9999px var(--color-white);
```
faival: I think there's no need to auto-prefix the property

src/components/toolbox/dropdown/dropdown.css
ginacontrino: hm well i see your point, but this is not a name we gave, it's part of the react toolbox dropdown. they gave that classname, so we won't have influence on that. do you still want me to change that?

src/components/toolbox/dropdown/dropdown.css
ginacontrino: I did that because otherwise there is this ugly yellow background on auto-fill:
screen shot 2018-06-22 at 16 19 26

You don't think we should remove that background colour?

faival:
image

faival:
* also, when filling localhost:4 --> localhost:40, the position of the input seems to bounce, may be due to some extra px.

PR #938, opened 2018-06-22
Implement logout button - Closes #540

### LINKED ISSUE #540: Implement logout button
### Expected behaviour
- There should be a logout button next to the account avatar.
screen shot 2018-08-16 at 14 41 18

- The button leads to a confirmation screen.
screen shot 2018-08-15 at 13 45 33
- Clicking the account avatar no longer links to saved accounts 
- Accounts are no longer saved on login, so closing and opening the app leads to login page
- The timeout no longer locks ID, but rather logs use out. First it shows a warning 1 minute before the timeout

screen shot 2018-08-15 at 13 47 41
Then on the timeout this screen is shown:
screen shot 2018-08-15 at 13 46 50
- The "Auto-Lock" switcher in settings needs to be updated to say "Auto-logout" and work control the logout.

https://taikonauten.atlassian.net/browse/L20-411

### Actual behaviour
There is no logout button

michaeltomasik:
### What was the problem?
No Logout button
### How did I fix it?
Implemented logout button
### How to test it?
- Login to your page and click logout it should open dialog with logout confirmation
- When less than 1 min opens dashboard `Timeout in mm:ss`
- When timer reaches 0:00 opens `Session timeout` dialog
All designs are accessible via this link
https://taikonauten.atlassian.net/browse/L20-411
### Review checklist
- The PR solves #540


src/components/dialog/dialog.js
```diff
@@ -12,33 +9,33 @@ class DialogElement extends Component {
...
-    clearTimeout(this.timeout);
-    this.setState({ hidden: false });
-    this.props.dialogDisplayed({
-      title: this.props.dialog.title,
-      theme: this.props.dialog.theme,
-      childComponent: this.props.dialog.childComponent,
-      childComponentProps: parseSearchParams(this.props.history.location.search),
-    });
-  }
+  // open() {
```
faival: why not remove, and leave as commented code?

src/components/dialog/dialog.js
```diff
@@ -50,19 +47,19 @@ class DialogElement extends Component {
...
               </Navigation>
             </AppBar> :
-            null }
+            null } */}
           <div className={`modal-dialog-body ${theme.innerBody}`}>
             {this.props.dialog.childComponent ?
               <this.props.dialog.childComponent
                 {...(this.props.dialog.childComponentProps || {})}
-                closeDialog={this.goBack.bind(this)}
               /> :
```
faival: I think if we comment the if, then the else or ```: null``` should be removed?

src/components/dialog/options.js
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react';
+import grid from 'flexboxgrid/dist/flexboxgrid.css';
+import { Button, RedButton } from '../toolbox/buttons/button';
+
+import styles from './options.css';
+
+const Options = ({ firstButton, secondButton, text, title }) =>
+  (<div className={styles.optionsBody}>
+    <div className={styles.title}>{title}</div>
+    <br />
```
faival: I think is more maintainable if we add the spacings with css styles instead of <br /> markup elements

src/components/dialog/dialog.js
michaeltomasik: It is needed because in some cases it will not render properly

src/components/dialog/options.js
```diff
@@ -0,0 +1,17 @@
...
+const Options = ({ firstButton, secondButton, text }) =>
+  (<div className={styles.optionsBody}>
+    <p className={styles.text}>{text}</p>
+    <section className={`${grid.row} ${styles.buttonsRow} ${grid['between-xs']}`}>
+      <Button label={firstButton.text} onClick={firstButton.onClickHandler} className='ok-button' />
+      <RedButton label={secondButton.text} onClick={secondButton.onClickHandler} className='ok-button'/>
+    </section>
+  </div>);
+
+export default translate()(Options);
```
faival: I think is not using translate and if needed it can could be passed through props.

src/components/dialog/options.js
```diff
@@ -0,0 +1,17 @@
+import { translate } from 'react-i18next';
+import React from 'react';
+import grid from 'flexboxgrid/dist/flexboxgrid.css';
+import { Button, RedButton } from '../toolbox/buttons/button';
+
+import styles from './options.css';
+
+const Options = ({ firstButton, secondButton, text }) =>
```
faival: I think it will be clearer if these options default to expected object to be passed, to expose what the component needs to receive.
```
{ firstButton: { text: '', onClickHandler: () => {}}, ...}
```
It could be using children props, to reduce the amount of arguments passed and make the component more flexible.

src/components/header/header.js
```diff
@@ -79,23 +105,19 @@ class Header extends React.Component {
                           </div>
                           : <div></div>
                         }
+                        <div
+                          className={`${styles.logout} logout`}
+                          onClick={() => this.openLogoutDialog() }>
+                          <FontIcon value='logout' className={styles.logoutIcon} />Logout
```
faival: please use props.t to translate this Logout key.

src/components/header/header.test.js
```diff
@@ -63,4 +65,16 @@ describe('Header', () => {
     wrapper = mountWithRouter(<Header {...mockInputProps} />, options);
     expect(wrapper.find('Countdown')).to.have.length(1);
   });
+
+  it('should display logout dialog window', () => {
```
faival: Do we need a test to assert that when the dialog is open, and the user confirms to renew timeout, the dialog is then closed?

faival:
some text styles differ from designs.
screen shot 2018-06-26 at 10 26 34
untitled

faival:
- when pressing cancel, the element of the dialog hide first and then the overlay hides, leaving an empty white box visible for a while. 
- smaller breakpoints need some improvements
image

src/components/dialog/options.js
michaeltomasik: Ther is no children component involved there. These buttons are passed as props in action

src/components/header/header.test.js
michaeltomasik: I'm creating a different task for rest of tests this one is already too big

src/components/header/header.js
```diff
@@ -58,47 +75,52 @@ class Header extends React.Component {
...
+                                  closeDialog={this.props.closeDialog}
+                                  history={this.props.history}
+                                  setActiveDialog={this.props.setActiveDialog}
+                                  resetTimer={this.props.resetTimer}
+                                  autoLog={this.props.autoLog}
+                                  t={this.props.t}
+                                />
+                              </Countdown>
+                            </div>
+                            : <div></div>
```
faival: not possible to render null in the else clause?

src/store/reducers/savedAccounts.js
```diff
@@ -62,14 +62,12 @@ const savedAccounts = (state = { accounts: [] }, action) => {
     case actionTypes.removeSavedAccountPassphrase:
       return {
         ...state,
-        accounts: state.accounts.map((account) => {
-          if (!action.data ||
-            (action.data.peerAddress === account.peerAddress &&
-            action.data.passphrase === account.passphrase)) {
+        accounts: [].concat(state.accounts.map((account) => {
```
faival: this is removing the peerAddress check, what about customNetwork with different addresses? I think all of their passphrase will be removed.

slaweet:
@michaeltomasik I like that refreshing the page keeps me logged in. But it also keeps me logged in if I log out and then refresh.

Efefefef:
dashbArd ⌨️ 

image

Efefefef:
Is it okay to have Explorer listed as possible activity in logout message ? We do not have this module yet

Efefefef:
🐛 As discussed with Vit , 'Explorer' should be substituted for 'Search'

Efefefef:
🐛 I know it was not relevant to PR but see extended specs #540 
The "Auto-Lock" switcher in settings needs to be updated to say "Auto-logout" and work control the logout. "Lock ID" near the avatar should be "Timeout"

slaweet:
It shouldn't say`Timeout` but rather `Session timeout`. 
At first, I've put a wrong screenshot to the issue description. Now I fixed it.

Efefefef:
🎉 Yeeah

PR #953, opened 2018-06-25
Redirect to valid entity results page when no autosuggestion result used - Closes #946

### LINKED ISSUE #946: Autosuggest should redirect properly when not using suggested results - Delegate search #745
### Expected behaviour
When no suggestions present, autosuggest should:
```
get current value typed in input
match against existing patters, {address, transactions}
if match
 use proper url {/explorer/account/address, /explorer/account/address?id=transactionId}
else
 use /explorer/result/inputVal

```

### Actual behaviour
When quick C&P an address in autosuggest field, and hitting immediately afterwards Enter key, autosuggest understands is a non existing record, and redirects to /explorer/result/. 
### Steps to reproduce

- child of #745

faival:
### What was the problem?
- using a valid address, or transaction id, was not redirecting properly to the entity results page. 

### How did I fix it?
- add pattern matching and reuse redirection method in case of match
### How to test it?
- copy a valid {transactionId, addressId}, really quickly paste it on the autosuggest field and immediately press enter without giving time for the search to be triggered. 
- the result should be the same, as when suggestion is shown and a click is performed on it. 
### Review checklist
- The PR solves #946 

### DIFF
--- a/src/components/autoSuggest/index.js
+++ b/src/components/autoSuggest/index.js
@@ -9,6 +9,7 @@ import keyCodes from './../../constants/keyCodes';
 import localJSONStorage from './../../utils/localJSONStorage';
 import regex from './../../utils/regex';
 import { saveSearch } from './../search/keyAction';
+import { searchEntities } from './../../constants/search';
 
 class AutoSuggest extends React.Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


michaeltomasik:
How To test it ?

src/components/autoSuggest/index.js
```diff
@@ -81,8 +81,18 @@ class AutoSuggest extends React.Component {
   }
 
   submitAnySearch() {
-    this.inputRef.blur();
-    this.props.history.push(`${routes.searchResult.pathPrefix}${routes.searchResult.path}/${encodeURIComponent(this.state.value)}`);
+    let searchType = null;
+    if (this.state.value.match(regex.address)) {
+      searchType = 'addresses';
+    } else if (this.state.value.match(regex.transactionId)) {
+      searchType = 'transactions';
```
michaeltomasik: Please put this strings to constants file

PR #986, opened 2018-07-06
Add reference field - Closes #813

### LINKED ISSUE #813: Add support for reference field in transactions
### Expected behaviour
Lisk core 1.0.0 will add support for optional data field https://github.com/LiskHQ/lisk/issues/26
Lisk Hub should have a field in the send module to support it and display it in transactions list.

Design: https://taikonauten.invisionapp.com/d/main/#/console/12665756/284010013/preview

**should also be part of the launch protocol.**

### Actual behaviour
There is no support for reference field.

michaeltomasik:
### What was the problem?
No reference field in Transferbox
### How did I fix it?
Added Reference field
### How to test it?
Go to Wallet and make transaction with reference field and check if transaction have referance on transactionDetailView
### Review checklist
- The PR solves #813

### DIFF
--- a/src/components/sendWritable/send.js
+++ b/src/components/sendWritable/send.js
@@ -64,7 +64,7 @@ class SendWritable extends React.Component {
     if (!value && required) {
       return this.props.t('Required');
     } else if (name === 'reference' && byteCount > 64) {
-      return this.props.t('Max 64 bytes message');
+      return this.props.t('Maximum length exceeded');
     } else if (!value.match(this.inputValidationRegexps[name])) {
       return name === 'amount' ? this.props.t('Invalid amount') : this.props.t('Invalid address');
     } else if (name === 'amount' && value > parseFloat(this.getMaxAmount())) {
--- a/src/components/sendWritable/send.js
+++ b/src/components/sendWritable/send.js
@@ -8,6 +8,7 @@ import fees from './../../constants/fees';
 import styles from './sendWritable.css';
 import regex from './../../utils/regex';
 import AddressInput from './../addressInput';
+import ReferenceInput from './../referenceInput';
 
 class SendWritable extends React.Component {
   constructor(props) {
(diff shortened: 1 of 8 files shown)


src/components/sendWritable/send.js
```diff
@@ -37,24 +41,30 @@ class SendWritable extends React.Component {
...
   }
 
-  validateInput(name, value) {
-    if (!value) {
+  validateInput(name, value, required) {
+    const byteCount = encodeURI(value).split(/%..|./).length - 1;
+    if (!value && required) {
       return this.props.t('Required');
+    } else if (name === 'reference' && byteCount > 64) {
+      return this.props.t('Max 64 bytes message');
```
slaweet: I would say `Maximum length exceeded` in order not to confuse users with what bytes means

src/utils/api/transactions.js
```diff
@@ -30,10 +30,12 @@ export const getTransactions = ({
 export const getSingleTransaction = ({ activePeer, id }) => activePeer.transactions.get({ id });
 
 export const unconfirmedTransactions = (activePeer, address, limit = 20, offset = 0, sort = 'timestamp:desc') =>
-  activePeer.node.getTransactions('unconfirmed', {
-    senderId: address,
-    recipientId: address,
-    limit,
-    offset,
-    sort,
+  new Promise(() => {
```
slaweet: Why are you wrapping it with `new Promise`? `activePeer.node.getTransactions` already returns a promise. 
The pending transactions stay in the list, but not sure if it's related to this.

michaeltomasik: I had a problem in one spot not sure now which one, I would have to investigate it and let you know maybe it is already fixed with a current changes.

slaweet: Wrapping the promise here means that it never gets resolved, so you'll prevent if there was some error, but some feature will be broken.

PR #1026, opened 2018-07-12
Show reference field on confirmation screen - Closes #1018

### LINKED ISSUE #1018: Send: reference field not showing on confirmation screen
### Expected behaviour

All input fields for transactions should show on the confirmation screen.

### Actual behaviour

Reference field disappears.

screen shot 2018-07-10 at 16 47 58    screen shot 2018-07-10 at 16 48 01

michaeltomasik:
### What was the problem?
No reference field on a confirmation screen
### How did I fix it?
added reference field
### How to test it?
Send transactoin with reference field and check if it is on confirmation page
### Review checklist
- The PR solves #1018

### DIFF
--- a/src/components/sendReadable/send.js
+++ b/src/components/sendReadable/send.js
@@ -119,6 +119,15 @@ class SendReadable extends React.Component {
               onChange={this.handleChange.bind(this, 'recipient')}
               disabled={true}
             />
+            {this.state.reference.value ?
+              <Input label={this.props.t('Reference')}
+                className={`reference ${styles.disabledInput}`}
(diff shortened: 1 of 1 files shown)


src/components/sendReadable/send.js
```diff
@@ -119,6 +119,15 @@ class SendReadable extends React.Component {
               onChange={this.handleChange.bind(this, 'recipient')}
               disabled={true}
             />
+            {this.state.reference.value ?
```
faival: Can't we just use some p> with styles?
also, name `reference` for the Junit tests is a bit generic

PR #1041, opened 2018-07-13
Fix main menu to have always enabled Settings item - Closes #1036

### LINKED ISSUE #1036: Settings menu item disabled when not logged in and delegates section enabled
### Expected behaviour
Settings menu item should be the only one enabled without login
### Actual behaviour
Sidechains menu item is enabled instead of settings when delegates feature is enabled.

### Steps to reproduce
1. go to lisk://settings
2. enable delegates advanced feature
3. go to login page
4. sidechains is the only enabled menu item

slaweet:
### What was the problem?
The enabled menu item was defined as the one at index 3, which resulted in #1036

### How did I fix it?
I added `enabledWhenNotLoggedIn` property to settings menu item object to control if it's enabled.

### How to test it?
Follow steps in #1036

### Review checklist
- The PR solves #1036

### DIFF
--- a/src/components/mainMenu/mainMenu.js
+++ b/src/components/mainMenu/mainMenu.js
@@ -48,7 +48,7 @@ class MainMenu extends React.Component {
   navigate(history, tabs, index) {
     if (!isCurrent(history, index, tabs)) {
       this.setState({ active: false, index });
-      history.replace(tabs[index].route);
+      history.push(tabs[index].route);
     }
   }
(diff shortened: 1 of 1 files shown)


slaweet:
@faival thanks for your feedback, it made me look more thoroughly on the code. 
Disabling current menu item didn't work before either, so now I removed it altogether. More info in commit messages of the two commits above.

PR #1055, opened 2018-07-19
Fix passphrase saved in localstorage - Closes #1054

### LINKED ISSUE #1054: Passphrase saved in localhost
### Expected behaviour
Passphrase cant be store in localstorage
### Actual behaviour
passphrase is stored in localhost
### Steps to reproduce
Login on your account and check localstorage is passphrase is saved there

michaeltomasik:
### What was the problem?
#1054
### How did I fix it?
Added condition to save only network changes
### How to test it?
Login on your account and check localstorage is passphrase is saved there

### Review checklist
- The PR solves #1054

### DIFF
--- a/src/components/login/login.js
+++ b/src/components/login/login.js
@@ -124,7 +124,9 @@ class Login extends React.Component {
       ...validator(value, error),
     });
 
-    this.props.settingsUpdated({ [name]: value });
+    if (name === 'network') {
+      this.props.settingsUpdated({ network: value });
+    }
   }
 
   onFormSubmit(event) {


alepop:
Good 👍 But I think that you also should remove previously saved passphrase from localStorage.
Maybe add a schema for data that will be stored in localStorage and remove fields that don't exist in this schema.

slaweet:
@alepop thanks for the feedback.
Right now, I would argue that we don't need to worry about removing previously saved passphrases, because this bug was only introduced in 1.0.0-rc.1, which doesn't work with mainnet, so there should be no mainnet passphrase saved in any user. 

For the future, yes, we should have a schema validation of what settings are loaded from localStorage.

PR #1064, opened 2018-07-20
Fix reducers unit tests migration - Closes #1011

### LINKED ISSUE #1011: Settings, toaster, transaction, transactions, voting, Reducers Unit Tests should be covered - Core 1.0.0 Migration

michaeltomasik:
### What was the problem?
test not migrated to 1.0.0

### Review checklist
- The PR solves #1011

### DIFF
--- a/src/store/reducers/transaction.js
+++ b/src/store/reducers/transaction.js
@@ -7,7 +7,7 @@ const transaction = (state = {}, action) => {
     case actionTypes.transactionLoaded:
       return {
         votesName: state.votesName || {},
-        success: action.data && action.data.senderId,
+        senderId: action.data && action.data.senderId,
         ...action.data,
       };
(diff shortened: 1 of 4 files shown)


src/store/reducers/transaction.js
```diff
@@ -7,7 +7,7 @@ const transaction = (state = {}, action) => {
     case actionTypes.transactionLoaded:
       return {
         votesName: state.votesName || {},
-        success: action.data && action.data.senderId,
+        senderId: action.data && action.data.senderId,
```
slaweet: Changing `success` to `senderId` in the reducer and it's unit test and not anywhere else in the application logic means one of these two options:
- `success` was not used anywhere, so it can be removed here.
- `success` was used somewhere and that somewhere is now broken.
Either way, there is a problem.

PR #1077, opened 2018-07-23
Enable non-api utils unit tests - Closes #1075

### LINKED ISSUE #1049: Accounts: You can have listed 2 accounts of the same address in devnet
### Steps to reproduce

1. Login in devnet
1. Login in devnet with same address one more time
1. go to /accounts

### Actual result
There are 2 accounts listed in the same time
In localstorage there is a diff:
 first with "peerAddress":"localhost:4000"
 second with "address":"localhost:4000" 

### Expected result
There should be only one acc in a list of accs

image

### LINKED ISSUE #1075: Fix unit tests of non-api utils
### Expected behaviour
Unit tests of non-API utils should be enabled
### Actual behaviour
Unit tests of non-API utils are disabled
https://github.com/LiskHQ/lisk-hub/blob/d83113b83e3cc43eded77bd9a87d704837ebd266/src/testFiles.js#L2-L14
https://github.com/LiskHQ/lisk-hub/blob/d83113b83e3cc43eded77bd9a87d704837ebd266/src/testFiles.js#L21-L22

slaweet:
### What was the problem?
See #1075

In addition to that, saved account util was using inconsistently `address` and `peerAddress` and wasn't backward compatible. Fixing this closes #1049.

### Review checklist
- The PR solves #1075


src/utils/savedAccounts.js
```diff
@@ -11,36 +11,47 @@ const isValidSavedAccount = ({ publicKey, network, peerAddress }) => {
   }
 };
 
+const fixBackwardsCompatibility = (account) => {
+  // 'address' property was renamed to 'peerAddress' to avoid confusion with account address
+  if (account && account.address) {
```
faival: wouldn't it be enough to just check for address property?

```
if (account.address) {
```

slaweet: I'm trying to be defensive because `fixBackwardsCompatibility` is called before `isValidSavedAccount`

PR #1119, opened 2018-07-30
Fix the logic to disable checking pending transactions - Closes #1029

### LINKED ISSUE #1029: Pending transaction appears/disappears/reappears after a while - Core 1.0.0 Migration
### Expected behaviour
- child of #812 

### Actual behaviour
- When sending a transaction, the pending one stays for a while, then disappears from the list and suddenly is shown again.

slaweet:
### What was the problem?
The condition to disable checking pending transactions was wrong.
The original PR tried to disable one function that checks if there are unconfirmed transactions on the server. The problem was that Core is telling us that after the new transactions is forged into a block, there are no unconfirmed transactions, but it doesn't show the new transactions in the confirmed transactions either.

### How did I fix it?
Now the condition is always false, so that `transactionsUpdateUnconfirmed` is never dispatched, so  it doesn't try to check unconfirmed transactions. The effect that a transaction stays in the `pending` state until it is found in the confirmed transactions. So we don't have disappearing transactions anymore. 

### Review checklist
- The PR solves #1029

### DIFF
--- a/src/actions/transactions.js
+++ b/src/actions/transactions.js
@@ -188,8 +188,8 @@ export const transactionsUpdated = ({
           type: actionTypes.transactionsUpdated,
         });
         // eslint-disable-next-line no-constant-condition
-        if (pendingTransactions.length || true) {
-          // "|| true" above was added to disable this, because this caused pending transactions
+        if (pendingTransactions.length && false) {
+          // "&& false" above was added to disable this, because this caused pending transactions
           // to disappear from the list before they appeared again as confirmed.
           // Currently, the problem is that a pending transaction will not be removed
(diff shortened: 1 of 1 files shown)


src/actions/transactions.js
```diff
@@ -188,8 +188,8 @@ export const transactionsUpdated = ({
           type: actionTypes.transactionsUpdated,
         });
         // eslint-disable-next-line no-constant-condition
-        if (pendingTransactions.length || true) {
-          // "|| true" above was added to disable this, because this caused pending transactions
+        if (pendingTransactions.length && false) {
```
faival: why not commenting the whole section and moving the TODO at the top of the comment and linking the core 1.0.0 issue?

PR #1120, opened 2018-07-31
Hide Fee field on Request transfer - Closes #1052

### LINKED ISSUE #1052: Request transfer: While entering amount "fee" is shown
### Steps to reproduce

1. Got to wallet/dashboard
1. Choose Request tab
1. Request specific transfer

### Actual result
I see the amount of "Fee"
It makes no sence to show requesting specific amount
it makes me think either should i deduct it or add up to the amount
or it makes me think that i should pay it to request 

### Expected result
I don't see any fee

image

michaeltomasik:
### What was the problem?
Fee field was show on requesting amount
### How did I fix it?
Hide fee field for request amount

### Review checklist
- The PR solves #1052

### DIFF
--- a/src/components/request/specifyRequest.js
+++ b/src/components/request/specifyRequest.js
@@ -68,6 +68,7 @@ class SpecifyRequest extends React.Component {
             value={this.state.amount.value}
             onChange={this.handleChange.bind(this, 'amount')}
             t={this.props.t}
+            isRequesting
           />
         </div>
         <footer>
--- a/src/components/converter/converter.js
+++ b/src/components/converter/converter.js
@@ -84,7 +84,7 @@ class Converter extends React.Component {
             : <div></div>
           }
         </div>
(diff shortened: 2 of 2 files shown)


src/components/request/specifyRequest.js
```diff
@@ -68,6 +68,7 @@ class SpecifyRequest extends React.Component {
             value={this.state.amount.value}
             onChange={this.handleChange.bind(this, 'amount')}
             t={this.props.t}
+            isRequesting
```
faival: can't find the value of `isRequesting` in any line along `specifyRequest.js` can you clarify where is coming from or how is initialised?

michaeltomasik: This is short for `isRequesting = true` so it works like switch between Requesting and sending module

faival: all clear 🌮

PR #1122, opened 2018-07-31
Hide Added Removed votes for creation of secound passphrase - Closes #1060

### LINKED ISSUE #1060: Second passphrase tx details: Unnecessary vote fields
### Steps to reproduce

1. Find and open any Second passphrase tx details

### Actual result
+ Unnecessary "Added votes" and "Remove votes" fields
+ Inconsistent naming "Signature" instead of "Passphrase" 

image

michaeltomasik:
### What was the problem?
Wrong fields `added votes` & `removed votes` for creating the second passphrase
### How did I fix it?
Hide it for transaction type 1

### Review checklist
- The PR solves #1060


src/components/transactions/transactionDetailView.js
```diff
@@ -154,8 +154,9 @@ class TransactionsDetailView extends React.Component {
           </TransactionDetailViewRow>
 
           <TransactionDetailViewRow shouldShow={
-            this.props.transaction.amount === '0' &&
-            this.props.transaction.recipientId}>
+            (this.props.transaction.amount === '0' &&
+            this.props.transaction.recipientId &&
+            this.props.transaction.type !== 1) }>
```
faival: please use available constants for tx's so that is clear which type it is

slaweet: Also, the other two conditions (amount, recipientID) seem to be redundant if you already check it is vote transaction type

src/components/transactions/transactionDetailViewRow.js
```diff
@@ -6,7 +6,7 @@ import styles from './transactionDetailView.css';
 class TransactionDetailViewRow extends React.Component {
   render() {
     const { children, shouldShow } = this.props;
-    return ((shouldShow === null || shouldShow === false) ? null :
+    return ((shouldShow === null || shouldShow === false || shouldShow === '') ? null :
```
faival: I think this does the same:
```
(!shouldShow)
```

michaeltomasik: `! shouldShow` will also cover `undefined` and We cant do it

PR #1137, opened 2018-08-06
Dashboard accessible without login - Quick Tip Module - Closes #926

### LINKED ISSUE #926: Make dashboard accessible without login
Create a Quick Tip module that is accessible on dashboard when you are not logged in.
https://taikonauten.invisionapp.com/d/main#/console/12665756/311970510/inspect

### Step to reproduce:
1. Go to dashboard
2. Instead of transferList there should be Quick Tip module

michaeltomasik:
### What was the problem?
The dashboard was not accessible without login
### How did I fix it?
Added Quick Tip Module
### How to test it?
Go to Dashboard when logout

### Review checklist
- The PR solves #926


src/components/mainMenu/mainMenu.test.js
```diff
@@ -99,11 +99,10 @@ describe('MainMenu', () => {
     expect(wrapper.find(Tab)).to.have.lengthOf(5);
   });
 
-  it('should all Tab be disabled except Settings if !props.account.address', () => {
+  it('should all Tab be disabled except Settings and Dashboard if !props.account.address', () => {
     const wrapper = mount(<MemoryRouter>
       <MainMenu {...props} account={{}} />
     </MemoryRouter>, options);
-    expect(wrapper.find(Tab).at(0).props().disabled).to.be.equal(true);
```
faival: can we assert that is not disabled?

slaweet:
Some comments:
- In the build, images are not loaded correctly https://jenkins.lisk.io/test/lisk-hub/PR-1137/#/dashboard IMO it's because we have webpack to inline small svgs (so far all of them) but not big svgs, so the big ones need to be copied into the right location and referenced with right relative paths

- If it's not too hard, it would be cool to use the transitionWrapper component we have to animate next/previous switch
- Padding-top too small.

faival:
can we add some styles to the svg so that it looks more centered as in the designs?
image

image

slaweet:
The text color should be `#3C5068`, looks like it's just black

src/components/dashboard/index.js
```diff
@@ -66,9 +74,9 @@ class Dashboard extends React.Component {
           </div>
         </div>
       </div>
-      <div className={`${grid['col-md-4']} ${styles.sendWrapper}`}>
+      {isLoggedIn ? <div className={`${grid['col-md-4']} ${styles.sendWrapper}`}>
         <Send {...this.props} />
-      </div>
+      </div> : <div></div>}
```
faival: null ?

Efefefef:
All links to the academy in Quick Tip module should open in a new tab I guess?
The initial load is not smooth - you can see the pagination on top of the module for a half of second

slaweet:
screen shot 2018-08-14 at 16 11 29
The title “Why should I vote” has smaller font-size and margin/padding-top then title “Following”

slaweet:
The animation on switching steps changes the content and image too early, so it isn't smooth

Efefefef:
Nice

PR #1142, opened 2018-08-07
Update design network indicator - Closes #1103

### LINKED ISSUE #1103: Update design of network indicator
### Expected behaviour

https://taikonauten.invisionapp.com/d/main/#/console/12665756/302633917/inspect
https://taikonauten.invisionapp.com/d/main/#/console/12665756/302633916/inspect

Icons are included in the latest font https://drive.google.com/drive/folders/1h5HzWxbRXLlSfW-x3NnEHKgqp3R_keHR

### Actual behaviour
- not implemented

faival:
### What was the problem?
- #1103 
### How did I fix it?
- refactor styles for header and use flex instead of floating, change order of render html to be more according to document flow.
### How to test it?

- login with mainnet url by customNode dropdown (should show bright, and say connected to mainnet)
- login with testnet url by customNode dropdown (should show with opacity, and say connected to testnet)
- login to devnet url by customNode dropdown (should show with opacity, and say connected to devnet)

- in all cases URL of peer must be shown
- should show only if enabled in settings

### Review checklist
- The PR solves #1103

### DIFF
--- a/test/e2e/login.feature
+++ b/test/e2e/login.feature
@@ -25,7 +25,7 @@ Feature: Login
     And I select option no. 3 from "network" select
     And I click "login button"
     Then I should be logged in as "genesis" account
-    And I should see text "testnet" in "peer network" element
+    And I should see text "Connected to testnet" in "peer network" element
     And I click "joyride-tooltip__button--skip"
     And I click "joyride-tooltip__button--skip"
     When I click "transactions" menu
@@ -40,7 +40,7 @@ Feature: Login
     When I fill in "https://testnet.lisk.io" to "address" field
     And I click "login button"
     Then I should be logged in as "genesis" account
-    And I should see text "custom node" in "peer network" element
+    And I should see text "Connected to devnet" in "peer network" element
     And I click "joyride-tooltip__button--skip"
     And I click "joyride-tooltip__button--skip"
     When I click "transactions" menu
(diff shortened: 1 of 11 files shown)


slaweet:
We are really missing written specification on this one. And it's not easy to test the Mainnet scenario.

Some points that should have been in the specification:
- The network indicator is only displayed if network switcher is enabled in settings
- "Devnet" doesn't equal "Custom node", it should depend on nethash of the node, so if I use any mainnet/testnet url of one of the hundreds of nodes in the network as a custom node, it will say mainnet/testnet, not devnet.
- White color is for mainnet, gray color for everything else
- I feel like it shouldn't "jump" on hover, but I see it is like that in the designs, so ok for now...

slaweet:
I'd do this refinement to the specification:
- Show the address not just on hover, but always - it's been quite useful in the past when users were sending us screenshots when having some issues.

test/e2e/login.feature
```diff
@@ -34,13 +36,15 @@ Feature: Login
...
+    When I go to "/setting"
+    And I click "showNetwork"
     When I'm on login page
     Then I fill in passphrase of "genesis" to "passphrase" field
     And I select option no. 4 from "network" select
     When I fill in "https://testnet.lisk.io" to "address" field
     And I click "login button"
     Then I should be logged in as "genesis" account
-    And I should see text "custom node" in "peer network" element
+    And I should see text "Connected to testnet" in "status" element
```
slaweet: The className could be more specific than `status`. Maybe `network-status`?

PR #1136, opened 2018-08-03
Enable mocha jest parallel stages on jenkins - Closes #1133

faival:
### What was the problem?
- #1133 
### How did I fix it?
WIP
### Review checklist
- The PR solves #1133 
- The PR follows our [Test guide](/LiskHQ/lisk-hub/blob/development/docs/TEST_GUIDE.md)
- The PR follows our [CSS guide](/LiskHQ/lisk-hub/blob/development/docs/CSS_GUIDE.md)

PR #1143, opened 2018-08-08
Strategy to migrate mocha to jest - Closes #1134

### LINKED ISSUE #1134: Formalise strategy for Mocha to Jest migration
### Expected behaviour

- both stages run while migrating.
- components which require modifications are upgraded to Jest

- possible drawback of increasing build time while migrating
```
- Mocha excludes tests which are run by Jest stage
- Jest excludes tests which are run by Mocha stage
- Once Mocha excludes all tests the stage can be removed
```

faival:
### What was the problem?
#1134 
### How did I fix it?
- leaving mocha suite as it was, 
- providing jest configuration
- providing jest jenkins test stage
- enabling jest suites for some components
- fixing src base to allow tests to pass in both suites
### How to test it?
- go to build and verify jest runs a subset of the mocha suites, and both pass
### Review checklist
- The PR solves #1134 


faival:
https://github.com/LiskHQ/lisk-hub/pull/1136#issuecomment-411382411

package.json
```diff
@@ -226,5 +233,41 @@
       "protractor-cucumber-framework",
       "postcss-cssnext"
     ]
+  },
+  "jest": {
```
slaweet: I would rather put the config into a separate `jest.config.js` file. Then we have the flexibility of javascript over just JSON (e.g. for comments).

package.json
```diff
@@ -226,5 +233,41 @@
...
+    "modulePaths": ["<rootDir>/src/components"],
+    "testMatch": [
+      "<rootDir>/src/components/transactions/**/?(*.)+(spec|test).js?(x)",
+      "<rootDir>/test/**/?(*.)+(spec|transactionID.test).js?(x)"
```
slaweet: I would try to make the regexps here as specific as possible to make it easy to understand.  
Something like:
```
"<rootDir>/test/integration/transactionID.test.js?(x)"
```
BTW what is the `?(x)` in the end for?

package.json
```diff
@@ -226,5 +233,41 @@
       "protractor-cucumber-framework",
       "postcss-cssnext"
     ]
+  },
+  "jest": {
+    "modulePaths": ["<rootDir>/src/components"],
+    "testMatch": [
```
slaweet: the `testMatch` could contain commented out a list of all the test files, so that we can easily see the progress of the migration. Enabling a new test in jest would be just uncommenting it here.

slaweet:
Also, test documentation https://github.com/LiskHQ/lisk-hub/blob/development/docs/TEST_GUIDE.md  should be updated to say that we are in the middle of a migration to Jest and what is the current process.

package.json
faival: optional to match jsx files

package.json
faival: couldn't make jest config to reside inside /config folder as jest coverage messes up with the rootDir, and tries to find coverage inside /config folder.

faival:
changes applied.

PR #1168, opened 2018-08-14
Add reference to request - Closes #1154

### LINKED ISSUE #1154: Add reference field to specific amount LSK request
Add Reference Field to Request Lisk Amount
Requirements:
- Reference field validation check
- Reference field should appear in the confirmation step
- Reference field value should be added to url on confirmation step

michaeltomasik:
### What was the problem?
#1154 

### Review checklist
- The PR solves #1154

### DIFF
--- a/src/components/request/confirmRequest.js
+++ b/src/components/request/confirmRequest.js
@@ -21,7 +21,6 @@ class ConfirmRequest extends React.Component {
     } = this.props;
     let link = `lisk://wallet?recipient=${address}&amount=${amount}`;
     link = reference ? `${link}&reference=${encodeURIComponent(reference)}` : link;
-    link = encodeURI(link);
 
     return (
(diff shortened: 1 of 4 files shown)


slaweet:
It should url-encode the reference, because when I fill in, for example, something with spaces, like `1 LSK for me`, the link won't work.

src/components/request/confirmRequest.js
```diff
@@ -21,6 +21,7 @@ class ConfirmRequest extends React.Component {
     } = this.props;
     let link = `lisk://wallet?recipient=${address}&amount=${amount}`;
     link = reference ? `${link}&reference=${reference}` : link;
+    link = encodeURI(link);
```
slaweet: I would rather use `reference=${encodeURIComponent(reference)}` that way it also encodes/allows chars like `&`

PR #1170, opened 2018-08-14
Use cache function from library

abelboldu:
### What was the problem?

We need to use the cache function provided by the shared library

### How did I fix it?

Changing the inline bash core for the groovy function from lisk-jenkins

### How to test it?

https://jenkins.lisk.io/job/lisk-hub/job/use_library_cache/ (failed but passed the cache part)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,18 +16,12 @@ node('lisk-hub') {
 
     stage ('Install npm dependencies') {
       try {
+        cache_file = restoreCache("package.json")
         sh '''
-        cache_file="$HOME/cache/$( sha1sum package.json |awk '{ print $1 }' ).tar.gz"
-        if [ -f "$cache_file" ]; then
-            tar xf "$cache_file"
-        fi
         npm install
         ./node_modules/protractor/bin/webdriver-manager update
-        if [ ! -f "$cache_file" ]; then
-            GZIP=-4 tar czf "$cache_file" node_modules/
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -16,18 +16,9 @@ node('lisk-hub') {
 
     stage ('Install npm dependencies') {
       try {
-        sh '''
-        cache_file="$HOME/cache/$( sha1sum package.json |awk '{ print $1 }' ).tar.gz"
-        if [ -f "$cache_file" ]; then
-            tar xf "$cache_file"
-        fi
-        npm install
-        ./node_modules/protractor/bin/webdriver-manager update
```
fchavant: I think the build is failing because this line was removed.

abelboldu: Oops, missed that line, sorry.

Jenkinsfile
```diff
@@ -16,18 +16,12 @@ node('lisk-hub') {
...
     stage ('Install npm dependencies') {
       try {
+        cache_file = restoreCache("package.json")
         sh '''
-        cache_file="$HOME/cache/$( sha1sum package.json |awk '{ print $1 }' ).tar.gz"
-        if [ -f "$cache_file" ]; then
-            tar xf "$cache_file"
-        fi
-        npm install
+        npm install --verbose
```
slaweet: @abelboldu Now that Jenkins is green, do we need `--verbose` here?

PR #1175, opened 2018-08-15
Fix order of menu items, and i18n strings for onboarding - Closes #1174

### LINKED ISSUE #1174: Wrong order of items in main menu and wrong i18n strings in onboarding
### Expected behaviour
- Wallet menu entry should be before Delegates in document order.
- Onboarding item for settings should not longer say `and repeat the onboarding here`, because it was removed from settings.

faival:
### Review checklist
- The PR solves #1174 

### DIFF
--- a/src/components/mainMenu/mainMenu.js
+++ b/src/components/mainMenu/mainMenu.js
@@ -105,7 +105,7 @@ class MainMenu extends React.Component {
     ];
 
     if (showDelegate) {
-      tabs.splice(tabs.length - 2, 0, {
+      tabs.splice(tabs.length - 1, 0, {
         label: t('Delegates'),
         id: 'delegates',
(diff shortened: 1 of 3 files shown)


src/components/mainMenu/mainMenu.js
```diff
@@ -105,7 +105,7 @@ class MainMenu extends React.Component {
     ];
 
     if (showDelegate) {
-      tabs.splice(tabs.length - 2, 0, {
+      tabs.splice(tabs.length - 1, 0, {
```
slaweet: It's not the first time we have a problem with the index here, I would rather put delegates item inside `tabs` array and `.filter` it by `id === 'delegates'` instead of `splice`ing based on the index.

faival: 👍

PR #1196, opened 2018-08-21
Goal `e2e test` should fail with error message in case `dev` goal was not started-Closes #881

### LINKED ISSUE #881: Goal `e2e-test` should fail with error message in case `dev` goal was not started
### Expected behaviour
- If a developer runs `npm run e2e-test` without prior having ran `npm run dev` the goal should fail giving a comprehensive error message.
### Actual behaviour
- No comprehensive error message is shown.

Efefefef:
### What was the problem?
No comprehensive error message is not shown If a developer runs npm run e2e-test without prior having ran npm run dev and local lisk-core is not running

### How did I fix it?
Implement bash script to check both servers responses prior to running e2e tests 

### Review checklist
- The PR solves #881


slaweet:
This solution won't work if I override basURL (e.g.`npm run e2e-test -- --params.baseURL http://localhost:8088`). The whole problem will be easy to fix when migrating the e2e tests to Cypress. So closing this PR

PR #1214, opened 2018-08-24
Fix balance not reflecting send transaction made - Closes #1188

### LINKED ISSUE #1188: Following: Balance doesn't reflect sent transactions made
### Steps to reproduce

1. Add any account as a following one
1. Go to dashboard, remember the balance
1. Send some lisk to that account
1. Wait for transaction confirmation on dashboard page
1. Check out the balance

### Actual result
Balance has not changed

### Expected result
Balance has changed

michaeltomasik:
### What was the problem?
#1188
### Review checklist
- The PR solves #1188


faival:
no as expected

slaweet:
@Efefefef @michaeltomasik  If the solution doesn't fix all scenarios, then it is not sufficient. Not describing it in PR description also is not sufficient.

AFAIK, the problem is that the new block websocket tells us that there is a new transaction, but the RESTfull API doesn't have the new transaction yet. If that is the problem, then a solution that doesn't require to poll all the followed accounts all the time would be to set a timestamp when a followed account had a transaction in the websocket block, and poll that account for a minute or two.

slaweet:
Close in favor of https://github.com/LiskHQ/lisk-hub/pull/1292

PR #1221, opened 2018-08-27
Autosuggest should show error message when no results - Closes #1193

### LINKED ISSUE #1193: Autosuggest should show error message when no results found.
### Expected behaviour

In the cases where activePeerSet was not called, calls to autosuggest search will respond with failure.

In such cases a default 'No results found' should appear in the dropdown. 

Alternatively find a better way to distinguish: 

- peers connectivity down
- no results found

faival:
### What was the problem?
- see #1193 
### How did I fix it?
- always rejecting promises in case activePeer is not set
- show a i18n result key when no results are found
### How to test it?

- while being connected to a peer, try to find a non existing delegate name
- while not being connected to a peer, try to find any delegate name, tx id, or address id. 

### Review checklist
- The PR solves #1193 

### DIFF
--- a/src/components/autoSuggest/index.js
+++ b/src/components/autoSuggest/index.js
@@ -335,6 +335,26 @@ class AutoSuggest extends React.Component {
             />
             : null
           }
+
+          { this.state.value !== '' && this.state.resultsLength === 0 ?
+            <ResultsList
+              key='no-results'
+              results={[{
+                id: 'no-results',
+                valueLeft: t('No results found'),
+                valueRight: '',
+                isSelected: false,
+                type: 'no-results',
+              }]}
+              header={{
+                titleLeft: '',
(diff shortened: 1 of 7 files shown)


src/components/autoSuggest/index.js
```diff
@@ -335,6 +335,26 @@ class AutoSuggest extends React.Component {
...
+                valueRight: '',
+                isSelected: false,
+                type: 'no-results',
+              }]}
+              header={{
+                titleLeft: '',
+                titleRight: '',
+              }}
+              onMouseDown={() => {}}
+              setSelectedRow={() => {}}
```
michaeltomasik: Do We need these empty functions here?

faival: removed unnecessary usage of ResultList component

PR #1224, opened 2018-08-28
Implement new design of selected delegates - Closes #1147

### LINKED ISSUE #1147: Implement new design of selected delegates
### Expected behaviour
- Implement latest design of 'Theatre seats' layout from https://taikonauten.atlassian.net/browse/L20-475

- Layout that keeps vital information visible on top (numbers). 
- Hovering on seats shrinks rectangles & shows names above.
- In the case of border seats, tooltip will overlay whatever content is behind, no matter how long the delegate name is.
- Clicking on the seat will have no effect of removing or adding a vote. 
- The list should concatenate votes in this order {voted, vote, free-slot} and the matrix always restricted to 101 slots.
- an additional row with no label or visual separator will show unvoted delegates
- pending votes should animate transitioning between {free-slot,voted} colors, and it's tooltip should say '<delegate_name> (pending)'
- squares should keep same size as checkboxes, matrix margin should provide breakpoints to make profit of width of container. 

### Steps to reproduce

faival:
### What was the problem?
- #1147
### How did I fix it?
- Removing the circular progress bar usage
- Providing a VoteMatrix component
- Styling it as in the designs. 
### How to test it?
- test on testnet also with large delegate usernames
### Review checklist
- The PR solves #1147


faival:
PR changse requested:
* padding changing between L, and XL breakpoints
* initial jump when loading view, till votes for delegates are resolved
* M breakpoint needs to be shown.

faival:
updated specs:

* don't show delegates names, only (free, voted, downvoted, unvoted)
* don't order by ranking,
* pending should happen in all states (free, downvoted,unvoted, voted)

PR #1231, opened 2018-08-30
Add bookmarks to send LSK process - Closes #1100

### LINKED ISSUE #1100: Implement Bookmark accounts
### Expected behaviour
Implement Bookmark accounts
https://taikonauten.atlassian.net/browse/L20-474

### Specification
When sending LSK
- if the user has no followed accounts, then the "Send to address" is just a regular input
screen shot 2018-08-16 at 12 57 39
- If the user is following accounts, then focusing "Send to address" opens suggestions with followed accounts
screen shot 2018-08-16 at 12 59 09
- suggestions can be filtered by title or address
screen shot 2018-08-16 at 13 00 17
- after selecting a followed account, the input field displays the title and address 
screen shot 2018-08-16 at 13 01 00
- if the followed account doesn't have a title, the input field displays the address and text saying "Followed account" 
screen shot 2018-08-16 at 13 02 25
- the previous two points work the same in confirm step
screen shot 2018-08-16 at 13 02 41
 - if recipient is not followed yet, then the last step shows a button to follow it, leading to the next screens
screen shot 2018-08-16 at 13 03 54

### Changes to the original specification
- "Title" of bookmark/followed should be mandatory, instead of optional.

### Subtasks
- Create address autocomplete component *medium
- Update the send steps to use the autocomplete and show followed account label (if exists) *medium
- Allow following recipient from the final step of send process *medium

michaeltomasik:
### What was the problem?
#1100

### Review checklist
- The PR solves #1100


src/components/sendTo/followAccount.js
```diff
@@ -44,7 +45,19 @@ class FollowAccount extends React.Component {
...
-            prevStep();
+            // istanbul ignore else
+            if (typeof this.props.finalCallback === 'function') {
+              this.props.finalCallback();
+              this.props.reset();
+            } else {
+              this.props.nextStep({
+                success: true,
+                title: this.props.t('Success'),
+                body: title + this.props.t(' has been added to your Dashboard'),
```
slaweet: This should be
`body: this.props.t('{{title}} has been added to your Dashboard', { title })`,
because in another language the word order might need to be different.

src/components/bookmark/bookmark.test.js
```diff
@@ -0,0 +1,141 @@
...
+      childContextTypes: {
+        store: PropTypes.object.isRequired,
+        i18n: PropTypes.object.isRequired,
+      },
+    });
+  });
+
+  it('should render a component', () => {
+    expect(wrapper).to.have.exactly(1).descendants('Bookmark');
+  });
```
slaweet: I think this test is not testing anything at all and should be removed.

src/components/sendWritable/send.js
```diff
@@ -93,14 +95,23 @@ class SendWritable extends React.Component {
...
+              label={this.props.t('Send to address')}
+              address={this.state.recipient}
+              handleChange={this.handleChange.bind(this, 'recipient', true)}
+            />
+          }
           <ReferenceInput
-            className='reference'
+            context={this}
             label={this.props.t('Reference (optional)')}
             address={this.state.reference}
```
slaweet: Please fix the param name `address={this.state.reference}` -> `reference={this.state.reference}`. I get it's not related to this PR, but it's a small refactor not worth its own PR.

slaweet:
> 🐛 There is a borderline between sender input and suggestions list, see mockups

- Fixed

> 🐛 Filter by address doesn't work

- Fixed

> 🐛 if the followed account doesn't have a title it shows the address, not the 'Followed account' text

- The point of making title mandatory was not to have handle this case. 

> 🐛 After choosing suggestion it is not possible to activate field clicking on the title

- Fixed

> 🐛 If entered input doesnt match any suggestions I still can see the suggestions wrapper
image

- That's expected behaviour. It shouldn't show suggestions wrapper only if you don't have any followed accounts at all.

> 🐛 Button name should be 'Add to followed accounts', not 'Add Followed Account'
image

- Fixed

> 🐛 Description text of an action on Follow account is not complete as well

- Fixed, but used ", and" instead of "&"

> 🐛 It is possible to add followed account without a title on this step

- Fixed

> 🐛 On a confirmation step dot is missing

- Fixed

> 🐛 If the address is invalid
image

- Fixed

> 🐛 If there is no followed accounts, there is no 'Add to followed accounts' proposal on the last step

- Cannot reproduce

> Can we place the 'Send to address' sign a bit higher? It feels too tight
image

- Fixed

slaweet:
- Fixed to show 'Add to followed accounts' button when no followed accounts
- Fixed `tab` not to select account like `enter`
- Fixed `TypeError: Cannot read property 'props' of undefined `
- Styling: would be nice, but I will make a follow-up ticket to also check with Designer.
- Padding fixed

Efefefef:
Nice feature!

Issue #1326, opened 2018-10-08
Registration: Custom node reg always falls back on localhost:4000


Efefefef:
### Steps to reproduce

1. Run lisk core not on the 4000 port
1. Choose Custom Node
1. Enter Url on which you are running core
1. Walk through the registration process

### Actual result

Error popup: Unable to connect to node
Hub tries to connect to localhost:4000

### Expected result

Appear logged in devnet on specified port

### Version

1.3, 1.4

slaweet:
### Specification
- Custom node URL should be validated (try to connect to it) on login page when clicking "Create Lisk ID"
- If the request is successful, then go to the first step of creating a new account
- If the request fails, then show "Unable to connect the node" error toast and stay on the login page
- If network is not custom node, then clicking "Create Lisk ID" doesn't check the mainnet/testnet URL and goes directly to the first step of creating a new account

PR #1327, opened 2018-10-08
E2E for registration - Closes #1311

### LINKED ISSUE #1311: Create e2e tests for registration

Efefefef:
### What issue have I solved?

-- #1311

Scenarios take time to run, but as #1326 proves its really important to run the reg process from start to the end for each scenario, as it is critical functionality

### Review checklist

### DIFF
--- a/test/cypress/e2e/registration.spec.js
+++ b/test/cypress/e2e/registration.spec.js
@@ -1,12 +1,22 @@
 import networks from '../../constants/networks';
+import chooseNetwork from './utils/chooseNetwork';
 
 const ss = {
-  networkDropdown: '.network',
   networkStatus: '.network-status',
   newAccountBtn: '.new-account-button',
   nodeAddress: '.peer',
   errorPopup: '.toast',
+  getPassphraseButton: '.get-passphrase-button',
+  iUnderstandCheckbox: '.i-understand-checkbox',
+  revealCheckbox: '.reveal-checkbox',
+  passphraseTextarea: 'textarea.passphrase',
+  itsSafeBtn: '.yes-its-safe-button',
+  passphraseWordHolder: '.passphrase-holder label',
+  getToDashboardBtn: '.get-to-your-dashboard-button',
+  backButton: '.multistep-back',
 };
+
+const registerUrl = '/register';
 
 /**
  * Generates a sequence of random pairs of x,y coordinates on the screen that simulates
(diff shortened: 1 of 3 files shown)


test/cypress/e2e/registration.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  cy.get('.reveal-checkbox')
+    .trigger('mousedown')
+    .trigger('mousemove', { which: 1, pageX: 100, pageY: 0 })
+    .trigger('mouseup');
+  cy.get('textarea.passphrase').invoke('text').as('passphrase');
+  cy.get('.yes-its-safe-button').click();
+  cy.get('.passphrase-holder label').each(($el) => {
+    if (this.passphrase.includes($el[0].textContent)) cy.wrap($el).click();
+  });
+  cy.get('.get-to-your-dashboard-button').click();
```
slaweet: Is there a reason to use the selectors in `registerUI`, like `.get-to-your-dashboard-button` and not to define them in `ss` object like other selectors.

test/cypress/e2e/registration.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  for (let i = 0; i < 70; i += 1) {
+    cy.get('main').first().trigger('mousemove', {
+      which: 1,
+      pageX: 500 + (Math.floor((((i % 2) * 2) - 1) * (249 + (Math.random() * 250)))),
+      pageY: 500 + (Math.floor((((i % 2) * 2) - 1) * (249 + (Math.random() * 250)))),
+    });
+  }
+};
+
+const chooseNetwork = (network) => {
```
slaweet: `chooseNetwork` looks to be duplicated in `test/cypress/e2e/login.spec.js`.
 To avoid duplication, `chooseNetwork` could be defined in one "util" file, exported, and imported in both usages.

test/cypress/e2e/registration.spec.js
```diff
@@ -0,0 +1,118 @@
...
+describe('Registration', () => {
+  it('Create Lisk ID for Mainnet by default ("Switch Network" is not set)', function () {
+    cy.visit('/register');
+    registerUI.call(this);
+    cy.get(ss.networkStatus).should('not.exist');
+  });
+
+  it('Create Lisk ID for Mainnet ("Switch Network" is false)', function () {
+    cy.addLocalStorage('settings', 'showNetwork', false);
+    cy.visit('/register');
```
slaweet: `/register` path is used twice, so it could be defined in one place like helpUrl in https://github.com/LiskHQ/lisk-hub/blob/dae2d24299c3b2deaa55ee171e98edc9e4e68c66/test/cypress/e2e/help.spec.js#L12

test/cypress/e2e/registration.spec.js
Efefefef: yes, exactly, it is duplicated. ok, will move it to util for now

test/cypress/e2e/registration.spec.js
Efefefef: The reason is I don't see what to do better yet. When I need to reuse selectors.
1) To move all selectors inside a special file. So we could import and use it in dev and qa bidirectionally.
2) To import some selectors that we need from one spec file from the other spec file.

Depending on the decision this batch of selectors are awaiting their fate cause they are bulky and only will be used inside this function.
Any thoughts?

test/cypress/e2e/registration.spec.js
Efefefef: makes sence

PR #1354, opened 2018-10-15
Open direct link with autologin - Closes #1339

### LINKED ISSUE #1339: Account: When opened by direct link shows empty activity
### Steps to reproduce

1. Open some account with activity history by direct link e.g. `http://localhost:8080/#/explorer/accounts/16313739661670634666L` 

### Actual result

No activity yet

### Expected result

Activity is loaded and shown

### Version

1.4, 1.5

michaeltomasik:
### What issue have I solved?

-- #1339

Stopped overwriting activePeer when We are in autologin mode

### How has this been tested?

set up autologin and try to access e`explorer/transactions`

### Review checklist

### DIFF
--- a/src/store/middlewares/peers.js
+++ b/src/store/middlewares/peers.js
@@ -2,6 +2,7 @@ import { activePeerSet, activePeerUpdate } from '../../actions/peers';
 import actionTypes from '../../constants/actions';
 import networks from './../../constants/networks';
 import getNetwork from './../../utils/getNetwork';
+import { shouldAutoLogIn } from './../../utils/login';
 
 const peersMiddleware = store => next => (action) => {
   next(action);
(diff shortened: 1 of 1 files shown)


src/store/middlewares/peers.js
```diff
@@ -10,7 +11,10 @@ const peersMiddleware = store => next => (action) => {
 
   switch (action.type) {
     case actionTypes.storeCreated:
-      store.dispatch(activePeerSet({ network }));
+      // It changes network back to mainnet when we want to autologin for explorer account
+      if (!shouldAutoLogIn) {
```
slaweet: `shouldAutoLogIn` us a function that should be used like this:
https://github.com/LiskHQ/lisk-hub/blob/9a74df7a8f9a3a333d307efc2b3b2115f6cbe1e8/src/store/middlewares/account.js#L124

Currently, the condition is never met.

src/store/middlewares/peers.js
```diff
@@ -10,7 +11,10 @@ const peersMiddleware = store => next => (action) => {
 
   switch (action.type) {
     case actionTypes.storeCreated:
-      store.dispatch(activePeerSet({ network }));
+      // It changes network back to mainnet when we want to autologin for explorer account
```
slaweet: The comment explains well why you added the if statement and makes sense in the context of this PR. But if I would read it after the PR is merged, I would expect that the comment explains why the thing inside the condition is executed, and then it doesn't make much sense.

PR #1355, opened 2018-10-15
Update test doc - Closes #1329

### LINKED ISSUE #1329: Update e2e test README.md

Currently, we using Cypress for tests but the command in `README.md` are not updated

### Suggested solution

Update `README.md`

Efefefef:
### What issue have I solved?

-- #1329

### Review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,31 +100,29 @@ npm run test-live
 ## Run end-to-end tests
 In order to run e2e tests you need to install [lisk-core](https://github.com/LiskHQ/lisk)
 
-#### Setup
+#### Setup core
 
-Setup protractor
+Setup a lisk test node as described in [https://github.com/LiskHQ/lisk#tests](https://github.com/LiskHQ/lisk#tests)
 
-```
-./node_modules/protractor/bin/webdriver-manager update
-```
-
-Setup a lisk test node to run on localhost:4000 as described in https://github.com/LiskHQ/lisk#tests
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -100,17 +100,11 @@ npm run test-live
...
 
-Setup protractor
-
-```
-./node_modules/protractor/bin/webdriver-manager update
-```
-
-Setup a lisk test node to run on localhost:4000 as described in https://github.com/LiskHQ/lisk#tests
-
-And run it with [pm2](http://pm2.keymetrics.io/).
```
slaweet: This line is still needed as `e2e-test-setup.sh` uses `pm2`

README.md
```diff
@@ -100,17 +100,11 @@ npm run test-live
...
+1) Setup a lisk test node as described in [https://github.com/LiskHQ/lisk#tests](https://github.com/LiskHQ/lisk#tests)
+2) Apply blockchain snapshot `./e2e-test-setup.sh ~/git/lisk/` (replace `~/git/lisk/` with your path to lisk core)
```
slaweet: I think `./e2e-test-setup.sh ~/git/lisk/`  should be in the "Run" section, because it's needed every time before running the e2e suite of tests

README.md
Efefefef: it automatically starts with that script

slaweet: not if you don't have pm2 installed. So maybe the wording should be "Install [pm2](http://pm2.keymetrics.io/)"

PR #1362, opened 2018-10-16
Display full voters - Closes #1344 

### LINKED ISSUE #1344: Only first 10 voters are displayed
This is a follow-up to https://github.com/LiskHQ/lisk-hub/issues/1325

It turned out that fixing voters is more complex because with the current API we cannot keep the functionality

### Specification

- "Who voted for this delegate" should show the correct number 
- Only first 100 items are fetched at first
- Keep currenty ipmemented behaviour: only first 35 are displayed at first, rest after clicking "Show more"
- Further items are fetched with infinite scroll
- "Filter votes" functionality should be removed because the new API doesn't allow to do this.

bmaggi-lisk:
### What issue have I solved?

-- #1344

1. Use searchAccount to fetch account info 
2. Map state.search in prop when need to locate delegate statistics

### How has this been tested?

1. Login as 
2. Search for delegate
3. Open Delegate statistics
4. Scroll down on `Who voted for this delegate` then if delegate has more than 100 voters continue to scroll until all voters are disaplyed

### Review checklist


src/actions/search.js
```diff
@@ -32,19 +32,40 @@ const searchVotes = ({ address }) =>
...
 
-const searchVoters = ({ address, publicKey }) =>
+const searchVoters = ({
+  address, publicKey, offset, limit, append,
+}) =>
   (dispatch, getState) => {
     const activePeer = getState().peers.data;
-    getVoters(activePeer, publicKey).then(response =>
+    getVoters(activePeer, {
+      publicKey, offset: (offset || 0), limit: (limit || 100),
```
slaweet: Why the default values here again? You have them already in the definition of `getVoters` in `src/utils/api/delegate.js`
```suggestion
      publicKey, offset, limit,
```

src/components/transactions/votedDelegates.js
```diff
@@ -17,19 +17,25 @@ class VotedDelegates extends AccountList {
...
+              {` (${this.props.votersSize || 0})`}
             </div>
-            {super.renderSearchFilter('votersFilterQuery', t('Filter voters'))}
           </div>
           <div className={styles.value}>
             {voters && voters
-              .slice(0, this.state.showVotersNumber)}
+                .slice(0, this.state.showVotersNumber)}
           </div>
           {voters.length > this.state.showVotersNumber ?
```
slaweet: The condition doesn't work right. "Show more" should be shown only before the first click on it. Now it keeps showing it.

src/components/transactions/votedDelegates.js
```diff
@@ -17,19 +17,25 @@ class VotedDelegates extends AccountList {
...
             <div onClick={() => { super.showMore('showVotersNumber'); }} className={`${styles.showMore} showMore  show-voters`}>
               <FontIcon className={styles.arrowDown} value='arrow-down'/>
               {this.props.t('Show more')}
-            </div> : ''
+            </div> : <Waypoint
```
slaweet: IMO it's easier to read if there is a component in the ternary when it starts on its own line.
```suggestion
            </div> :
            <Waypoint
```

PR #1379, opened 2018-10-23
Set autoDisplayName for Xvfb plugin - Closes #1378

### LINKED ISSUE #1378: Jenkins Xvfb plugin does fails for concurrent builds on the same machine
This is a follow-up of #1366 

### Steps to reproduce

1. Trigger Jenkins jobs so that two builds run on the same machine at the same time.

### Actual result

Xvfb plugin does picks the same display number for both build; the second one fails.

### Expected result

Different display numbers should be chosen.

### Version

`>=1.5.0`

ghost:
### What issue have I solved?

#1378 

Set [autoDisplayName](https://github.com/jenkinsci/xvfb-plugin/blob/master/src/main/java/org/jenkinsci/plugins/xvfb/Xvfb.java#L414) option for Xvfb plugin.

### How has this been tested?

Build on Jenkins.

### Review checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -94,7 +94,7 @@ pipeline {
 					"cypress": {
 						withCredentials([string(credentialsId: 'lisk-hub-testnet-passphrase', variable: 'TESTNET_PASSPHRASE')]) {
 							ansiColor('xterm') {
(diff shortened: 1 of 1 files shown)


Efefefef (approved):
Thank tou

PR #1389, opened 2018-10-24
Report Cypress results to GitHub - Closes #1382

### LINKED ISSUE #1382: Find a better place for Cypress link
### Current behaviour
Cypress post a link to the test results only in the Jenkins console output.
E.g.  https://dashboard.cypress.io/#/projects/1it63b/runs/454 in https://jenkins.lisk.io/job/lisk-hub/job/development/335/consoleFull

### Expected behaviour
There should be an easy way to get to the cypress link

- One option would be to post cypress results to GitHub PR checks with the link in "details":
screen shot 2018-10-23 at 11 03 07

- Another option would be to add it to Jenkins build report:
screen shot 2018-10-23 at 11 07 25

Either would be fine, depends on how much work each is.

ghost:
### What issue have I solved?

#1382 

Added GitHub notifications as described in [#1382](https://github.com/LiskHQ/lisk-hub/issues/1382#issuecomment-432197545)

### How has this been tested?

This PR will be the test: the new GitHub notifications should show up under "Checks"

### Review checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -114,7 +114,11 @@ pipeline {
 									cd -
 
 									npm run serve -- $WORKSPACE/app/build -p 300$N -a 127.0.0.1 &>server.log &
-									npm run cypress:run -- --record
+									set +e
+									npm run cypress:run -- --record |tee cypress.log
+									ret=$?
+									grep --extended-regexp --only-matching 'https://dashboard.cypress.io/#/projects/1it63b/runs/[0-9]+' cypress.log |tail --lines=1 >.cypress
+									exit $ret
 									'''
 								}
 							}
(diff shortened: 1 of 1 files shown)


slaweet:
The main part why we want this is still missing. "Details" link of the check should link to the cypress build (something like https://dashboard.cypress.io/#/projects/1it63b/runs/454)

fchavant:
@slaweet @Efefefef I don't see any way to get that link from within the `Jenkinsfile`

slaweet:
@fchavant I googled a bit and didn't find a nice way to get the link.

So we can do it a hacky way. We can make output of `npm run cypress:run -- --record` go a log file in addition to `stdout` (with [tee](http://www.gnu.org/software/coreutils/manual/html_node/tee-invocation.html)) and then `grep` the log file for `https://dashboard.cypress.io/#/projects/`

Jenkinsfile
```diff
@@ -111,7 +114,11 @@ pipeline {
 									cd -
 
 									npm run serve -- $WORKSPACE/app/build -p 300$N -a 127.0.0.1 &>server.log &
-									npm run cypress:run -- --record
+									set +e
+									npm run cypress:run -- --record |tee cypress.log
+									ret=$?
```
slaweet: `$?` is the return code from `tee ...` not from `npm run cypress:run ...`, so `exit 0` even if `2 of 13 failed`  in https://jenkins.lisk.io/job/lisk-hub/job/PR-1389/3/console

Not sure if there is a way to get the return code. If not, we can grep `cypress.log` for `All specs passed!`

fchavant: @slaweet you're right, let's set `pipefail`

slaweet (approved):
Thank you, @fchavant.

Efefefef:
@fchavant thank you!

PR #1397, opened 2018-10-26
Transactions to 'yourself' are all seen as "account initialization" - Closes #1281

### LINKED ISSUE #1281: Transactions to 'yourself' are all seen as "account initialization"
If you send a transaction to yourself using the hub, they are all seen as "account initialization", this look weird/wrong. If an initialization has already performed, the message shouldn't be repeated.

### Steps to reproduce

Use the hub to send lisk to your **own** address.

### Actual result

Address is mentioned as "account initialization"

### Expected result

Address should mention the senders address. 

### Solution

Save locally if an "account initialization" is performed. After the first time the hub should mentioned the senders address, even though its the same as the receiver.

image

bmaggi-lisk:
### What issue have I solved?

-- #1281

1. Type 0 transactions to yourself (same sender & receiver) show address instead of "Account initialization"
2. When initializing account the reference message for the transaction is set to "Account initialization"

### How has this been tested?

1. Transfer some amount of LSK to yourself
2. Go to wallet outgoing transactions check for the transaction you made to yourself you should see  recipient (self) address instead of "Account initialization"

### Review checklist

### DIFF
--- a/src/components/sendReadable/send.js
+++ b/src/components/sendReadable/send.js
@@ -79,13 +79,14 @@ class SendReadable extends React.Component {
   send(event) {
     event.preventDefault();
     this.setState({ loading: true });
+    const isAccountInit = this.props.accountInit && (this.props.account.address === this.state.recipient.value);
     this.props.sent({
       account: this.props.account,
       recipientId: this.state.recipient.value,
       amount: this.state.amount.value,
       passphrase: this.props.passphrase.value,
       secondPassphrase: this.props.secondPassphrase.value,
-      data: this.props.reference,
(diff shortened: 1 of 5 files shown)


src/components/sendReadable/send.js
```diff
@@ -79,13 +79,15 @@ class SendReadable extends React.Component {
   send(event) {
     event.preventDefault();
     this.setState({ loading: true });
+    const isAccountInit = this.props.accountInit &&
+      (this.props.account.address === this.state.recipient.value);
```
slaweet: The second line of the condition is not needed, `this.props.accountInit` should be already all you need.

PR #1457, opened 2018-11-07
Ensure that all component names match their file/folder name - Closes #1395

### LINKED ISSUE #1395: Ensure that all component names match their file/folder name
Some problems are "Search" vs. "SearchBar" or "TransactionsList" vs. "TransactionList"

osvaldovega:
### What issue have I solved?

-- #1395 

Check all components under src/components folder and review all the components names and verify that the component name match their folder name in case it doesn't then fix it by changing the component name or folder if need and then fix all the import paths where that component is called.

### How has this been tested?

I run the app and confirm that everything works as should be and run the test-live and cypress to ensure that nothing breaks.

### Review checklist


src/components/search/index.js
```diff
@@ -4,12 +4,12 @@ import Box from '../box';
 import EmptyState from '../emptyState';
 import styles from './search.css';
 
-const SearchResult = ({ t }) => (
+const Search = ({ t }) => (
```
slaweet: `SearchResult` is IMO more descriptive for what this component does. So rather change the folder name

PR #1490, opened 2018-11-14
Limit number of Jenkins builds retained - Closes #1488

### LINKED ISSUE #1488: Set discard old builds property in Jenkinsfile
Old builds are staying around forever on Jenkins, causing us to run out of disk space
### Steps to reproduce

1. build the development branch about 30 times

### Actual result
old builds live forever

### Expected result
needs manual intervention

### Version
1.6.0, 1.7.0

Nazgolze:
Limit number of Jenkins builds retained
### What issue have I solved?

-- #1488 

Add limit to number of builds retained in Jenkinsfile

### How has this been tested?

Tested in replay run in development branch

### Review checklist


Jenkinsfile
```diff
@@ -5,6 +5,7 @@ pipeline {
 	agent { node { label 'lisk-hub' } }
 	options {
 		skipDefaultCheckout true
+		buildDiscarder(logRotator(numToKeepStr: '15'))
```
fchavant: That's 15 build per branch. Do we have a ballpark figure of how much disk space that corresponds to?

slaweet: With running `development` every hour, 15 is not even a day worth of history. I would like at least a week worth of history. 
If needed, we can run `development` only every 2 hours.

Nazgolze: each job is about 100M, so 1.5G per branch/PR right now. There are 7 open PRs right now, so the max at the moment should be about 14G, including the 1.7.0 and development branches

We could go with `numToKeepStr: 84`, we would have to throw more disk space at this then. It's not just the development branch, but the 1.7.0 branch. if 1.7.0 and development both run once very 2 hours, at the end of the week it's 17G for those 2 total. If the other ones max out as well, it comes out to 75G 

came up with 84 by doing number of runs (12) times days in a week (7). If it's not running on weekends though, 60 is a nicer number, should bring it down to 54G (best guess --- more open PRs will increase the theoretical max)

slaweet: @Nazgolze I'm thinking about if we can reduce build size from 100MB

60MB is just the `AppImage` artifact: https://jenkins.lisk.io/job/lisk-hub/job/development/567/artifact/dist/ another 15-20MB will be other artifacts. 

I don't need all of the old artifacts. Maybe there is a way to keep past 84 builds but their artifacts only for past 5? 


Also, PR builds should be smaller since they don't create the `AppImage` as branch builds

Nazgolze: @slaweet yes, we can do exactly this.

I just pushed a change, setting the number of builds retained to 168 and artifacts to 5

PR #1522, opened 2018-11-27
Implement detaching login logout - Closes #1512

### LINKED ISSUE #1512: Implement detaching hardware wallet - Child of #1101

Detach when smb connect hwWallet to show ledger login icon
and logout when disconnected

michaeltomasik:
### What issue have I solved?

-- #1512

### Review checklist

### DIFF
--- a/app/src/ledger.js
+++ b/app/src/ledger.js
@@ -1,19 +1,18 @@
 import { app, ipcMain } from 'electron'; // eslint-disable-line import/no-extraneous-dependencies
 import Lisk from 'lisk-elements'; // eslint-disable-line import/no-extraneous-dependencies
 import { LedgerAccount, SupportedCoin, DposLedger } from 'dpos-ledger-api'; // eslint-disable-line import/no-extraneous-dependencies
+import TransportNodeHid from '@ledgerhq/hw-transport-node-hid'; // eslint-disable-line import/no-extraneous-dependencies
+
 import win from './modules/win';
 
-
-// eslint-disable-next-line import/no-extraneous-dependencies
-// TransportNodeHid = require('@ledgerhq/hw-transport-node-hid');
 // mock transportnodehid, todo fix this to work also on windows and linux
-const TransportNodeHid = {
-  open: () => {},
-  listen: () => {},
-  setListenDevicesDebounce: () => {},
-  setListenDevicesPollingSkip: () => {},
-};
-TransportNodeHid.setListenDevicesDebounce(200);
+// const TransportNodeHid = {
+//   open: () => {},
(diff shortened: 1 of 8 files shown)


slaweet:
Still getting the same errors.

michaeltomasik:
https://github.com/nwjs/nw.js/issues/6769#issuecomment-415770115

slaweet:
I was able to solve the previous error on Windows by 

- adding this commit https://github.com/LiskHQ/lisk-hub/pull/1522/commits/728df4b1e2417cdc04de8bc82170392b14e8e245 and 

- In PowerShell as administrator run:
```
npm install -g --production windows-build-tools
```

But then I get another error:
screen shot 2018-12-10 at 08 35 55

Efefefef:
Works with Ubuntu
to run first execute
`wget -q -O - https://raw.githubusercontent.com/LedgerHQ/udev-rules/master/add_udev_rules.sh | sudo bash`

Efefefef:
🐛  Auto-logout initiated by ledger doesn't log you out

🐛 'Confirm transaction on Ledger Nano S' should be the label of the send _confirmation_ screen only

@slaweet 2ph reg is disabled, but delegate reg link is removed. is it fine? Shouldn't we disable it as well?
image

🐛 As seen above you disabled 'Security' header as well 

🐛 I can't edit ledger acc label. When I click into acc name field it works as a click onto account

@slaweet Why we have both namings: 'Account name' for ledger and 'Title' for bookmarks?

🐛 Whats up with height here?
image

🐛 App doesn't show previously created account if the activity remained zero. Don't we have a way to show it anyhow? 
  
🐛 Latest activity is empty when signing in ledger account

🐛 Custom node address input become empty after logging out

slaweet:
> @slaweet 2ph reg is disabled, but delegate reg link is removed. is it fine? Shouldn't we disable it as well?
It is inconsistent, but fine for now. Because it should also be fixed for non-ledger: delegate reg link is hidden after done, bug 2nd pass changes to "already registered". Or 2nd pass is disabled if not logged in.

> @slaweet Why we have both namings: 'Account name' for ledger and 'Title' for bookmarks?

Good point, let's use 'Title' instead of 'Account name'

> 🐛 beta?
Yes

Efefefef:
🐛 If unable to connect to node - still proceeds to account selection screen
🐛 Win: if plugged in while app is already working - doesn't show ledger login interface
🐛 Win: After was disconnected while app was closed - Then it will always show the ledger login interface. Even after reinstallation.

Efefefef:
I retested regressions and functionality on mac build. Check crossed-outs above

PR #1525, opened 2018-11-27
Create standalone send page - Closes #1495 

### LINKED ISSUE #1495: Create standalone Send page
Subtask of #1466

The goal of this ticket is to create standalone Send page that can be accessed by path `/wallet/send`, is accessible only after login and in another subtask of #1466 will be linked from Wallet page

osvaldovega:
### What issue have I solved?

-- #1495

Working on it.
Standalone page created, right now navigation is possible and route file updated to properly use /send in browser.

### Review checklist


osvaldovega:
requested changes are fixed, branch updated

src/components/multiStep/index.js
```diff
@@ -82,6 +82,7 @@ class MultiStep extends React.Component {
       prevStep: step.prevStep,
       reset: this.reset.bind(this),
       ...step.data[step.current],
+      finalCallback,
```
slaweet: If you add `finalCallback` here, the condition `if (step.current === (children.length - 1)) {` below is not needed

src/components/resultBox/resultBox.js
```diff
@@ -45,29 +45,17 @@ class ResultBox extends React.Component {
...
-              this.props.history.replace(this.props.history.location.pathname);
-            }}>
-            {this.props.t('Okay')}
-          </Button>
           {this.props.success &&
-            this.props.reciepientId && this.isNotYetFollowed(this.props.reciepientId) ?
+            this.props.recipientId && this.isNotYetFollowed(this.props.recipientId) ?
             <Button className={`add-follwed-account-button ${styles.addFollowedAccountButton}`}
               onClick={() => {
                 this.props.nextStep({ address: this.props.reciepientId });
```
slaweet: The type needs to be fixed on this line too. Currently, the followed account is saved without the address.

```suggestion
                 this.props.nextStep({ address: this.props.recipientId });
```

src/components/multiStep/index.js
osvaldovega: you are right it is fixed now

src/components/resultBox/resultBox.js
osvaldovega: fixed

PR #1534, opened 2018-12-03
Update process file to support process.argv - Closes #1531

### LINKED ISSUE #1531: Update process file to support process.argv
In the **app/modules/win.js** in line 47, the process.argv is not working properly as the process was override by the process.js file.

```
if (!process.isPlatform('darwin')) {
    win.send({ event: 'openUrl', value: process.argv[1] || '/' });
}
```

The process file should include a function that returns the argv values as well to this work properly.

### Steps to reproduce
This error happens only on a windows or linux machine.

1. Download the lisk hub beta version on a windows/linux machine
2. Install the application 
3.  During the installation process an error will display, about JavaScript error, cannot read property 1 of undefined.

### Actual result

During the installation process in a windows/linux machine the following error shows up.
screen shot 2018-12-03 at 17 04 20

### Expected result

That the process.argv can be read properly and then the installation process could finish without any problem

osvaldovega:
### What issue have I solved?

-- #1531 

In the **process.j**s file under **app/src/modules** add a new function that returns the process.argv value and then in the **win.js** file this can be call as a function.

With this now the **win.test.js** file the mock that function and run properly.

### Review checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -39,7 +39,6 @@ module.exports = {
     '/node_modules/',
     'app/src/ipc.js',
     'app/src/ledger.js',
-    'app/src/modules/process.js',
     'src/components/receive/index.js',
     'src/actions/liskService.js',
     'src/actions/peers.js', // FollowUp #1515
--- a/jest.config.js
+++ b/jest.config.js
@@ -39,6 +39,7 @@ module.exports = {
     '/node_modules/',
     'app/src/ipc.js',
     'app/src/ledger.js',
+    'app/src/modules/process.js',
     'src/components/receive/index.js',
     'src/actions/liskService.js',
     'src/actions/peers.js', // FollowUp #1515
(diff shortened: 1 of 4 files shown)


jest.config.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
     '/node_modules/',
     'app/src/ipc.js',
     'app/src/ledger.js',
+    'app/src/modules/process.js',
```
slaweet: I miss some comment explaining why the file coverage is ignored. 
```suggestion
    'app/src/modules/process.js', // this file is ignored because it's just a simple wrapper created to allow mocking `process` in unit tests. 
```
Maybe it would be even more clear if the the explanation and `istanbul ignore` would be inside the `process.js` file

osvaldovega: done

PR #1558, opened 2018-12-11
Setup routes tracking with Matomo (Piwik) - Closes #1452 

### LINKED ISSUE #1452: Setup route tracking with Matomo, disabled by default
Subtask of https://github.com/LiskHQ/lisk-hub/issues/1102

Show a switcher to enable tracking in settings only if `/settings/?showrTackingSwitch=true`.

osvaldovega:
Add implementation and basic configuration of Matomo (Piwik) into the project.
At this point the tracker is only configure to track URL changes and assigned to a specific user.

### What issue have I solved?
-- #1452

Following the Matomo configuration on this URL https://developer.matomo.org/guides/ab-tests/browser and based on the library https://www.npmjs.com/package/react-piwik

### How has this been tested?

I made some tests to see if Matomo is able to track the URL changes and check how many visits and so far works good. More configuration will be need in case to want something more specific.

### Review checklist


src/utils/piwik.js
```diff
@@ -0,0 +1,52 @@
...
+    url: piwikOptions.URL,
+    siteId: piwikOptions.SITE_ID,
+    trackErrors: true,
+  });
+
+  setPiwikParameters();
+  return piwik;
+};
+
+const checkIfPiwikIsEnable = () => {
```
slaweet: Typo:
```suggestion
const checkIfPiwikIsEnabled = () => {
```

src/utils/piwik.js
```diff
@@ -0,0 +1,52 @@
...
+    piwikSetAlready = true;
+  }
+};
+
+const disabledPiwikTracking = () => {
+  ReactPiwik.push([piwikOptions.FORGET_CONSENT_GIVEN]);
+};
+
+const connectPiwikWithHistory = (history) => {
+  checkIfPiwikIsEnable();
```
slaweet: Typo:
```suggestion
  checkIfPiwikIsEnabled();
```

src/utils/piwik.js
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  setPiwikParameters();
+  return piwik;
+};
+
+const checkIfPiwikIsEnable = () => {
+  const isPiwikSet = localJSONStorage.get('statistics', false);
+
+  if (isPiwikSet && !piwik && !piwikSetAlready) {
```
slaweet: These two variable names are IMO too similar: `isPiwikSet` and `piwikSetAlready`.

 IMO a more accurate name for the first one is `isPiwikEnabled`. 
And do we need the second one (`piwikSetAlready`) at all? Isn't checking `&& !piwik` enough?

PR #1613, opened 2018-12-20
Fix Auto-logout timer number format - Closes #1608

### LINKED ISSUE #1608: Auto-logout timer number format gone wrong
### Steps to reproduce

1. Signin
1. Check the auto-logout timer

### Actual result

image

renatomassao:
### What issue have I solved?

-- #1608 

Changed how we handled the verification to include or not the leading 0 on numbers. Now we always add the leading 0, and only get the last two digits.

### How has this been tested?

Checked that all tests pass, and that the UI is correct now.

### Review checklist


renatomassao:
image

PR #1615, opened 2018-12-20
Create a function that check if the port use for electron is being in use - Closes #1611

### LINKED ISSUE #1611: Create a function that check if the port use for electron is being in use
A port validation needs be to done before run the app with electron as express is running inside, we need to ensure that the port use is available and in case it is not then use a different one.

osvaldovega:
This PR is to address the static port set on express application that runs in electron.
So to avoid that electron can show the absolute file path, what we are trying to do is run the app in express and then pass the URL to electron to be able to open the project from express.

So the request is to create a function/module that provide a specific port for run the express server however if this port is being used then the app can reassign the port to any that is available.
At this point there is one file called **portAssign.js** under **app/src/modules**, is using promises to address this request, however at the time to start the app by electron this fails.

This PR will be open and we need to rethink in a different way to address this problem.

### What issue have I solved?
-- #1611 

### Review checklist

### DIFF
--- a/app/server.js
+++ b/app/server.js
@@ -1,3 +1,5 @@
+// const findAvailablePort = require('./src/modules/portAssign');
+
 const server = {
   // eslint-disable-next-line max-statements
   init: () => {
@@ -6,7 +8,7 @@ const server = {
     const bodyParser = require('body-parser');
 
     const app = express();
-    const port = process.env.PORT || 8080;
+    const port = process.env.PORT || 8081;
 
     app.use(bodyParser.json());
     app.use(bodyParser.urlencoded({ extended: true }));
--- a/jest.config.js
+++ b/jest.config.js
@@ -39,6 +39,7 @@ module.exports = {
     '/node_modules/',
     'app/src/ipc.js',
     'app/src/ledger.js',
+    'app/src/modules/portAssign.js',
     'src/actions/liskService.js',
     'src/actions/transactions.js',
     'src/components/account/stories.js',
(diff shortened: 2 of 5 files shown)


app/server.js
```diff
@@ -1,12 +1,14 @@
+// const findAvailablePort = require('./src/modules/portAssign');
+
```
slaweet: The comment can be removed?

app/src/main.js
```diff
@@ -47,8 +55,14 @@ app.on('activate', () => {
...
-    win.create({
-      electron, path, electronLocalshortcut, storage, checkForUpdates,
+    fp(3000, (err, port) => {
+      if (err) {
+        throw new Error(err);
+      }
+
+      win.create({
+        electron, path, electronLocalshortcut, storage, checkForUpdates, port,
+      });
```
slaweet: This whole `fp(3000, (err, port) => {` is duplicated twice in this file. Maybe some helper function? Or at least define the default port in one place (a constant).

jest.config.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
     '/node_modules/',
     'app/src/ipc.js',
     'app/src/ledger.js',
+    'app/src/modules/portAssign.js',
```
slaweet: Is this needed?

osvaldovega: removed it

app/src/main.js
osvaldovega: done

PR #1631, opened 2019-01-08
Implement V2 Confirm your Passphrase of registration - Closes #1583

### LINKED ISSUE #1583: Implement Step 3/4: Confirm your Passphrase of registration
Subtask of #1574

The new design:
https://netguru.invisionapp.com/share/EWOJMVMSNBQ#/screens/336109817

renatomassao:
### What issue have I solved?

#1583 

Created the component for the confirm passphrase step, and the component for displaying the options.
The component always start with the words at the indexes `[2, 9]` missing, and it gets random ones if the user pick the wrong ones.
Clicking in a selected word, before hitting confirm will show the options for that index again.
The user still have 3 tries, as it is on our current version, and if the user pick the wrong ones 3 times in a row, it shows a message and the user has to go back before attempting again.
The right most options are only enabled if the user picked the left most first, and the confirm button only get enabled after picking both.
Clicking the confirm button set a timer before shuffling the words, or going to the next step.

### How has this been tested?

Going through the new registration process at `/registration`, and when getting to the confirm passphrase step, just need to pick the words and check if it's validating correctly.
**Known issue:** Right now, picking the correct words will get the user to a loader, because it's pointing to the next step on the process, which isn't implemented yet.

### Review checklist


src/components/registerV2/confirmPassphrase.js
```diff
@@ -0,0 +1,180 @@
...
+
+class ConfirmPassphrase extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      numberOfOptions: 3,
+      numberOfWords: 2,
+      words: [2, 9],
+      tries: 0,
+      maxTries: 3,
```
slaweet: `numberOfOptions`, `numberOfWords`, `maxTries` doesn't have to be on the state, since it doesn't change.

src/components/registerV2/confirmPassphrase.js
```diff
@@ -0,0 +1,180 @@
...
+      answers: [],
+      options: [],
+      hasErrors: false,
+      isCorrect: false,
+      outOfTries: false,
+    };
+
+    this.handleSelect = this.handleSelect.bind(this);
+    this.handleConfirm = this.handleConfirm.bind(this);
+    this.verifyChoices = this.verifyChoices.bind(this);
```
slaweet: Most of the current codebase uses .bind(this) before the method is used in JSX. Is there any advantage of this way?

src/components/registerV2/backupPassphrase.js
```diff
@@ -120,7 +128,8 @@ class BackupPassphrase extends React.Component {
             </SecondaryButtonV2>
           </span>
           <span className={`${registerStyles.button} ${grid['col-xs-4']}`}>
-            <PrimaryButtonV2>
+            <PrimaryButtonV2 disabled={!this.state.hasSaved}
```
slaweet: This button should be always enabled. The user can just write the passphrase down on a piece of paper instead of copying it.

src/components/registerV2/confirmPassphrase.js
renatomassao: This is the way that is recommended right know at the React documentation, and also doing the binding in the render method, may cause some perform issues, almost always unnoticeable.
Because when binding the function it generates a new reference for the function, so if we bind it in the constructor, it keeps the same reference, but binding on render, when React compares the virtual DOM for changes, it will consider the function as being a new one all the time.

PR #1651, opened 2019-01-10
Store bookmarks public key - Closes #1643

### LINKED ISSUE #1643: Store public key in bookmarks
As it is planned to change the address system, we should start storing account's public key, in addition, to address. To be able to support and switch to the new system in the future.

Efefefef:
### What issue have I solved?

#1643 

Add/Update publicKey on followerUpdate if its different but not undefined

### Review checklist

### DIFF
--- a/src/actions/followedAccounts.test.js
+++ b/src/actions/followedAccounts.test.js
@@ -54,10 +54,9 @@ describe('actions: followedAccount', () => {
     expect(followedAccountRemoved(data)).to.be.deep.equal(expectedAction);
   });
 
-  it('should update a followed account if balance or privateKey changed', () => {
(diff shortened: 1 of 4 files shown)


src/actions/followedAccounts.test.js
```diff
@@ -54,17 +54,23 @@ describe('actions: followedAccount', () => {
     expect(followedAccountRemoved(data)).to.be.deep.equal(expectedAction);
   });
 
-  it('should update a followed account if balance changed', () => {
+  it('should update a followed account if balance or privateKey changed', () => {
```
slaweet: This `it` is really bigger than it should. IMO each Case (1-5) should be a separate `it`.

I see that previously the first two cases were in the same `it`, even those should have been split.

PR #1661, opened 2019-01-16
Implement the new bottom bar - Closes #1646

### LINKED ISSUE #1646: Implement the new bottom bar
Subtask of #1645 
The goal is to implement this bar:
screen shot 2019-01-10 at 11 01 08

https://netguru.invisionapp.com/share/X9PMLKSUNYE#/screens/339318625

osvaldovega:
### What issue have I solved?

-- #1646

This new feature is the status bar/bottom bar, that contains 3 mains elements, the timer that is for the auto logout, the second element is about the network at what is connected the user and the last element are the links for help and feedback.

The status bar is placed in the bottom of the browser window.

There is one main component called StatusBar and this has 3 child components, the countdown element, the network element and the externalLinks element. All these 3 child elements are reused from existed version.

NOTE: the positioning of the status bar, should be fix/change once the side bar menu will be remove.

### How has this been tested?
1-) Open the app and check the status bar at the bottom of the page.
2-) log in
3-) check the first element is the timer in the left of the status bar
4-) after 5 minutes the counter should change the color as the mockups and the reset button should be display on the right.
5-) doing a click on the reset button the timer should start again at 10 min remains.
6-) the second element should be green if has connection with the selected network after login
7-) if you logout and change another network and login again this will change. saying to what network are you connected and the green dot at the left means that has connection. if not should be red dot.
8-) for the last element just do a click in the links and this should redirect to the corresponding page.

### Review checklist


slaweet:
- the bottom bar is too wide on smaller screen, the links on the right are out of view
screen shot 2019-01-18 at 08 39 22
- I thought that things that are on the bottom bar should be removed for their original position (timeout, network indicator, help and feedback), but it too much hustle with tests, can be done only in the next PR
- the network indicator is good for mainnet, but for custom node it should show also the URL of the node. (Note that custom node can also be mainnet node, e.g. https://liskwallet.punkrock.me)

osvaldovega:
changes done. Once the side bar menu be remove, I will properly realign the status bar to the center of the screen

dmrkv:
Everything looks like in a mockup, except font-family, which is Open Sans right now, but must be Gilroy. And it is not only in status-bar, but everywhere.

But we already discussed it with @osvaldovega, he said, that it will be fixed then.

osvaldovega:
@Efefefef I made the changes that you requested, I add the font that @dmrkv is using in the mockups.
About the network, @slaweet told me that in this way is how should be working, even if you check the screenshot that Vit put in the PR earlier you can see that the url was not present now it.

Efefefef:
Nice

PR #1662, opened 2019-01-17
Implement Initialize account banner on dashboard - Closes #1660

### LINKED ISSUE #1660: Add account initialization banner to Dashboard
The account initialization should be also advertized on dashboard.

### Specification
- The module should be displayed if account doesn't have public key stored in the blockchain and the balance is at least the fee 0.1 LSK.
- The copy should be as in this design, except the title should say "Initialize Lisk ID"
screen shot 2019-01-16 at 15 53 49
https://netguru.invisionapp.com/share/X9PMLKSUNYE#/screens/340950916

- Clicking the button should lead to the account initialization already implemented in `/wallet/send`

renatomassao:
### What issue have I solved?

#1660 

Created new Banner component that have `title. children and footer` props, title being an `String`, and children and footer being nodes.
Added the same conditions that we have on the wallet, for showing or not the banner for initializing the account.
The background and button are different from the design, to be more consistent with the layout.

### How has this been tested?

The banner should not appear if the user is not logged in any account.
Create a new account, and transfer some LSK to it, or login into an account that has LSK and didn't activate the account yet, and check if the banner shows up on the `/dashboard`, after activating the account, the banner shouldn't appear again.

### Review checklist


slaweet:
As discussed, some changes to the specification:
- The "Create first transaction" button should skip the current first step of "Account initialization", as it was duplicate to the banner. 
- Second step of initializaiton should contain to link from the frist step
- going to `/wallet/send` page (e.g. Send button on wallet page) should never show Account initialization.

This might require some e2e test to be updated.

test/cypress/e2e/send.spec.js
```diff
@@ -162,29 +162,44 @@ describe('Send', () => {
   });
 
   /**
-   * Initialization dialogue and functionality is shown for the fresh account
+   * Initialization dialogue and functionality shouldn't be shown through wallet send.
```
Efefefef: The annotation here does not correlate the test title. It should be about the positive case showing the banner, not negative about not showing the dialogue in send

test/cypress/e2e/send.spec.js
```diff
@@ -162,29 +162,44 @@ describe('Send', () => {
...
     cy.get(ss.resultMessage).should('have.text', msg.transferTxSuccess);
     cy.get(ss.accountInitializationMsg).should('not.exist');
     cy.wait(txConfirmationTimeout);
     cy.reload();
-    cy.visit(urls.send);
-    cy.get(ss.accountInitializationMsg).should('not.exist');
     cy.visit(urls.wallet);
     cy.get(ss.transactionRow).eq(0).as('tx');
     cy.get('@tx').find(ss.transactionAddress).should('have.text', accounts['without initialization'].address);
     cy.get('@tx').find(ss.transactionReference).should('have.text', 'Account initialization');
```
Efefefef: Its worthy to check that the initialization banner is not shown on the dashboard anymore

test/cypress/e2e/send.spec.js
```diff
@@ -162,29 +162,44 @@ describe('Send', () => {
...
+  it('Should be able to send even with account not initialized', () => {
+    cy.autologin(accounts['without initialization'].passphrase, networks.devnet.node);
+    cy.reload();
+    cy.visit(urls.wallet);
+    cy.get(ss.transactionSendButton).click();
+    cy.get(ss.accountInitializationMsg).should('not.exist');
+    cy.get(ss.recipientInput).should('exist');
+  });
+
   /**
```
Efefefef: Here you use the 'without init' account, which has already been initialized in the previous test. The blockchain snapshot is reapplied only before the start of the tests, so you shouldn't use this acc anymore or implement test somehow dependent on other tests. As this one is a negative check and you will need another not initialized account I would suggest to include the check into the previous existing test

PR #1678, opened 2019-01-18
Implement ledger login on LoginV2 - Closes #1670

### LINKED ISSUE #1670: Implement Ledger login on LoginV2
Subtask of #1574 

We should be able to login with Ledger on the V2 login page, as on the current login page.

renatomassao:
### What issue have I solved?

#1670

Copied code from previous Login screen to have the same interaction with the hardware wallet. And updated styles according to the layout

### How has this been tested?

Needs to build and run in electron to test, to be able to access this page, update the routes.js file, and make the `loginV2` or `splashcreen` have:
```
{
  path: '/',
  exact: true,
}
```
So when the app is open with electron you can access it.
Go to the LoginV2 page, and plug the hardware wallet, and input the pin in the hardware, the link to login with the hardware should appear bellow the passphrase inputs and have the same functionalities as in the old design.

### Review checklist


osvaldovega:
@renatomassao everything looks good, but I think that something that you can do is to avoid the elements move when the nano ledger link show/hide in the page

renatomassao:
@osvaldovega Updated to have some animation when the hardware wallet is detected, could you check it again?
ledger

osvaldovega:
looks much better, great thanks @renatomassao.

Efefefef:
@renatomassao Why we are not keeping the ledger icon and other icons in new design?
image

renatomassao:
@Efefefef Have to check with @dmrkv because it's like this on the design
image

PR #1684, opened 2019-01-22
Make liskCoreUrl work as default network on start - Closes #1623

### LINKED ISSUE #1623: Restart with set custom node failed due to network error
### Steps to reproduce

1. Set `https://liskwallet.punkrock.me` as custom node and log into any account
2. Close Lisk Hub and restart it 

You'll get this error:
 grafik

### Expected result
Lisk Hub should normally connect to the custom node after a restart, so you get updated results on dashboard for your bookmarks or any other stuff you can do **without logging into any account**, like checking accounts and so on.

### Version
Lisk Hub 1.8.0 on Windows 10

Efefefef:
### What issue have I solved?

#1623 

Include liskCoreUrl in the condition in peers middleware

### How has this been tested?

Try to login and reboot to mainnet, testnet, custom node other than localhost:4000

### Review checklist

### DIFF
--- a/src/store/middlewares/peers.js
+++ b/src/store/middlewares/peers.js
@@ -1,7 +1,6 @@
 import { liskAPIClientSet, liskAPIClientUpdate } from '../../actions/peers';
 import actionTypes from '../../constants/actions';
 import networks from './../../constants/networks';
-import getNetwork from './../../utils/getNetwork';
 import { shouldAutoLogIn, getAutoLogInData, findMatchingLoginNetwork } from './../../utils/login';
 
 const peersMiddleware = store => next => (action) => {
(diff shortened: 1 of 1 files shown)


src/store/middlewares/peers.js
```diff
@@ -16,20 +15,20 @@ const peersMiddleware = store => next => (action) => {
   // else default network
   if (loginNetwork) {
     loginNetwork = loginNetwork.slice(-1).shift();
-  } else if (!loginNetwork) {
-    loginNetwork = liskCoreUrl ? networks.customNode : networks.default;
+  } else if (!loginNetwork && liskCoreUrl) {
+    loginNetwork = { ...networks.customNode, address: liskCoreUrl };
+  } else if (!loginNetwork && !liskCoreUrl) {
+    loginNetwork = networks.default;
```
renatomassao: Wouldn't be better to use a ternary here? Instead of having another `else if`. Like:
```
loginNetwork = liskCoreUrl ? { ...networks.customNode, address: liskCoreUrl } : networks.default
```

PR #1690, opened 2019-01-25
Autosuggest delegate fetching bug - Closes #1351

### LINKED ISSUE #1351: Search: Delegate search occur not waiting results to be fetched
### Steps to reproduce

1. Enter existing delegate name into the search field
1. Immediately press enter

### Actual result

No results

### Expected result

Delegate account page
App should wait for results being fetched, then show results

### Version

1.3, 1.4, 1.5

image

michaeltomasik:
### What issue have I solved?

#1351 

Added a simple regex to test if the input is a transaction ID or and address, if not and there is no results, it delays the action of the enter until the component receives the results.
If the value changes after the enter being queued, it removes the delay, and user has to press enter again.

### How has this been tested?

Quickly type the starting of the delegate name or paste something and hits enter, it should wait a bit, and go to the page after the component receives the results.
If it's an address or transaction, it goes directly to the page, or to a result not found page.

### Review checklist


Efefefef:
For delegates the solution is good. But for txs and addresses feels not very good
Can we do it conditionally? If the value doesn’t match tx or address - enter is disabled until results

test/cypress/e2e/search.spec.js
```diff
@@ -213,15 +211,4 @@ describe('Search', () => {
     cy.get(ss.searchInput).click().type('43th3j4bt324');
     cy.get(ss.searchNoResultMessage).eq(0).should('have.text', 'No results found');
   });
-
-
-  /**
-   * Search for nonexistent item
-   * @expect no results plug
-   */
-  it('Search for nonexistent item - shows no results plug', () => {
```
renatomassao: Had to remove this test, since the string `43th3j4bt324` will be considered as a possible delegate name, and the enter key press will be delayed, so the user won't be sent to the `No results` page.

Efefefef: You can put a made-up tx id instead

Efefefef:
Recent searches autosuggestion is broken. I can see, that happened in the last sprint already, but know we implemented new logic on top. Let's check 1.9 version as a reference and discuss

Efefefef:
👍 Finally, we have fully operational search

Efefefef:
Please check the code review comment above

PR #1701, opened 2019-01-29
Adjustment of transactions table on Wallet - Closes #1636

### LINKED ISSUE #1636: Implement wallet new transactions table style
The transaction table in the wallet should be implemented on a new page in the new style:

screen shot 2019-01-09 at 16 52 44

Adding the "Balance after transaction",  "Filter Transactions" and "Export to CSV" will be a separate future task

https://netguru.invisionapp.com/share/HGOHX8L6U3M#/screens/336327393

renatomassao:
### What issue have I solved?

#1636 

Duplicated the necessary files of the transactions table and updated styles and functionalities as needed on the `/walletV2` route. And created a walletTransactionsV2.
The Other tab, delegate info and votes, will be done in a separate ticket.

### How has this been tested?

On `/wallet` nothing should have changed.
On /walletV2 the transaction table should have the new layout and clicking on a transaction takes the user to `explorer/transactions/{id}` routes instead of changing the content on the wallet route.
Known issue: When refreshing the page on /walletV2 causes an error, will be fixed when it's set to the /wallet route.

### Review checklist


src/components/spinnerV2/spinnerV2.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { translate } from 'react-i18next';
+import styles from './spinnerV2.css';
+
+const SpinnerV2 = ({ t }) => (
+  <span className={styles.wrapper}>
+    <span className={`${styles.spinner} spinner`}/>
+    <span className={`${styles.label}`}>{t('Pending...')}</span>
```
slaweet: I think `t('Pending...')` should be passed from outside as `label` prop, since it can be other text in other places.

PR #1708, opened 2019-01-30
Update top bar css breakpoints design from (S, M, L) to (M, L, XL) - Closes #1707

### LINKED ISSUE #1707: Update top bar css breakpoints design from (S, M, L) to (M, L, XL)
The design should match the breakpoints (M, L, XL) and not (S, M, L), for what the current design for top bar needs to be update.

osvaldovega:
### What issue have I solved?

-- #1707 

The menuItems.css and statusBar.css files were update to properly math the design with the correct view breakpoint.

as the small view doesn't exist in the project breakpoint I had to move the design from small to medium and medium to large and large to Xlarge right now the behavior is the correct one.

### How has this been tested?

Run the application and resize the main window from biggest possible width to the medium (max-width: 1024px) and then you will see that the menu items in the top bar change the display and positioning.

There is a fix for the status bar to, that include the fix for the calculation of the width in the different views, right now the alignment is the same for all the views. 

### Review checklist

### DIFF
--- a/src/components/autoSuggestV2/autoSuggest.css
+++ b/src/components/autoSuggestV2/autoSuggest.css
@@ -133,7 +133,7 @@
   .autoSuggest {
     right: 0;
   }
-  
+
   .placeholder {
     left: 22px;
   }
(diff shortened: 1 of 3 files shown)


osvaldovega:
thanks @Efefefef I just fixed, lets check again

PR #1765, opened 2019-02-12
Try out snapshots - Closes #1556

### LINKED ISSUE #1556: Try Jest snapshot on help component to see how it works
UPDATED: Try out Jest snapshot testing workflow on request component

Efefefef:
### What issue have I solved?

#1556

expect(wrapper).toMatchSnapshot();

### Review checklist


slaweet:
Wow, 6000 lines file with two snapshots. It looks quite scary. @renatomassao @michaeltomasik @osvaldovega what do you think?

Maybe `help` is not the best example component. @Efefefef can you use it also on some smaller component to see how big the snapshot will be, e.g. `src/components/statusBar/externalLinks.test.js`?

Efefefef:
@slaweet Well, it fails with 
`FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`
After all, I don't think snapshots is a very good idea anymore

Efefefef:
@slaweet Any other example component?

slaweet:
@Efefefef Another component to try could be `src/components/statusBar/network.test.js`

Efefefef:
@slaweet done
I think it sucks 👎

slaweet:
I see. 1946 lines for the snapshot of a simple component (https://github.com/LiskHQ/lisk-hub/blob/94f9df56e436053b636a9a980eccd69fe5a9dbe1/src/components/statusBar/__snapshots__/network.test.js.snap). 

So I sugest not to use snapshots, not to merge this PR and close https://github.com/LiskHQ/lisk-hub/issues/1556 as `wontfix`. 

Any objections @renatomassao @osvaldovega ? If not, please put "👍" reaction on this comment.

osvaldovega:
@Efefefef @slaweet @renatomassao guys I see the file and looks like is very big, don't know if the output format is correct, I was quickly checking on internet and I found this maybe @Efefefef and check it.

https://github.com/facebook/jest/blob/master/docs/SnapshotTesting.md

```
Jest uses pretty-format to make snapshots human-readable during code review. On subsequent test runs Jest will simply compare the rendered output with the previous snapshot. 
```
According to this, the snapshot should be only render the component in a readable form, so I dont know if that is the case.

Here is the link for the pretty format

https://github.com/facebook/jest/tree/master/packages/pretty-format

renatomassao:
Ok, I got the problem with the help of @osvaldovega comment, the problem here is that we are checking the mounted version of enzyme, it generates an `ReactWrapper Object` to test against, thats why it generates this much of lines.
If we use just the `react-test-renderer`, like on line:
https://github.com/LiskHQ/lisk-hub/blob/94f9df56e436053b636a9a980eccd69fe5a9dbe1/src/components/statusBar/network.test.js#L23
`const setup = props => renderer.create(<Network {...props} />).toJSON();`

The snapshot will look better, just the rendered markup, I'm pushing a commit to this branch with the updated `network` snapshot and `.test.js` file.

renatomassao:
Actually, found a better solution for this, still using the `ReactWrapper` from enzyme but getting the actual html of that before trying to match with the snapshot: https://github.com/LiskHQ/lisk-hub/pull/1765/commits/912a4371e424dd4ce505463636d0b3477193432b https://github.com/LiskHQ/lisk-hub/pull/1765/commits/74e1b94325634d47edcff9c2051aa73c067ade3d

Uses the prettier to format the HTML string, because the `.html()` function of enzyme returns a one line string.

slaweet:
From technical meeting: We won't use snapshots as they make the tests less readable.

PR #1773, opened 2019-02-13
Implement wallet details module - Closes #1668

### LINKED ISSUE #1668: Implement My wallet details module
This task is part of #1624

renatomassao:
### What issue have I solved?

#1668 

Create new component `walletDetails` that shows the account balance, last transaction and change since last Visit. And add it on `/wallet` page.
Created `loadLastTransaction` action, that gets the last transaction of the address, since we can't rely on the `transactions[0]` of the store, because it's updated when filtered.
Created actions, reducers and subscriber to `wallets`, to store the balance of the last visit on the localStorage, to display the difference from last visit.

PS: The module is stretching to use all the available space, since we currently don't have the graph module, once we implement the graph module, we'll update this module to look how it was designed.

### How has this been tested?

Go to `/wallet` and the component should be rendered with the account balance, the last amount of the last transaction and change since last visit.
Should be working on the new another account page, even though it's not accessible yet, go to `/explorer/accountsV2/[some_address]`

### Review checklist


Efefefef:
@slaweet 1) Can you please specify the appearance of Last transaction icon and text in case of every tx type. 
2) What counts as last visit? From buisness view? From technical view?

michaeltomasik:
@Efefefef
@sansus answer
1) no changes, it will always be the same
2) last visit from local app. If user is displaying a given profile then the change value should be 0
3) yes, if fee was detracted from the account balance then it's a change within balance

Efefefef:
2. So if it is stored in localStorage, 
1) we should save the network as well
2) show `-` if there was no data stored (proposed by @sansus ). Because `0` means 0 changes which would be misleading financial info. 
3. All these numbers should always include a fee.
4. While loading -  value should not be zero. Can we please avoid showing `x LSK` span until it does load the value. This would help automated tests

renatomassao:
> 2. All these numbers should always include a fee.
So when it's just a vote or 2nd passphrase registration, it would show just the Fee, in the last transaction Field?

sansus:
@renatomassao Yes, it would show just the fee in this case.

Last transaction = Transaction amount + Fee

osvaldovega:
check and fix the conflicts with the svgIcons file

Efefefef:
According to @sansus in 'Since last visit' we should show the change in balance happened after the last visit of /wallet page. That means if while staying in-app, we opened some other page for example Settings and then went back to Wallet page, then we should see the sum of transactions happened

slaweet:
Talked to @Efefefef to clarify how 'Since last visit' should work. Summary:
- it should show the difference since the last visit to Hub, not from the last visit of wallet page. That is, switching pages inside Hub doesn't change the value and current implementation is correct.

PR #1786, opened 2019-02-18
Implement a terms of use page - Closes #1739

### LINKED ISSUE #1739: Implement a pop-up with Terms of use
Implement a pop-up based on designs from https://github.com/LiskHQ/lisk-hub/issues/1732

osvaldovega:
### What issue have I solved?

-- #1739 

For this request 2 things were implemented.
1-) Add the links terms of use in the Help page where the users can check the updated terms and conditions.

2-) A screen that will load at the beginning where the user MUST accepter and read the terms of use without this the user can continue. Once accepted the user dont need to accept it anymore, unless cache is clear.

### How has this been tested?

Clear the browser cache and open the app, the new screen will display and the user need to accepter it, if the user write in the search bar / or /login this will redirect to the terms-of-use page if this wasn't accepted first, once the user do click on accept this will redirect to the splash screen, the next time the app start will redirect automatically to splash screen.

### Review checklist


src/store/reducers/settings.js
```diff
@@ -19,6 +19,7 @@ const initialState = JSON.parse(localStorage.getItem('settings')) || {
   isHarwareWalletConnected: false,
   isRequestHowItWorksDisable: false,
   statistics: false,
+  isTermsOfUse: true,
```
slaweet: Naming the flag `isTermsOfUse` IMO doesn't clearly say if `true` means terms were accepted or `false` means terms were accepted. I suggest:

```suggestion
  areTermsOfUseNotAccepted: true,
```

osvaldovega: change done.

osvaldovega:
finally e2e test pass

Efefefef:
@slaweet With all respect to Osvaldo's description, conditions are not clear and should not come from the developer himself. The parent-parent task contains a list of 'legal requirements' which does not seem related. A question is also what are the final designs. Plus I can no longer access https://netguru.invisionapp.com

PR #1810, opened 2019-03-06
Implement Follow Account Dropdown - Closes #1805

### LINKED ISSUE #1805: Implement "Follow account" dropdown on another account page
Based on designs from https://github.com/LiskHQ/lisk-hub/issues/1758

With these exceptions:
- "Notifications" Checkbox is not present at all (this will change when we implement the "Notifications" feature)
- "On your dashboard"  checkbox is checked and cannot be unchecked (this will change once we have a dedicated "Followed accounts" page)

renatomassao:
### What issue have I solved?

#1805 

Created FollowAccount component that has the form for adding a new followed account, validates the length of the inputed name.
Show on dashboard is as readonly for now.
If account already followed, should show a different button for opening the dropdown, and show the saved account name, and a danger button.

### How has this been tested?

Search any account on the search bar, on the account page, change `explorer/accounts/` at the URL to `explorer/accountsV2/`, follow/following  button should be on header.

### Review checklist


src/components/followAccount/followAccount.js
```diff
@@ -0,0 +1,200 @@
...
+        },
+      },
+      isValid: false,
+    });
+  }
+
+  // eslint-disable-next-line complexity
+  render() {
+    const { t, isFollowing } = this.props;
+    const { isValid, fields } = this.state;
```
osvaldovega: try to not use destructuring in the render

src/components/followAccount/index.js
```diff
@@ -0,0 +1,15 @@
...
+import FollowAccount from './followAccount';
+
+const mapStateToProps = state => ({
+  accounts: state.followedAccounts.accounts,
+});
+
+const mapDispatchToProps = dispatch => ({
+  addAccount: data => dispatch(followedAccountAdded(data)),
+  removeAccount: data => dispatch(followedAccountRemoved(data)),
+});
```
osvaldovega: right now we are trying to use the mapDispatchToProps just as an object instead use the dispatch, could you changed or you need it in this way.

```
const mapDispatchToProps = {
  addAccount,
  removeAccount,
};
```

src/components/toolbox/buttons/css/baseV2.css
```diff
@@ -27,6 +26,20 @@
...
+  background-color: var(--color-danger) !important;
+  color: var(--color-white) !important;
+
+  &:disabled {
+    opacity: 0.8;
+  }
+
+  &:focus,
+  &:hover {
+    background: #ff6300 !important;
```
osvaldovega: I think you can use a variable in here

src/components/followAccount/index.js
renatomassao: done

src/components/toolbox/buttons/css/baseV2.css
renatomassao: Right now it's because this isn't the final color, once we get the correct hover color this should be updated

PR #1812, opened 2019-03-11
Implement read only delegate tab - Closes #1778

### LINKED ISSUE #1778: Implement read-only delegates detail page
Implement based on designs from https://github.com/liskhq/lisk-hub/issues/1756

renatomassao:
### What issue have I solved?

#1778 

Implemented request to API to get the register delegate transaction and last block forged.
Created DelegateTab component, that displays all informations.
As discussed with @sansus we are showing the `Last Forged Block` as DateTime, with the same format used in the wallet (DD MMM YY, HH:mm).
If there is no last Forged block, it should appear as `-`

### How has this been tested?

Search any delegate account, and change the URL to `/explorer/accountsV2/:address`, the delegate tab should be loaded besides the wallet tab.

### Review checklist


src/actions/account.js
```diff
@@ -248,3 +255,19 @@ export const accountDataUpdated = ({
       dispatch(liskAPIClientUpdate({ online: false, code: res.error.code }));
     });
   };
+
+export const updateAccountDelegateStats = account =>
+  (dispatch, getState) => {
+    const liskAPIClient = getState().peers.liskAPIClient;
+    const { address, publicKey } = account;
+    getTransactions({
```
osvaldovega: man there is no problem here with nothing everything works fine, is just that after check this with @slaweet every new code that we create based on promises we need to do it with Async/Await, so you dont need to change much but yes please change it to async/await

src/actions/search.js
```diff
@@ -1,21 +1,33 @@
...
     getDelegate(liskAPIClient, { publicKey }).then((response) => {
-      dispatch({
-        data: {
-          delegate: response.data[0],
-          address,
-        },
-        type: actionTypes.searchDelegate,
+      getTransactions({
+        liskAPIClient, address, limit: 1, type: transactionTypes.registerDelegate,
+      }).then((transactions) => {
```
osvaldovega: same here, I know this is old code, but know you make changes is good to have it with async/await

src/components/transactionsV2/walletTransactionsV2/walletTransactionsV2.js
```diff
@@ -168,6 +175,10 @@ class WalletTransactionsV2 extends React.Component {
 
     const { t, account } = this.props;
 
+    // const delegate = account.isDelegate
```
osvaldovega: I think you forget to remove this from the code

src/components/transactionsV2/walletTransactionsV2/walletTransactionsV2.js
```diff
@@ -198,8 +209,14 @@ class WalletTransactionsV2 extends React.Component {
         }
 
         <TabsContainer>
-          <WalletTab tabName={this.props.t('Wallet')}
+          <WalletTab tabName={t('Wallet')}
             {...overviewProps}/>
+          {/* account.isDelegate && delegate.txDelegateRegister
```
osvaldovega: Here too.... or I dont know if it is because you need it for later

src/components/transactionsV2/walletTransactionsV2/walletTransactionsV2.test.js
```diff
@@ -53,79 +49,103 @@ describe('WalletTransactions V2 Component', () => {
...
     it('should render onboarding banner and hide when close is clicked', () => {
-      expect(wrapper).to.have.descendants('.onboarding');
+      expect(wrapper).toContainMatchingElement('.onboarding');
       wrapper.find('.onboarding .closeBtn').simulate('click');
-      expect(wrapper).to.not.have.descendants('.onboarding');
+      expect(wrapper).not.toContainMatchingElement('.onboarding');
+    });
+  });
+
+  describe.skip('Delegate Account', () => {
```
osvaldovega: is it useless this unit test or you have it for later?

src/utils/api/transactions.test.js
```diff
@@ -50,6 +50,27 @@ describe('Utils: Transactions API', () => {
       });
     });
 
+    it('should call transactions.get with type', () => {
+      const params = {
+        liskAPIClient,
+        address: '123L',
```
osvaldovega: I think could not be necessary as it is just one line, but maybe you can use accounts constants and get the address from any account that you need

src/components/transactionsV2/walletTransactionsV2/walletTransactionsV2.js
renatomassao: This one is because the delegate tab will be used on the wallet page, but not right now.

src/components/transactionsV2/walletTransactionsV2/walletTransactionsV2.js
renatomassao: This one is because the delegate tab will be used on the wallet page, but not right now.

src/components/transactionsV2/walletTransactionsV2/walletTransactionsV2.test.js
renatomassao: This one is because the delegate tab will be used on the wallet page, but not right now.

PR #1817, opened 2019-03-12
Replace old send page with new new one - Closes #1743

### LINKED ISSUE #1743: Replace old send page with new new one
This is subtask of https://github.com/LiskHQ/lisk-hub/issues/1691 Should be done after all other subtasks.

osvaldovega:
### What issue have I solved?

-- #1743

In this PR the routes file has been update to redirect to the folder that contains the new version (design) of the send page.

Then update the E2E tests that are related to the send page, all the tests for this page were update

### How has this been tested?

Open the app the go to wallet and do a click on the send button, this will display the new version of send and then user can go through the process.
You can send and LSK by typing address or selecting one if there is a bookmark already or using the HW ledger.

### Review checklist


slaweet:
I quickly tried it out and found these two things to fix:
- Top bar "My wallet" icon shold be also blue
- "Edit transaction"  doesn't keep the values. Clicking it will go to the previous step with all inputs empty.
Screen Shot 2019-03-13 at 10 20 44

src/components/sendV2/summary/summary.js
```diff
@@ -24,22 +26,76 @@ class Summary extends React.Component {
...
           ...this.state.secondPassphrase,
           hasSecondPassphrase: true,
         },
+      };
+    }
+
+    // istanbul ignore else
+    if (account.hwInfo && account.hwInfo.deviceId) {
+      newState = {
+        ...this.state,
```
renatomassao: Maybe you don't need this one here? Since the setState will just receive the 2 following keys, it should alter the previous state. Same on line `43`

src/components/sendV2/summary/summary.js
```diff
@@ -71,46 +127,51 @@ class Summary extends React.Component {
...
+      },
     });
-    this.props.nextStep({ fields: { ...this.props.fields } });
   }
 
   render() {
-    const { secondPassphrase } = this.state;
-    const isBtnDisabled = secondPassphrase.hasSecondPassphrase
+    const { secondPassphrase, isHardwareWalletConnected } = this.state;
+    let isBtnDisabled = secondPassphrase.hasSecondPassphrase
```
renatomassao: Can't you just do:
`secondPassphrase.hasSecondPassphrase && secondPassphrase.isValid !== '' && !secondPassphrase.isValid`
Looks better than a ternary that just puts `false`, IMO

src/components/sendV2/summary/summary.test.js
```diff
@@ -98,4 +105,29 @@ describe('Summary', () => {
     wrapper.update();
     expect(props.nextStep).toBeCalled();
   });
+
+  it('should goind to next page if everyting is successfull by Hardware Wallet', () => {
```
renatomassao: Fix typo

test/cypress/utils/enterSecondPassphrase.js
```diff
@@ -1,12 +1,22 @@
-const ss = {
+import ss from '../../constants/selectors';
+
+const ss2 = {
```
renatomassao: don't like the name, but also, not sure what we could but instead of this 😕

src/components/sendV2/summary/summary.js
osvaldovega: done

src/components/sendV2/summary/summary.js
osvaldovega: good, fixed

src/components/sendV2/summary/summary.test.js
osvaldovega: done

test/cypress/utils/enterSecondPassphrase.js
osvaldovega: correct, I just fixed

renatomassao:
🥇

slaweet:
@Efefefef  
> 1) Is that a final layout of tx summary page? For me it looks like the layout is broken/

Yes, the layout is on purpose. I see your concern, we'll be doing some user testing with everything, so we'll see if more people feel it's weird.

> 2) Do we need this fake checkbox?

I thought it's giving some info about this being on the dashboard, but you're right - we don't need it for now. @osvaldovega please comment it out with a comment: it should be enabled when the new UX of the dashboard and the followed accounts page is implemented.

osvaldovega:
Hi @Efefefef, @slaweet 

I just update the PR as in case there was some issue, here is the status update after that.

* Texts are not corrected => All texts are taking from design mockups and checked from someone called Jacob and marketing department. In the other hand we have the document where are all the texts that we are using in the app.

* Amount gets rechecked every time messange changed => this issue is fixed.

* It is said that "The validation happens after user moves to another input or leaves the current one.", now it happens on every change => No this time, in the meeting where we checked the design I asked about this and Daria mentioned that now this works in the same way that twitch do it, and this is how it works now. You can check it with @slaweet 

* Higher amount validation is incorrect. It still shows the error if amount number was decreased => Correct this issue is fixed.

* (optional) should be different font, not braces
*  You should count Bytes in message not characters => The problem here was that I forget to update the text but it is counting in bytes, so I fixed the text now.

* 'Read more' text in tooltip is not aligned => correct, I just fixed.

*  Tx submitted page - we got rid of follows/ers notation . => what part exactly, I took all the text from the mockups so or the mock up it is wrong or maybe I do.

* Bookmarks input doesn't show up when I click into the recipient input, suggestions only apear when I start typing => correct this change in the new design, as soon you type if there are bookmarks you will see it


If there is any other doubt just let me know as I check all these details with Daria and Vit, so we can check it with @slaweet tomorrow if you want.

Efefefef:
🐛 Click send on Confirmation page: Uncaught (in promise) Error: Status 404 : Delegate not found
🐛 First option is always highlighted as chosen 

image

slaweet:
Based on the feedback from @Efefefef there should be these changes to specification to make things more consistent:
1. Change "Amount of transactions" to "Amount"
2. Change transaction fee tooltip text  "because of that we ask for a small fee" -> "because of that there is a small fee"
3. Show address suggestions on focus before the user starts typing. Like it was in the old send.
4. Change "Follow recipient" to "Bookmark recipient" as so far we have "Bookmarks" on the dashboard. This will have to be changed only when we implement the new UX of the dashboard, where you really can "Follow" accounts and get notifications about their new transactions etc.
5. Change success message to "You'll find it in your wallet and it will be confirmed in a matter of minutes" to already use the one from the final UI https://netguru.invisionapp.com/share/H6R0WTTNYUD#/screens/352120454

And some discussed things that should not be changed:

1. The epic says that error validation happens only on unfocusing an input. The way it's been implemented
2. In "Message (optional)" label was in designs with two fonts and without parentheses, but let's keep it will be as implemented in the final UI https://netguru.invisionapp.com/share/H6R0WTTNYUD#/screens/352119907

osvaldovega:
@slaweet @Efefefef I just update the PR based on the last comments right now everything is working and have the texts based on the same.
Please let me know if it is ok like this.

Efefefef:
🐛 
Should be bookmarked account
image

Efefefef:
`Uncaught (in promise) Error: Status 404` : Delegate not found is still not fixed

Please check the comment above. Why you put `bookmark` account, not `bookmarked` account ?

osvaldovega:
@Efefefef I update the bookmark labels with the correct text, I just checked with @slaweet.
About the error message in the console, that is a problem with the lisk core, I need to create a ticket with them it is not related to us because the problem is randomly, thats why I didn't saw it before submit the PR the last time and then you saw it now. it is not all the times that happen

Efefefef:
🐛 Create First Transaction from Initialize Lisk ID widget doesn't fill in the fields
@slaweet Should we show Account bookmarked button on Confirmation page if that was a bookmarked account?

slaweet:
@osvaldovega @Efefefef Account bookmarked button should be only displayed in the flow the user bookmarked the account. It should not be displayed the next time the user sends to that address.

Efefefef:
Ok guys then there is a bug
🐛 If you click Send to the acc from account's page then you'll see the bookmarked dialog shown

osvaldovega:
@Efefefef can you please check all the functionality related the send component and confirm that everything else is working as should be, in case something else it is not working properly can you please put it in just one update, then I don't need to to fix just one bug and after that you check again the code and find another one.

In case this is the last bug, can you please confirm this is the last thing to fix then I can proceed to fix it?

Efefefef:
👍 Good to go!

Efefefef:
🐛 Man you accidentally removed the link that prepopulates fields

PR #1821, opened 2019-03-14
Fetch all data for Votes tab - Closes #1820

### LINKED ISSUE #1820: Fetch all data for "Votes" tab on account page
Follow up to https://github.com/LiskHQ/lisk-hub/issues/1779, as it turned out additional API requests are needed, which means more work

renatomassao:
### What issue have I solved?

#1820

Created action that receives the votes, how many are show, and filter that fetches delegates data as needed.

### How has this been tested?

Go to `/explorer/accountsV2/2329183154524426169L` since this account has votes for some low ranked delegates.
- Go to votes tab;
- Scroll down and click on show more;
  - A loading state should show up while fetching the new data.
- Reload page and check that filtering by username also triggers the additional fetch.

### Review checklist


src/components/votes/votesTab.js
```diff
@@ -112,11 +147,17 @@ VotesTab.propTypes = {
...
     username: PropTypes.string,
     address: PropTypes.string,
+    rank: PropTypes.number,
+    productivity: PropTypes.number,
+    rewards: PropTypes.string,
+    vote: PropTypes.string,
   })),
   loading: PropTypes.array,
   t: PropTypes.func,
+  searchVotesDelegate: PropTypes.func,
```
osvaldovega: Do you think could be possible change this function names, I believe I know what this do, but it is more clear, **searchForDelegateVotes** or **searchDelegateVotes**, I think it is more clear in this way IMPO

src/actions/search.js
```diff
@@ -29,22 +29,61 @@ const searchDelegate = ({ publicKey, address }) =>
...
+    const liskAPIClient = getState().peers.liskAPIClient;
+    /* istanbul ignore else */
+    if (!liskAPIClient) return;
+    dispatch(loadingStarted(actionTypes.searchVotes));
+    const votes = await getVotes(liskAPIClient, { address, offset, limit })
+      .then(res => res.data.votes);
+
+    dispatch({
+      type: actionTypes.searchVotes,
+      data: { votes, address },
```
osvaldovega: could you change this dispatch to instead of trigger this function in this way dispatch it in this other way

example
```
dispatch(searchVotesDelegate(votes, { address, limit: 30 }))
```

src/components/votes/votesTab.js
renatomassao: This one is kind of tricky, because it's not for getting the votes of a delegate, it's for fetching data of the voted delegate, to make it clearer would need to maybe be **fetchVoteDelegateInfo**

src/actions/search.js
renatomassao: Actually removed this one, since it was causing trouble on the old version

PR #1822, opened 2019-03-15
Implement votes tab on account page - Closes #1779

### LINKED ISSUE #1779: Implement the Votes tab on the account page
Based on designs from https://github.com/liskhq/lisk-hub/issues/1754

renatomassao:
### What issue have I solved?

#1779 

Implemented Votes tab on another account page, that consumes the votes api to show the data, loading additional data for delegates will be handled in a follow up ticket: #1820

### How has this been tested?

Go to `/explorer/accountsV2/:address` of some account and check if:
- it has more than 30 it should appear the show more button at the bottom;
- it has no votes should show a message.

The filter should filter by username:
- If a username meets the criteria it should show up;
- if no username meets the criteria, a message shows up.

### Review checklist


i18n/locales/en/common.json
```diff
@@ -311,6 +313,7 @@
   "Password": "Password",
   "Paste": "Paste",
   "Pending...": "Pending...",
+  "Percentage of successfully forged blocks of when the delegate should have forged a block of transactions.": "Percentage of successfully forged blocks of when the delegate should have forged a block of transactions.",
```
osvaldovega: If I am not wrong this is not properly write, should be without the **of** between blocks and when, like this.

Percentage of successfully forged blocks **of** when the delegate should have forged a block of transactions
Percentage of successfully forged blocks when the delegate should have forged a block of transactions

src/components/votes/votesTab.css
```diff
@@ -0,0 +1,140 @@
...
+  overflow: hidden;
+  position: relative;
+
+  &.hasMore {
+    padding-bottom: 30px;
+  }
+
+  &.isLoading {
+    & .loadingOverlay {
+      background: rgba(255, 255, 255, 0.8);
```
osvaldovega: maybe you can use a constant in here

src/components/votes/votesTab.css
```diff
@@ -0,0 +1,140 @@
...
+    & :global(.empty-message) {
+      display: none;
+    }
+  }
+
+  & .showMore {
+    align-items: center;
+    bottom: 0;
+    background: var(--color-white);
+    box-shadow: 0 -1px 5px 0 rgba(30, 25, 77, 0.06);
```
osvaldovega: is better to use a constant in here

src/components/votes/votesTab.js
```diff
@@ -0,0 +1,128 @@
...
+
+  handleFilter({ target }) {
+    this.setState({
+      filterValue: target.value,
+    });
+  }
+
+  render() {
+    const { t, votes, loading } = this.props;
+    const { filterValue } = this.state;
```
osvaldovega: I think we need to check this with @slaweet the next Monday because I think they prefer not use destructuring in the render. I like it but I think they have this as an standard so maybe next Monday we can check it

src/components/votes/votesTab.js
```diff
@@ -0,0 +1,128 @@
...
+    const canLoadMore = filteredVotes.length > this.state.showing;
+    const isLoading = loading.length > 0;
+
+    return (
+      <BoxV2 className={`${styles.wrapper}`}>
+        <header>
+          <h1>{t('Voted delegates')}</h1>
+          <div className={`${styles.filterHolder}`}>
+            <InputV2
+              disabled={votes && votes.length === 0}
```
osvaldovega: I think this can be replace for **{votes && !votes.length}**

src/components/votes/votesTab.js
```diff
@@ -0,0 +1,128 @@
...
+        <main className={`${styles.results} ${canLoadMore ? styles.hasMore : ''} ${isLoading ? styles.isLoading : ''}`}>
+          <VotesTableHeader />
+          {
+            isLoading ? (
+              <div className={styles.loadingOverlay}>
+                <SpinnerV2 className={`${styles.loadingSpinner}`} />
+              </div>
+            ) : null
+          }
+          {filteredVotes.length > 0
```
osvaldovega: I think in here too.
you can just use filteredVotes.length.
By the way, maybe you can try to use the same way in all the files when open/close brackets for dont affect the indentation.
in line 60 you open brackets and then use the next line to code but in here you open brackets and the code next to the brackets, so it is not a real problem is just that is easier to read when you do the same and not change it very often and the indentation is better to follow. Same in lines 97 and 103
(really minor details nothing related to the code) 👍🏻

src/components/votes/votesTab.js
renatomassao: was discussed we will talk about it in the next technical meeting.

src/components/votes/votesTab.css
renatomassao: This one is because it's just for the mask when in the loading state, that we will probably change for the final design.

i18n/locales/en/common.json
renatomassao: Checked all tooltips with marketing team

src/components/votes/votesTab.css
osvaldovega: good

Issue #1830, opened 2019-03-19
Create Storybook structure


massao:
Create Storybook initial structure, with some parts of the styleguide and some components

PR #1850, opened 2019-03-22
Create storybook structure - Closes 1830

massao:
### What issue have I solved?

#1830 

Created initial structure of storybook, with own `package.json` so the webpack configuration doesn't conflict with our own, since storybooks uses webpack 4.
Added alias to webpack to be easier to reference the `components` folder inside stories. `Components` alias.
Created buttons stories with action addons as initial example.
Added storybook script to package.json to be easier to run storybook from the root folder.

**Improvements**
- Need to check which addons we want to use, and that are relevant to us. [addons](https://storybook.js.org/docs/addons/addon-gallery/);
- New components stories should be added as they are updated.

### How has this been tested?

Install dependencies of storybook on `/storybook` folder with `npm i`, after you can run `npm run storybook` at the root of the project or on the `/storybook` folder, and the storybook should open on a browser tab.

### Review checklist


package.json
```diff
@@ -35,8 +35,7 @@
     "eslint": "eslint ./",
     "pack": "npm install && npm run build && npm run clean-dist && npm run dist",
     "pack:win": "cmd /c npm install && npm run install-electron-dependencies && npm run clean-build && npm run copy-files && npm run build-prod && npm run build-electron && npm run clean-dist && npm run dist:win",
-    "storybook": "start-storybook -p 6006 -s ./src/",
-    "build-storybook": "build-storybook",
+    "storybook": "cd ./storybook/; npm run storybook",
```
slaweet: It would be nice to continue in the same folder after stopping storybook
```suggestion
    "storybook": "cd ./storybook/; npm run storybook; cd ..",
```

Efefefef:
@massao I have errors running storybook
`Failed to load preset: "/Users/andreiklimenok/lisk/hub/node_modules/@storybook/react/dist/server/framework-preset-cra.js"
ERR! TypeError: Cannot destructure property `createHash` of 'undefined' or 'null'.`

storybook/stories/buttons.stories.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+
+import { PrimaryButtonV2, SecondaryButtonV2, DangerButtonV2 } from 'Components/toolbox/buttons/button';
+
+storiesOf('Button', module)
+  .add('Primary Button', () => (
+    <React.Fragment>
```
slaweet: I suggest this wrapper (and others) to have some padding so that it doesn't look like this
Screen Shot 2019-03-26 at 11 26 33

massao: Created StoryWrapper Component that has a padding.

Issue #1905, opened 2019-04-05
Users' shouldn't be able to send 0 LSK


reyraa:
### Steps to reproduce

1. Navigate to the Send page
1. Fill the form with a valid address and 0 LSK and proceed
1. Finalize the transaction.

### Actual result
You'll receive this error:
![Screenshot 2019-04-05 at 15 44 02](https://user-images.githubusercontent.com/666685/55632025-a9274480-57b9-11e9-8cbc-8e3ffde12cfc.png)

### Expected result
Since 0 is an invalid amount, the users must not be able to proceed with 0 LSK. 

### Version
1.14.0

PR #1907, opened 2019-04-05
Fix send LSK navigation data form - Closes #1906 #1905

### LINKED ISSUE #1906: Fix send LSK navigation data form
The problem it is related, when you select any address (account) from the wallet table and want to send some LSK to this address.

### Steps to reproduce

1. From the Wallet table select any address (click)
2. Once the send LSK form page show up, type any amount and message, (the message it is option but is better for testing purposes). 
3. Do click in the "Go to Confirmation" button
4. Then from the confirmation button go back to the previous page
5. Then the amount and message fields are empty and not with the proper information. 

### Actual result

1. From the Wallet table select any address (click)
2. Once the send LSK form page show up, type any amount and message, (the message it is option but is better for testing purposes). 
3. Do click in the "Go to Confirmation" button
4. Then from the confirmation button go back to the previous page
5. Then the amount and message fields are empty and not with the proper information. 

### Expected result

1. From the Wallet table select any address (click)
2. Once the send LSK form page show up, type any amount and message, (the message it is option but is better for testing purposes). 
3. Do click in the "Go to Confirmation" button
4. Then from the confirmation button go back to the previous page
5. Then the amount and message fields are populated with the provided information. 

### Version

1.15.0

osvaldovega:
### What issue have I solved?

-- #1906 
-- #1905 

This PR include the fix for the send LSK form so know works properly.

From the Wallet table select any address (click)
Once the send LSK form page show up, type any amount and message, (the message it is option but is better for testing purposes).
Do click in the "Go to Confirmation" button
Then from the confirmation button go back to the previous page
Then the amount and message fields are populated with the provided information.

### How has this been tested?

Like this.
From the Wallet table select any address (click)
Once the send LSK form page show up, type any amount and message, (the message it is option but is better for testing purposes).
Do click in the "Go to Confirmation" button
Then from the confirmation button go back to the previous page
Then the amount and message fields are populated with the provided information.

### Review checklist

### DIFF
--- a/src/components/sendV2/form/form.js
+++ b/src/components/sendV2/form/form.js
@@ -67,18 +67,16 @@ class Form extends React.Component {
   }
 
   componentDidMount() {
-    this.ifDataFromPrevState();
-    this.ifDataFromUrl();
+    if (!Object.entries(this.props.prevState).length) this.ifDataFromUrl();
+    if (Object.entries(this.props.prevState).length) this.ifDataFromPrevState();
     this.checkIfBoormakedAccount();
   }
 
   ifDataFromPrevState() {
     const { prevState } = this.props;
-
-    if (prevState.fields && Object.entries(prevState.fields).length > 0) {
+    if (prevState.fields && Object.entries(prevState.fields).length) {
       this.setState({
         fields: {
-          ...this.state.fields,
           ...prevState.fields,
         },
       });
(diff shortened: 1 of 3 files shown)


src/components/sendV2/form/form.js
```diff
@@ -85,23 +85,22 @@ class Form extends React.Component {
 
   ifDataFromUrl() {
     const { fields = {} } = this.props;
-
     if (fields.recipient.address !== '' || fields.amount.value !== '' || fields.reference.value !== '') {
       this.setState(prevState => ({
         fields: {
           ...prevState.fields,
```
massao: isn't better to just use the `this.state` here?

osvaldovega: this is part of the solution I had to change it to work based on how the validation is being made so yes was intentional

massao: 👌

PR #1920, opened 2019-04-11
Create utilities to connect to btc - Closes #1895

### LINKED ISSUE #1895: Create utilities for connecting to BTC
This is a subtask of #1831.

The goal is to create utilities for connecting to BTC based on the utils from Lisk Mobile.

massao:
### What issue have I solved?

#1895

Copied the utilities from the lisk-mobile repository to be able to connect to the BTC node.
- Not saving nothing in the store right now, also not calling the utilities from any file.
- Had to change the code a little, due to we having the network switcher on Hub.
  - It will connect to the based on the netCode passed to the functions, by default it's 1 (testnet).
  - If connected to a custom lisk node it connect to the btc testnet.
- I was able to test the functions for fetching some account information, with the `extractAddress` and `getAccount` functions that are being exported temporarily from the `api/accounts.js` file.
- Also changed from plain fetch to the popsicle lib similar to we do with the lskService.

If you want to try the utilities, you can use some existing action and import the functions and call from there.

## Refactors needed
- See which api utilities are "shared" between multiple currencies and use the function mapper to map to the correct one, and have only one export point for the mapped function.
- ~Update to final Services API once it's done by mobile team.~

## Known issues
- The btc functions right now are all skipped on unit tests due to it using other APIs that we would need to mock, and since we also skipped the services API we don't have a example to use, maybe a point to be discussed on next technical meeting.

### How has this been tested?

Created unit tests to helper functions, and tested that we are able to connect and fetch some data from the btc `testnet`.
you can update the `accountDataUpdated` action to use the `btc.extractAddress` and `btc.getAccount` exported by `api/account`, to get the account summary from testnet.

### Review checklist


src/utils/api/btc/config.js
```diff
@@ -0,0 +1,12 @@
+import * as bitcoin from 'bitcoinjs-lib';
+
+const getBtcConfig = netCode => ({
+  isTestnet: netCode !== 0,
+  url: netCode !== 0 ? 'http://localhost:7201' : 'https://blockchain.info',
```
aydieneue:
```suggestion
  url: netCode !== 0 'https://ams1-bitcoin-test-001.ed148.net' : 'https://ams1-bitcoin-001.ed148.net'
```

src/utils/api/btc/account.js
```diff
@@ -0,0 +1,42 @@
+import * as bitcoin from 'bitcoinjs-lib';
+import Lisk from 'lisk-elements';
+import bip32 from 'bip32';
+import * as popsicle from 'popsicle';
+import getBtcConfig from './config';
+
+export const getSummary = (address, netCode = 1) => new Promise(async (resolve, reject) => {
+  try {
+    const config = getBtcConfig(netCode);
+    const response = await popsicle.get(`${config.url}/balance?active=${address}`)
```
aydieneue: This endpoint is different on the bridge -> https://github.com/LiskHQ/bitcoin-bridge#account-data-by-account-id

```suggestion
    const response = await popsicle.get(`${config.url}/account/${address}`)
```

src/utils/api/btc/account.js
```diff
@@ -0,0 +1,42 @@
...
+      .use(popsicle.plugins.parse('json'));
+    const json = response.body;
+
+    if (response) {
+      resolve({
+        address,
+        balance: json[address].final_balance,
```
aydieneue: again, new endpoint 🤗 

```suggestion
        balance: json.data.confirmed_balance,
```

src/utils/api/btc/transactions.js
```diff
@@ -0,0 +1,232 @@
...
+        resolve(json.data);
+      } else {
+        reject(json);
+      }
+    } catch (error) {
+      reject(error);
+    }
+  });
+
+export const create = ({
```
osvaldovega: I think this function could be split in more functions, there are a lot of code inside this function that do different things, so maybe those could be in a different place

massao: Yeah, I was discussing this with @aydieneue on Friday, but we are not really sure on how to split this function without turning this more complex or what kind of logic would make sense to not keep on this function and move to a separate function, since in the end all the logic are only used to create the transaction.

PR #1925, opened 2019-04-15
Create redux refactor proposal to store BTC data - Closes #1896

### LINKED ISSUE #1896: Create redux refactor proposal for storing BTC data
This is a subtask of #1831.

The goal is to create a document that describes:
- how redux should be structured to allow storing account and transactions for multiple currencies (BTC and LSK at the moment)
- what actions are needed for send tokens, get transactions, get one transaction

massao:
### What issue have I solved?

#1896

### Review checklist


docs/REDUX_STRUCTURE.md
```diff
@@ -0,0 +1,118 @@
...
+```
+- Rename `liskService` to just `service` so we can have all service information centralized in one place.  
+- New token accounts being already on the `accounts` structure.
+- Specific new token data grouped inside a `tokenKey` node.
+
+Ideally in the future we should end up with a structure similar to:
+```json
+// lisk-hub final store example
+{
+  "accounts": { "info": { "tokenKey": {} }, "followed": {}, "passphrase": "", "other account common info" },
```
slaweet: IMO it's better to make the `followed` an array of objects with `tokenKey` property. We can always filter `followed` by `tokenKey` in situations where we want to display only accounts of one `tokenKey`

With the current proposal of `followed` being an object with `tokenKey` keys and separate lists of accounts, it will be impossible to have a single list all tokens together with a given order where user can have their most important accounts from all tokens at the top, as planned here:
https://netguru.invisionapp.com/share/M5RKJHPQZKF#/screens/357779158

With the structure from Lisk Mobile, it would mean that the list has all LSK accounts and only after that all BTC accounts (or vice versa).

```suggestion
  "accounts": { "info": { "tokenKey": {} }, "followed":  [{ "tokenKey": "...", ...}, ...], "passphrase": "", "other account common info" },
```

slaweet: I also see a problem of `followed` being inside the `accounts` object, because unlike in Mobile, in Hub the followed accounts are not linked to the account and can be accessed on the dashboard in the guest mode when there is no user at all. Keeping `followed` inside `user` will make reducers for actions like `login` and `logout` more complicated because they will have to preserve this part of `account` store.

docs/REDUX_STRUCTURE.md
```diff
@@ -0,0 +1,118 @@
...
+The reducers would have to take in account the token type, or if no token is set, consider it as being common data and not putting inside a node, but directly on the root.  
+Something like:
+```javascript
+function genericReducer(state = {}, action) {
+  switch(action.type) {
+    case genericType: {
+      const hasToken = !!action.data.activeToken || false;
+      const data = hasToken
+        ? { ...state, [action.data.activeToken]: { ...action.data.payload } }
+        : { ...state, ...action.data.payload, };
```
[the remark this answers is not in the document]
massao: Oh this is just miss explained, the idea is not to have a function called genericReducer, but a reducer that should work for any token passed, and creating a node with the `tokenKey` as name, so we only have one reducer that work for any kind of currency passed.

docs/REDUX_STRUCTURE.md
massao: So something like this would be better? @slaweet 
```suggestion
  "accounts": { "info": { "tokenKey": {} }, "passphrase": "", "other account common info" },
  "followed": [],
```

osvaldovega: I think is a real good point what Vit mentioned, for the follow accounts we can handle it separated from accounts same for the array of objects with the tokenKey it will be better as Vit mentioned

docs/REDUX_STRUCTURE.md
```diff
@@ -0,0 +1,120 @@
...
+In the case of Lisk-Hub we could have initialy a structure that integrate some of the changes just for other currencies, while we don't update the current LSK structure.  
+To have a better structure for new currencies, while also keeping the LSK token working without too much work, something like:
+```json
+// lisk-hub intermediate store example
+{
+  "..."
+  "account": { "LSK account info" },
+  "accounts": { "info": { "BTC": {} }, "followed": {} },
+  "BTC": { "BTC specific data" },
+  "service": {},
```
yasharAyari: This reducer needs to support multi token. the structure of this reducer should be similar accounts reduce. Also, lets the app to store dynamic fee system for BTC.

yasharAyari: the service data structure should be like the below example :
 service: { dynamicFees: {} , priceTicker: { "BTC": {}, "LSK": {} }};

Issue #1849, opened 2019-03-21
Implement disconnect detection for Trezor Model T


slaweet:
This is a subtask of #1319.
In this task, it should not be available to end users yet (e.g. set some local storage flag). There will be a final task to enable it.

The goal is to implement disconnect detection for Trezor Model T

osvaldovega:
this feature is implemented in PR #1949

Issue #1847, opened 2019-03-21
Implement vote with Trezor Model T


slaweet:
This is a subtask of #1319. 
In this task, it should not be available to end users yet (e.g. set some local storage flag). There will be a final task to enable it.

The goal is to implement vote with Trezor Model T

osvaldovega:
this feature is implemented in PR #1949

PR #1949, opened 2019-04-18
Implement send LSK with Hardware Wallet Trezor Model T - Closes #1845 #1849 #1847

### LINKED ISSUE #1845: Implement send LSK with Trezor Model T
This is a subtask of #1319. 

The goal is to implement send LSK with Trezor Model T

osvaldovega:
### What issue have I solved?

-- #1845 
-- #1847 
-- #1849 

For this implementation the same code for send LSK, using HW Ledger Nano S worked with the only difference that for the user of Trezor Model T you need to use the Trezor device ID and for Ledger you need to get the device ID from the key.

In summary for make this to work I had to update the TrezorLogin component to properly update the account reducer to provide the device ID, and with this change now I can use the same functionality as before but using the Trezor ID.

### How has this been tested?

1. Run the app with Electron
2. Open the developers mode and then add to the local storage one more item with key = trezor and value = true, after this refresh the app.
2. Login using the Trezor model T
3. Select the Trezor account
4. Go to Wallet -> Send LSK
5. Type any LSK address/account and amount
6. Go to the confirm screen after press the confirm button
7. Then here you should check the Trezor device for accept the transaction or cancel it
8. Then the final screen will show the result of the transaction

### Review checklist


src/components/sendV2/summary/summary.js
```diff
@@ -144,16 +144,17 @@ class Summary extends React.Component {
...
 
   render() {
+    const { account } = this.props;
     const { secondPassphrase, isHardwareWalletConnected } = this.state;
     let isBtnDisabled = secondPassphrase.hasSecondPassphrase && secondPassphrase.isValid !== '' && !secondPassphrase.isValid;
     isBtnDisabled = !isBtnDisabled && isHardwareWalletConnected;
 
     const confirmBtnMessage = isHardwareWalletConnected
-      ? this.props.t('Confirm on Ledger')
+      ? this.props.t(`Confirm on ${account.hwInfo.deviceModel}`)
```
slaweet: For the translations to work properly it should be
```suggestion
      ? this.props.t('Confirm on {{deviceModel}}', { deviceModel: account.hwInfo.deviceModel })
```

src/components/sendV2/summary/summary.js
```diff
@@ -144,16 +144,17 @@ class Summary extends React.Component {
...
       : this.props.t('Send {{amount}} LSK', { amount: this.props.fields.amount.value });
 
     const title = isHardwareWalletConnected
-      ? this.props.t('Confirm transaction on Ledger Nano S')
+      ? this.props.t(`Confirm transaction on ${account.hwInfo.deviceModel}`)
```
slaweet: For the translations to work properly it should be
```suggestion
      ? this.props.t('Confirm transaction on {{deviceModel}}', { deviceModel: account.hwInfo.deviceModel })
```

src/utils/hwWallet.js
```diff
@@ -10,37 +10,57 @@ import { getTransactionBytes, calculateTxId, getBufferToHex } from './rawTransac
...
+import { accountLoggedOut } from '../actions/account';
+import { dialogDisplayed, dialogHidden } from '../actions/dialog';
+import Alert from '../components/dialog/alert';
 
 const util = require('util');
 
 const { ipc } = window;
 
+const handleConnect = () => {
+  store.dispatch({
```
slaweet: I know it's not new in this PR, but IMO we shouldn't use `import store from '../store';`, because it causes lot of problems with unit tests when it gets imported in completely unrelated unit tests.

So far, the redux architecture we used was that if something needs to `dispatch` an action, it should be a `middleware`, and utils should be redux-independent.

src/utils/hwWallet.js
```diff
@@ -10,37 +10,57 @@ import { getTransactionBytes, calculateTxId, getBufferToHex } from './rawTransac
...
+    type: actionTypes.settingsUpdated,
+    data: { isHarwareWalletConnected: true },
+  });
+  store.dispatch(errorToastDisplayed({ label: HW_MSG.LEDGER_CONNECTED }));
```
slaweet: here it says `HW_MSG.LEDGER_CONNECTED`, but this function `handleConnect` is used also for `ipc.on('trezorConnected', handleConnect);` below.

src/utils/hwWallet.js
```diff
@@ -10,37 +10,57 @@ import { getTransactionBytes, calculateTxId, getBufferToHex } from './rawTransac
...
-  // });
+  ipc.on('ledgerConnected', handleConnect);
+  ipc.on('ledgerDisconnected', () => handleDisconnect('Ledger Nano S'));
 
-  // ipc.on('ledgerDisconnected', () => {
-  //   console.log('ledgerDisconnected');
-  //   store.dispatch(errorToastDisplayed({ label: HW_MSG.LEDGER_DISCONNECTED }));
-  // });
+  ipc.on('trezorConnected', handleConnect);
+  ipc.on('trezorDisconnected', () => handleDisconnect('Trezor Model T'));
```
slaweet: IMO here is an unnecessary duplication of having separate events for ledger and trezor. 
It can already on the electron side dispatch the same event with model, e.g.:
change
https://github.com/LiskHQ/lisk-hub/blob/29871a8a23f32391ac9f06adb4a9ab7492feaebe/app/src/ledger.js#L95
to 
```
win.send({ event: 'hwWalletDisconnected', value: {model: 'Trezor Model T' });
```

And then here:
```suggestion
  ipc.on('trezorDisconnected', (event, { model }) => handleDisconnect(model));
```

Relevant docs for Electron ipc: https://electronjs.org/docs/api/ipc-main#ipcmainonchannel-listener

osvaldovega: thx Vit I made the correction you are right about it

src/utils/hwWallet.js
osvaldovega: fixed

src/components/sendV2/summary/summary.js
osvaldovega: right, I fixed

src/components/sendV2/summary/summary.js
osvaldovega: fixed

src/utils/hwWallet.js
osvaldovega: ok Vit right now I just put the login in the login middleware file so right now it is better

src/utils/hwWallet.js
```diff
@@ -16,31 +16,10 @@ const util = require('util');
...
-  });
-
-  ipc.on('trezorDisconnected', (event, data) => {
-    store.dispatch(errorToastDisplayed({
-      label: util.format(HW_MSG.TREZOR_DISCONNECTED, data.model, data.label),
-    }));
-  });
-
   ipc.on('trezorButtonCallback', (event, data) => {
     store.dispatch(infoToastDisplayed({
```
slaweet: There are still some more `store.dispatch`s in this util file. IMO it doesn't make sense why are some `ipc.on`s here and some in the middleware and all `ipc.on`s should be together in the middleware.

src/store/middlewares/login.js
```diff
@@ -17,30 +14,29 @@ const loginMiddleware = store => next => (action) => { // eslint-disable-line ma
...
         });
 
-        ipc.on('ledgerConnected', () => {
+        ipc.on('hwConnected', (event, { model }) => {
           store.dispatch({
             type: actionTypes.settingsUpdated,
             data: { isHarwareWalletConnected: true },
           });
-          store.dispatch(errorToastDisplayed({ label: HW_MSG.LEDGER_CONNECTED }));
+          store.dispatch(errorToastDisplayed({ label: `${model} connected` }));
```
slaweet: I would expect green (or neutral) toast here. There is nothing "erroneous" about this event
```suggestion
          store.dispatch(successToastDisplayed({ label: `${model} connected` }));
```

src/store/middlewares/login.js
```diff
@@ -17,30 +14,29 @@ const loginMiddleware = store => next => (action) => { // eslint-disable-line ma
           data: { isHarwareWalletConnected: false },
         });
 
-        ipc.on('ledgerConnected', () => {
+        ipc.on('hwConnected', (event, { model }) => {
```
slaweet: IMO it doesn't make sense why listening to hwWallet events is in `login` as it has nothing to do with login. It should be rather in `src/store/middlewares/hwWallet.js`

yasharAyari: I think it should be part of the login process but it doesn't need to be part of middleware and it should happen inside the login action.

src/store/middlewares/login.js
```diff
@@ -17,30 +14,29 @@ const loginMiddleware = store => next => (action) => { // eslint-disable-line ma
           data: { isHarwareWalletConnected: false },
```
yasharAyari: you don't need to store isHarwareWalletConnected value in the redux store because it doesn't use in any presetional component. So, dispatching settingsUpdated action here is completely unnecessary.

src/store/middlewares/login.js
yasharAyari: I think it should be part of the login process but it doesn't need to be part of middleware and it should happen inside the login action.

src/store/middlewares/login.js
yasharAyari: I think it should be part of the login process but it doesn't need to be part of middleware and it should happen inside the login action.

slaweet: Listening to `hwConnected` event **cannot** be inside `login` action. We first need to detect/display that the device was connected, and only then the user can log in.

src/store/middlewares/login.js
slaweet: Of course, it is used in a presentational component. Specifically here:
https://github.com/LiskHQ/lisk-hub/blob/1736e98a13b8b5cc78207f8c2db5ebf5fd346f31/src/components/loginV2/loginV2.js#L276

yasharAyari: still, it doesn't make sense to store this value in the setting reducer.

slaweet: I agree, but it's way beyond the scope of this PR.

PR #1952, opened 2019-04-23
Create basic logic for BTC peer store - Closes #1946

### LINKED ISSUE #1946: Create redux logic that can connect to both LSK and BTC
This is a subtask of #1831.

Should change how activePeer is used

slaweet:
### What issue have I solved?

#1946

I created a new reducer called `network`, which stores all data that Hub needs to connect to networks (LSK and BTC at the moment). All we need to store is:

- network code (mainnet/testnet/customNode)
- only for LSK: address (one of hubXX.lisk.io for mainnet, any other for custom node)
- only for LSK: nethash (for custom node, it can be abything, so it's obtained form API call)

Everything for TBC can be derived from network code.

There are `TODO` comments in the old `peers` redux store and actions that it should be removed in the future after all code is updated to use the new store in follow-up tickets.

### How has this been tested?

You can enable it by running this in console:
```
localStorage.setItem('btc', true);
```
Then check with browser redux extension that new actions are dispatched and new `network` store is created.

### Review checklist


osvaldovega:
@slaweet I checked and looks easy to understand/maintain we still have all the information that we need and in case that we need to add more is properly separated. I would say it is ok

src/actions/network/lsk.js
```diff
@@ -0,0 +1,48 @@
...
+        if (error && error.message) {
+          dispatch(errorToastDisplayed({ label: i18next.t(`Unable to connect to the node, Error: ${error.message}`) }));
+        } else {
+          dispatch(errorToastDisplayed({ label: i18next.t('Unable to connect to the node, no response from the server.') }));
+        }
+      });
+    } else if (data.code === networks.testnet.code) {
+      dispatch(generateAction(data, {
+        nethash: Lisk.APIClient.constants.TESTNET_NETHASH,
+        address: pickRandomArrayItem(networks.testnet.nodes),
```
yasharAyari: it needs to change to nodeUrl instead of the address.

src/actions/network/lsk.js
```diff
@@ -0,0 +1,48 @@
...
+  },
+  type: actionTypes.networkSet,
+});
+
+const pickRandomArrayItem = a => a[Math.floor(Math.random() * a.length)];
+
+/* eslint-disable-next-line import/prefer-default-export */
+export const networkSet = data =>
+  async (dispatch) => {
+    if (data.code === networks.customNode.code) {
```
yasharAyari: Change the code here to the name of the network.

src/actions/account.js
```diff
@@ -188,7 +189,11 @@ export const loadAccount = ({
   isSameAccount,
 }) =>
   (dispatch, getState) => {
-    const liskAPIClient = getState().peers.liskAPIClient;
+    // TODO remove the localStorage condition after peers store is removed.
+    // This is a show-case of how liskAPIClient should be obtained with the new network store
+    const liskAPIClient = localStorage.getItem('btc') ?
```
yasharAyari: liskAPIClient should move to the lisk API file

src/actions/network/index.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+  describe('networkSet', () => {
+    it('should create networkSet action with name and token', () => {
+      const { name } = networks.testnet;
+      networkSet({ name })(dispatch);
+      expect(dispatch).toHaveBeenNthCalledWith(1, {
+        data: {
+          name,
+          token: tokenMap.BTC.key,
+          network: {},
```
yasharAyari: why do you store an empty object as a network in this action? this action only stores a value of a constant in a redux reducer.

src/actions/network/index.js
```diff
@@ -0,0 +1,24 @@
+import { networkSet as btcNetworkSet } from './btc';
+import { networkSet as lskNetworkSet } from './lsk';
+import actionTypes from '../../constants/actions';
+
+
+export const networkSet = data =>
+  (dispatch) => {
+    [
+      btcNetworkSet,
+      lskNetworkSet,
```
yasharAyari: we don't need to store a network object for any assets expect lisk. So, this action only set network config for lisk.

slaweet: I store the network object because we wanted to make things modular and generic. At the moment we don't need to store anything for BTC, but it is only because we don't allow BTC custom node. If we will want to have BTC custom node (e.g. for e2e tests) in the future, it will be easy to add.

I think it will be useful also for any other token in the future as we don't know what might be needed to store for other tokens.

PR #1955, opened 2019-04-25
Implement storing BTC transactions data - Closes #1945

### LINKED ISSUE #1945: Implement storing BTC transactions data
This is a subtask of #1831.

- adjust transactions actions (including 1 transaction) to use both LSK and BTC

slaweet:
### What issue have I solved?

#1945

I changed `getTransactions` API util to be token-agnostic and to get token-specific `getTransactions` (BTC or LSK) based on whether the passed `address` parameter is BTC or LSK address.

Analogically, I changed `getSignleTransaction` to be token-agnostic based on whether the passed `id` parameter is BTC or LSK id.

Note that currently it is based on branch `1946-logic-for-btc-peer`. After that branch PR is merged, I will change the base branch of this PR to `development`

### How has this been tested?

1. Go to  https://jenkins.lisk.io/test/lisk-hub/PR-1955/
2. log it to testnet
2. Set `localStorage.setItem('btc', true);`
1. Go to  https://jenkins.lisk.io/test/lisk-hub/PR-1955/#/explorer/accounts/n3aZt7uZhnBeC9quq6btKyC8qXvskEiE1B
3. see there are BTC transactions loaded.
1. Go to https://jenkins.lisk.io/test/lisk-hub/PR-1955/#/explorer/transactions/553787aa91a55e0385beb81a0a549f02cb3be8492b452fe8b09c1729165a8362
3. see there a BTC transaction details.

Note that there will be another ticket to make the page show everything properly.

### Review checklist


src/utils/api/lsk/transactions.js
```diff
@@ -0,0 +1,36 @@
...
+  apiClient, address, limit, offset, type = undefined,
+  sort = 'timestamp:desc', filter = txFilters.all, customFilters = {},
+}) => {
+  const params = {
+    limit,
+    offset,
+    sort,
+  };
+
+  if (type !== undefined) params.type = type;
```
osvaldovega: I think you can just put like
```suggestion
  if (type) params.type = type;
```

slaweet: I cannot, because send type value is `0` and it wouldn't set it https://github.com/LiskHQ/lisk-hub/blob/0425e51423a85feb8be5cf349f69d63e38fee62d/src/constants/transactionTypes.js#L2

osvaldovega: oh ok good point, then everything clear

src/utils/api/transactions.js
```diff
@@ -21,53 +21,34 @@ export const send = (
...
+const getTokenFromTransactionId = id => (
+  // TODO remove the localStorage condition after BTC features is enabled.
+  // eslint-disable-next-line no-nested-ternary
+  localStorage.getItem('btc') ?
+    (id && id.length === 64 ? tokenMap.BTC.key : tokenMap.LSK.key) :
+    /* istanbul ignore next */
+    tokenMap.LSK.key
+);
+
+export const getTransactions = ({ offset = 0, limit = 25, ...params }) => (
```
yasharAyari: you need to get the active token as input in the getTransactions function

src/utils/api/btc/transactions.js
```diff
@@ -1,89 +1,83 @@
-import * as bitcoin from 'bitcoinjs-lib';
-import * as popsicle from 'popsicle';
+import bitcoin from 'bitcoinjs-lib';
```
massao: This still works like this?? Because I remember that this version of `bitcoinjs-lib` don't export a default, so it was needed to import as `* as bitcoin`

slaweet: Interesting, I had a problem with it at first, but now it seems fine. It's changed because I got the latest version from Altay that is compatible with the latest bitcoin-bridge.

slaweet:
There is a number of TODO comments referring to things that need to be done in the future when enabling the BTC feature to the end user. 
For the future reference (as suggested by @osvaldovega ), here's a command to find them all:
```
git grep TODO | grep -i btc
```

src/actions/account.js
```diff
@@ -264,8 +264,9 @@ export const updateAccountDelegateStats = account =>
   async (dispatch, getState) => {
     const liskAPIClient = getState().peers.liskAPIClient;
     const { address, publicKey } = account;
+    const networkConfig = getState().network;
     const transaction = await getTransactions({
```
yasharAyari: You need to pass token.active to the getTransactions method

src/actions/search.js
```diff
@@ -11,9 +11,10 @@ import { updateWallet } from './wallets';
 const searchDelegate = ({ publicKey, address }) =>
   async (dispatch, getState) => {
     const liskAPIClient = getState().peers.liskAPIClient;
+    const networkConfig = getState().network;
     const delegates = await getDelegate(liskAPIClient, { publicKey });
     const transactions = await getTransactions({
```
yasharAyari: You need to pass token.active to the getTransactions method

src/actions/account.js
slaweet: I think I shouldn’t pass token.active there, because the two functions are both for delegates (`updateAccountDelegateStats` and `searchDelegate`), so they should never be called with token=BTC

src/actions/search.js
slaweet: same here

src/actions/account.js
yasharAyari: You need to read key of the `lisk` token from token constant list and pass it as an active token value instead of getting that from the redux store.

slaweet: @yasharAyari can you be more specific what you mean, because what you're saying is what I think I did after your comment 5 days ago in this commit https://github.com/LiskHQ/lisk-hub/pull/1955/commits/0fd989b6795d7874093dca995a9412b4371d7cc1

src/utils/api/btc/transactions.js
```diff
@@ -1,89 +1,83 @@
...
+    data.recipientId = address;
     data.amount = output.satoshi;
   }
 
   return data;
 });
 
-export const get = ({
-  id,
+export const getTransactions = ({
```
yasharAyari: I don't see any needs for a separated `getTransactions` and `getSingleTransaction`. why did you implement them instead of a generic get method?

slaweet: We talked about it with @yasharAyari and concluded that a generic get method works in Lisk mobile but in Hub there is more logic for filtering transactions, so it's cleaner to keep them separate. 

There will be a refactor needed to filtering transactions https://github.com/LiskHQ/lisk-hub/issues/1988

PR #1978, opened 2019-05-02
Fix Ledger detection - Closes #1976

### LINKED ISSUE #1976: Ledger not being recognized if connected after lisk-hub is open.
### Steps to reproduce

1. Open Lisk-hub.
1. Set LocalStorage item `trezor = true`
1. Connect and unlock Ledger
1. Ledger isn't recognize

### Actual result

After implementing support for Trezor Wallets, Ledger isn't being recognized if it's not unlocked, and inside the lisk app before opening the lisk-hub

### Expected result

Ledger should be able to be unlocked after the lisk-hub application is open.

### Version

1.17

massao:
### What issue have I solved?

#1976 

Updated how the Ledger connection was being detected, to only check if it is connected, instead of checking if the Lisk App is open, and added a function to be able to update the device on the open App screen, passing an event through ipc.  
And handling going back on the multistep component if the devices are disconnected.

**Known Issue**: Detecting that the app was opened is buggy, because of how the events of Ledger are recognize by the HID, every time the app is open or closed, it sends a remove and add event to the USB

### How has this been tested?

1. Run this build on electron
1. Go to `/hw-wallet-login-v2`
1. Connect and unlock Ledger
1. Should go to the page asking for opening the app on ledger.
1. Opening the app should go to select account page.

### Review checklist


src/actions/hwWallets.js
```diff
@@ -0,0 +1,12 @@
+import actionTypes from '../constants/actions';
+
+/**
+ * An action to dispatch a device list to HwWallets
+ *
+ */
+export const devicesListUpdated = data => ({
+  data,
+  type: actionTypes.devicesListUpdate,
```
osvaldovega: I don't know if this is a typo but looks like the function name is past time a this function action not **devicesListUpdated**.

In the other hand I dont know if should be something like updateDevicesList... what do you think?

massao: Update to follow the same standard as we use in other actions and reducers

PR #1989, opened 2019-05-08
Implement btc bookmarks - Closes #1962

### LINKED ISSUE #1962: Implement bookmarks for BTC
This is a subtask of #1831.

The goal is to implement bookmarks for BTC.

Specification and designs to be added.

massao:
### What issue have I solved?

#1962

- Updated followed accounts to be able to store LSK and BTC addresses;
- Updated all components using the followed accounts;
- Created `flattenFollowedAccounts` to generate a single level array with all the accounts;
- Updated all unit tests related to followed accounts;
- Removed components that were using followed accounts but weren't being used on the application;
- Had to put a validation on `walletTab` component to hide account details when account is BTC, should be removed/improved on #2003.

### How has this been tested?

1. Enable BTC with `localStorage.setItem('btc', true)`;
2. Go to any BTC account page. eg `/explorer/accounts/mkakDp2f31btaXdATtAogoqwXcdx1PqqFo`;
3. Add to bookmark through the dropdown;
4. Go to dashboard and check that the account shows up on the followed accounts module;
5. Try editing and removing the bookmark.

### Review checklist


massao:
> * Previously we were storing `publicKey` of accounts to be ready for the new address system. Now only `address` is stored, which means that once lisk core team implements the new address system, all users of Hub will lose all bookmarks. I would like @reyraa  to confirm if this is ok.
This one is my bad, I ended removing it, because it was not being used anywhere, but since it's to be used in the future, I'll put it back.

About the other two points, I think that @reyraa or @yasharAyari could update the ticket.

PR #2003, opened 2019-05-13
Implement minimal BTC wallet page - Closes #1888

### LINKED ISSUE #1888: Implement minimal BTC wallet page
This is a subtask of #1831. 

The goal is to adjust the wallet page so that it can display BTC transactions (Can be a separate URL: `/wallet/` vs. `/wallet/btc`)

It this subtask it's not needed to have the LSK/BTC switch dropdown and balance chart as in:
https://netguru.invisionapp.com/share/ZCRCP30E2SR#/screens/356008055

slaweet:
### What issue have I solved?

#1888

- I updated `walletTransactionsV2` HOC to get data based on `activeToken`.
- I updated various components not to show if token is BTC

There are two refactoring changes not essential to the problem at hand
- 🔥 Remove functionMapper to make code cleaner https://github.com/LiskHQ/lisk-hub/pull/2003/commits/d2f8280c1c31995d60bd8543935533ff272a23d2
- ♻️ Move updateTransactionsIfNeeded to transactions actions https://github.com/LiskHQ/lisk-hub/pull/2003/commits/02b2ae4fe29f48bc802ded7bceb5cd436ee2fe24
### How has this been tested?

- Go to wallet
- Open React dev tools (because of https://stackoverflow.com/a/44318447/2080716)
- Enable BTC feature `localStorage.setItem('btc', true);`
- Set BTC active token by running this in console `$r.store.dispatch({type:'SETTINGS_UPDATED', data: { token: { active: 'BTC', list: {BTC: true, LSK: true} }}}); ` Note that React dev tools have to be active for `$r` to be defined.
- Switch active token back to LSK  by running this in console `$r.store.dispatch({type:'SETTINGS_UPDATED', data: { token: { active: 'LSK', list: {BTC: true, LSK: true} }}}); `. 

### Review checklist


src/actions/transactions.js
```diff
@@ -307,7 +307,30 @@ export const sent = ({
         fee: Fees.send,
         type: transactionTypes.send,
         asset: { data },
+        token: 'LSK', // TODO use the passed token once implemented
       }));
       dispatch(passphraseUsed(passphrase));
     }
   };
+
+export const updateTransactionsIfNeeded = ({ transactions, account }, windowFocus) =>
```
yasharAyari: this method is completely unnecessary and confusing. you can use other methods instead of this.

src/actions/account.js
```diff
@@ -274,9 +243,13 @@ export const updateAccountDelegateStats = account =>
       token, networkConfig, address, limit: 1, type: transactionTypes.registerDelegate,
     });
     const block = await getBlocks(liskAPIClient, { generatorPublicKey: publicKey, limit: 1 });
-    dispatch(delegateStatsLoaded({
-      lastBlock: (block.data[0] && block.data[0].timestamp) || '-',
-      txDelegateRegister: transaction.data[0],
+    dispatch(accountUpdated({
+      token: 'LSK',
```
yasharAyari: this method should update the balance of the active token, not only 'LSK'

slaweet: This method is `updateAccountDelegateStats`. It doesn't update balance, but delegate stats, which exist only for LSK.

src/actions/transactions.js
slaweet: I didn't create this function, I just moved it from `src/actions/account.js` when I was cleaning up that file a bit. I agree it is confusing, but I'm not sure if I can just remove it and not break anything. Either way, it's out of the scope of this pr.

slaweet:
@yasharAyari I solved some problem, now my console is clean:
Screenshot 2019-05-15 at 12 53 13

Please check again.

src/actions/account.js
slaweet: I changed it to use the token variable that is already defined in the action. https://github.com/LiskHQ/lisk-hub/pull/2003/commits/6c323489a54f00d2dc63833637171341489067da

src/components/transactionsV2/transactionRowV2.js
```diff
@@ -50,32 +52,46 @@ class TransactionRowV2 extends React.Component {
...
-          </div>
         <div className={`${grid['col-sm-2']} ${grid['col-lg-2']} transactions-cell`}>
           <div className={`${styles.status} ${!isConfirmed ? styles.showSpinner : styles.showDate}`}>
             <SpinnerV2 completed={hasConfirmations} label={props.t('Pending...')} />
             <DateTimeFromTimestamp time={props.value.timestamp} token={props.value.token} />
           </div>
         </div>
         <div className={`${grid['col-sm-1']} ${grid['col-lg-2']} transactions-cell`}>
-          <LiskAmount val={props.value.fee}/>&nbsp;{props.value.token || props.t('LSK')}
+          <LiskAmount val={props.value.fee}/>&nbsp;{props.value.token}
```
yasharAyari: The token shouldn't be part of transaction list data. You must get from setting store if it is needed

slaweet: I think we both want to avoid unnecessary extra complexity and having a token part of the transaction object causes less complexity than having an extra token prop on all components that render transaction amount or have such child. We already have a function the modifies each transaction object for BTC https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/utils/api/btc/transactions.js#L15 , so IMO it is quite clean to have a similar function also for LSK (and for potential future tokens).

### If the token is part of the transaction object, these files need to have extra complexity:
- https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/utils/api/lsk/transactions.js
- https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/utils/api/btc/transactions.js

### If the token is passed separately into all components, these files need to have extra complexity (extra prop):
- https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/components/singleTransactionV2/index.js
- https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/components/singleTransactionV2/singleTransactionV2.js
- https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/components/transactionsV2/transactionDetailViewV2/transactionDetailViewV2.js
- https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/components/transactionsV2/amountV2.js
- https://github.com/LiskHQ/lisk-hub/blob/b51ca76a4f5306fbff57074ef727571024f79eca/src/components/transactionsV2/transactionRowV2.js

To sum up, they are two different approaches to prevent some complexity in the code, each has some pros and cons. The differences are not super huge, so I suggest to keep the already implemented approach as it involves no extra work.

yasharAyari: First of all, by adding the active token to each transaction row you store data which is dynamic in two different places. Then, every time the active token changes you need to up two different places.
On the other hand, the whole concept of global state management was introduced to fix the problem that you encounter. So you there is no need to inject active token into each transaction row because as soon as a user changes the active token the whole transaction list will be fetched again.

slaweet: > First of all, by adding the active token to each transaction row you store data which is dynamic in two different places. Then, every time the active token changes you need to up two different places.

This is not true at all. Token of a transaction is not dynamic. If active token changes, an LSK transaction is still LSK transaction. I don't need to update anything extra.

> On the other hand, the whole concept of global state management was introduced to fix the problem that you encounter. So you there is no need to inject active token into each transaction row because as soon as a user changes the active token the whole transaction list will be fetched again.

This is also not true at all. The problem is not about state management, state management works the same no matter if I use activeToken or token inside transaction. If I don't pass token (LSK/BTC) into each row, I cannot display that the amount of the transaction is in LSK or BTC.
Screenshot 2019-06-05 at 10 56 37

slaweet: @yasharAyari, on a more personal note, reading your comment above makes me feel like you're telling me what to do while you have no clue what you're talking about, which makes me quite upset. I don't mean to offend you, but I want to express how I feel about this discussion.

slaweet:
Follow-up refactoring ticket created https://github.com/LiskHQ/lisk-hub/issues/2127

PR #2005, opened 2019-05-13
Implement success/error step 3 of send LSK/BTC - closes #1930

osvaldovega:
### What issue have I solved?

-- #1930 

### How have I implemented/fixed it?

For this implementation I took as based the mobile code as in there this functionality is already implemented.

The **send** function changed to properly receive the parameters from LSK and BTC and then based on that check the active token and send the parameters to the correct API function.

Now the **send** function do 2 things **create** the transactions (Lisk elements -> transfer) and then **broadcast** the transaction.

### How has this been tested?

For test the LSK send functionality I used the send LSK form and confirm that the transaction successfully submit and show up in the wallet transactions table.

For the BTC is necessary mock the BTC data and submit the transaction. Once that the part 1 and 2 of the send request is done supporting BTC, the whole flow needs to be tested.

### Review checklist
- The PR follows our [Test guide](/LiskHQ/lisk-hub/blob/development/docs/TEST_GUIDE.md)
- The PR follows our [CSS guide](/LiskHQ/lisk-hub/blob/development/docs/CSS_GUIDE.md)

PR #2019, opened 2019-05-15
Simplify transaction actions - Closes #2018

### LINKED ISSUE #2018: Simplify transaction actions
## Curent situation
There are too many transaction actions:
```
$ grep export src/actions/transactions.js
export const cleanTransactions = () => ({
export const transactionAdded = data => ({
export const testExtensions = () => ({
export const transactionsFilterSet = ({
export const transactionsUpdateUnconfirmed = ({ address, pendingTransactions }) =>
export const loadTransactionsFinish = accountUpdated =>
export const loadTransactions = ({ publicKey, address }) =>
export const transactionsRequested = ({
export const loadLastTransaction = address => (dispatch, getState) => {
export const loadTransaction = ({ id }) =>
export const transactionsUpdated = ({
export const sent = ({
```
The file is ove 300 lines

## Expected situation
The actions should be simplified, file size reduced, duplicates identified.
The actions that are really necessary should be documented why they're needed.

slaweet:
### What issue have I solved?

#2018

I went through all actions exported from https://github.com/LiskHQ/lisk-hub/blob/development/src/actions/transactions.js and with each did one of the following:
- Wrote comment above it that explains why we need that action
- Removed the action as duplicate and updated code that used it to use another action
- I didn't touch the `sent` action as it's currently worked on in https://github.com/LiskHQ/lisk-hub/pull/2005

### How has this been tested?

Check that all tests still pass.

### Review checklist


src/actions/transactions.js
```diff
@@ -311,3 +267,33 @@ export const sent = ({
...
+/**
+ * This action is used in account middleware to update transactions when new block was created.
+ * TODO this action should be likely removed when account middleware is cleaned up in
+ * https://github.com/LiskHQ/lisk-hub/issues/2021
+ *
+ * @param {Object} param.transactions - the transaction object with pending and confirmed list
+ * @param {Object} param.account - an object with active account
+ * @param {Boolean} windowFocus - flag that says if the Lisk Hub window is focused
+ */
+export const updateTransactionsIfNeeded = ({ transactions, account }, windowFocus) =>
```
yasharAyari: this function is completely unnecessary and it has a very inappropriate name and it needs to be removed. Move the condition to the account then you `transactionsUpdated` inside the account action or middleware.

src/actions/transactions.js
```diff
@@ -1,115 +1,51 @@
...
+ * @param {String} params.address - address of the account to fetch the transactions for
+ * @param {Number} params.limit - amount of transactions to fetch
+ * @param {Number} params.offset - index of the first transaction
+ * @param {Number} params.filter - one of values from src/constants/transactionFilters.js
+ * @param {Object} params.customFilters - object with filters for the filer dropdown
+ *   (e.g. minAmount, maxAmount, message, minDate, maxDate)
+ */
 export const transactionsRequested = ({
-  address, limit, offset, filter, customFilters = {},
+  address, limit, offset, filter, customFilters,
```
yasharAyari: we should have one object as filters instead of having filter and customFilters

src/actions/transactions.js
slaweet: Done

src/actions/transactions.js
slaweet: Done

slaweet:
I fixed the naming as discussed. Please check again @yasharAyari

PR #2036, opened 2019-05-20
Implement "start voting" - Closes #1971

### LINKED ISSUE #1971: Implement the "Start voting" and selecting rows in delegates table
This is a subtask of #1926. It should be done in a new delegates page. There will be a follow-up ticket to replace the old delegates page.

The goal is to implement the "Start voting" and selecting rows in delegates table
https://app.zeplin.io/project/5c82526ebf0d81bce46d5255/screen/5cc03ba8866b9b265421a26a

slaweet:
### What issue have I solved?

#1971

- Implemented the design.
- Cleaned up some unnecessary code
- Created two new Toolbox components (CheckBox and Tabs)

### How has this been tested?

- Go to https://ci.lisk.io/test/lisk-hub/PR-2036/#/delegatesV2
- Click "Start voting"
- Select/unselect some delegates and watch the numbers to update.

### Review checklist


src/components/toolbox/tabs/index.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import styles from './tabs.css';
+
+const Tabs = ({
+  tabs = [], active, onClick = () => {}, className,
```
massao: IMO, it's better written with the `defaultProps`, since here you could have just the names of props.

src/components/toolbox/checkBox/checkBox.css
```diff
@@ -1,10 +1,12 @@
-@import '../app/variablesV2.css';
+@import '../../app/variablesV2.css';
 
 .checkbox {
   display: inline-block;
   width: 16px;
   height: 16px;
   position: relative;
+  z-index: 2;
```
massao: Maybe would be good to create some variables for the z-index, so it's easier, something similar to:
```css
--background-index: 0;
--base-index: 3;
--foreground-index: 6;
--overlay-index: 10;
```
So it would be easier to visualize which element is in front and which are in the back.

src/components/toolbox/tabs/index.js
slaweet: Done

src/components/toolbox/checkBox/checkBox.css
slaweet: Somehow tried to do this :-)

slaweet:
> 🐛 When sign in with an account that have already voted, the votes column show up even without clicking start voting

This is actually agreed by @reyraa, because otherwise you cannot see your votes without "Start voting".  I forgot to mention it in the PR description.

> 🐛 When there is no votes yet, and you click on the start voting, select a vote and hit cancel, the vote column is kept on the table:

Fixed.

massao:
🐛 Now the votes column doesn't show up when having voted already.
image

slaweet:
Good catch @massao, thanks. 
Fixed. Please check again.

PR #2044, opened 2019-05-20
Implement my recent transactions in dashboard module - closes #2039

### LINKED ISSUE #2039: Implement My Recent Transactions dashboard module
A step to implement the new design of the dashboard page is to adapt My Recent Transactions to the new branding language. 
There are design guides provided for the [empty](https://projects.invisionapp.com/d/main#/console/17570736/364218738/preview) and [full](https://projects.invisionapp.com/d/main#/console/17570736/364218743/preview) states of this component.

The component only shows the latest 5 transactions of the active token. meaning that the title changes according to the active token: Recent Lisk transactions or Recent Bitcoin transactions.
Alsop, in case there's a new transaction, aside from being confirmed or pending, the new transaction pushes the 5th transaction out; keeping the list length always 5.

osvaldovega:
### What issue have I solved?

-- #2039 

Implement new design of the recent transactions in the dashboard, supporting BTC token too.
Using BoxV2 for the container element and based on the active token will display the last 5 transactions for LSK or BTC based on that what is enable, if use change the token this will automatically update and show the list again.

There are 3 components now, one call TransactionTypeFigure, TransactionAddress and TransactionAmount that will be on charge on display the information based on the new design, this means that these component just need tx details and active token for display the information.

### How has this been tested?

Right now this component was implemented but not added into the actual dashboard as the dashboard component needs to be update and rest of elements too, once the other elements are done this can be enable/add it into the dashboard.

For change the active token this action needs to be dispatch in the redux devTools in the browser.
LSK:
```
{type:‘SETTINGS_UPDATE_TOKEN’, data: { token: { active: ‘LSK’, list: {BTC: true, LSK: true} }}}
```
BTC:
```
{type:‘SETTINGS_UPDATE_TOKEN’, data: { token: { active: ‘BTC’, list: {BTC: true, LSK: true} }}}
```

**NOTE:** For see the transactions based on the active token this PR needs to be merge first https://github.com/LiskHQ/lisk-hub/pull/2003/files#diff-5b74ccb6f4da52978806178cbf93433e.
This is because in this PR now all the transactions will have a new property "token" will be the flag for filter the transactions.

**NOTE:** For now the empty state show a validation when the active token is LSK or BTC, what it does is that if the active token is LSK then a Lear more button will be display and the external link is for the LSK documentation, right now we don't have the external link for the BTC, so once the external link will be available for the BTC documentation the validation can be remove

### Review checklist


src/assets/images/icons-v2/tx-2nd-passphrase2.svg
```diff
@@ -0,0 +1,10 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+    <g fill="none" fill-rule="evenodd">
+        <path fill="#FFF" fill-opacity=".1" d="M0 0h24v24H0z" opacity=".1"/>
+        <g stroke="#8A8CA2" transform="translate(4 3)">
+            <circle cx="8" cy="6" r="5.5" fill-rule="nonzero"/>
+            <rect width="15" height="11" x=".5" y="6.5" fill="#FFF" fill-rule="nonzero" rx="2"/>
+            <path stroke-linecap="square" d="M8 10.5v3"/>
+        </g>
+    </g>
+</svg>
```
massao: Do we need to keep both files, the tx-2nd-passphrase and tx-2nd-passphrase2?
Also IMO, we should use some lib to optimize the SVGs, something like the [SVGo](https://github.com/svg/svgo).

src/components/app/variablesV2.css
```diff
@@ -96,6 +96,7 @@ or "warn/action" ineastd of "red/green"
   --color-white-smoke: #f5f7fa;
   --color-burnt-sienna: #ec6868;
   --color-ufo-green: #2bd67b;
+  --color-amount-green: #00d563;
```
massao: I would rename this, since in the future it may be used to something else rather than the amount.

src/components/transactionAddress/index.js
```diff
@@ -0,0 +1,57 @@
...
+            token === tokenMap.LSK.key ? address : address.replace(regex.btcAddressTrunk, '$1...$3')
+          }
+          </span>
+        : null
+      }
+    </div>
+  );
+};
+
+export default TransactionAddress;
```
massao: If i'm not wrong, we only use index file for components for HOC

massao:
Also you could create a new route for the dashboard, similar to what we did on other pages, dashboardV2, so it would be easier to replace on the future

osvaldovega:
@massao I did the changes man, when have time check it again

osvaldovega:
ok after check with Ali, I made some changes to fix the empty state.

Issue #2068, opened 2019-05-23
The BTC send page has glitchy behaviour 


reyraa:
### Steps to reproduce

1. Change the active token to BTC
1. Input t valid amount
1. Now change the amount to can invalid value

### Actual result
There's jump cause by removing the fee/process speed component and showing the error.

### Expected result
 - The error should have a preserved box space which prevents pushing down the reset of the contents.
 - The fee component must be there from the beginning. As the designs suggest we should implement a loading state and in the case of error just write "Invalid amount" instead of the fee value.

No jumping behaviour is desired. the box should have the same height at all times.

PR #2071, opened 2019-05-24
Adjust send form behaviour - Closes 2068

massao:
### What issue have I solved?

#2068 

Adjusted feedback to have a reserved space on the scree, for now it only fixed on the send form screen, should be updated across the application on a followup ticket, can be updated on the `feedback` toolbox component.  
Updated form to always show the processing speed field, having an invalid amount and a loading status.

### How has this been tested?

Enable `btc` and set as active token, and go to the send page, the form should have reserved space for the error messages, and show the processing speed.
1. Enabled BTC -> `localStoraget.setItem('btc', true)`;
2. Set as Active -> `{type:‘SETTINGS_UPDATE_TOKEN’, data: { token: { active: ‘BTC’, list: {BTC: true, LSK: true} }}}`.

### Review checklist


osvaldovega:
Massao quick question, the code looks good but when I checked the send form, I noticed that the dynamic fee value is always loading by default, is this the correct way to show it or will be only as soon the user type anything, like if amount value length !== 0 then loading or not? should be always spinning?

everything else if fine

reyraa:
Please adjust these changes for a better look:
```
.sendWrapper
maxWidth: 530px



.formFormSection .formFieldGroup
marginTop: 0; except the first one

.formFormSection .formAmountHint
position: relative;
top: -30px
```

Efefefef:
@massao Please rerequest a review when its ready

massao:
> @massao We have no validation of minimal BTC value in the amount input.
> I think we had no validation for LSK so far, and it's a good time to implement it.

Agree, but would be good having the conditions of the validation added to the ticket.

Efefefef:
@yasharAyari Can you add conditions to the ticket or create a new one for adding validation?

massao:
@Efefefef Added the validation to not having more than 8 numbers after the floating point, and this take cares of the minimal amount too.
The error message was aligned with @reyraa

PR #2080, opened 2019-05-29
Replace old delegates page with the new one - Closes #1975

### LINKED ISSUE #1975: Replace the old delegates page with the new one
This is the final subtask of #1926. 

The goal is to replace the old delegates page with the new one

slaweet:
### What issue have I solved?

#1975

- I changed the new delegates page to be the one in the top bar, always displayed.
- Removed the components for the old delegates page
- Fixed some bugs in the new delegates section from previous PRs
- I also changed `CustomRouter` component not to render anything if network or account is loading, because the delegate list was not fetched if delegates page was visited directly with autologin. This will prevent a lot of issues that we had on pages that tried to fetch data before the network and account was set up. Not having this in the first place was IMO the reason why we have conditions like this https://github.com/LiskHQ/lisk-hub/blob/bf1ad09e5c9c06a76cb239012c7cc15b3adce0cf/src/actions/search.js#L43 

### How has this been tested?

- Go to https://ci.lisk.io/test/lisk-hub/PR-2080/ and try voting

### Review checklist


Efefefef:
🐛 Onboarding: Info text should be narrower like in designs
🐛 Onboarding: Previous and Next buttons are swapped in design. Looks weird though. @yasharAyari ?
🐛 Onboarding: Mistakes. Why texts are not taken from the file with corrected wordings? 
```Want to dig deeper? We got you covered. You can read more about Lisk’s delgates, voting mechanism and benefits in a dedicated section of Lisk’s help centre.```
🐛 Table: Load more should not be available on Votes/Not voted tabs if there is less than 100 delegates to show
🐛 Table: Wording: `No search result in given criteria.` : should be results 
🐛 Table: No spacing between number and LSK
🐛 Table: Click on a row in Vote mode doesn't work as check/uncheck. Why to have hover state then?
🐛 Table: Font of Rank looks too small comparing to checkbox size, other columns as well - not like in design
🐛 Voting submitted: `Voting submitted
You will be notified when your votes are forged.`: Why Voting not Votes. Notified how? Icon diff  design
🐛 Tooltips: why do we have only for 'Total actions' and 'Transaction fee'?
  
1. Why we have delegateOnboarding as localStorage key, not like settings object key like for the other flags?
1. Should we support launch protocol voting as well?

slaweet:
> 🐛 Onboarding: Info text should be narrower like in designs

Fixed

> 🐛 Onboarding: Previous and Next buttons are swapped in design. Looks weird though. @yasharAyari ?

The order was weird, in latest design in inVission it's as implemented https://projects.invisionapp.com/d/main#/console/17570736/365903447/inspect

> 🐛 Onboarding: Mistakes. Why texts are not taken from the file with corrected wordings?
> `Want to dig deeper? We got you covered. You can read more about Lisk’s delgates, voting mechanism and benefits in a dedicated section of Lisk’s help centre.`

Good catch. Fixed from https://docs.google.com/spreadsheets/d/1157pqdDDu6AVVJNHOv4VKlf6tRSiy0Wm753Ovbcp5as/edit#gid=0

> 🐛 Table: Load more should not be available on Votes/Not voted tabs if there is less than 100 delegates to show

"load more" behavior seems to need bigger changes, it doesn't work in some cases. TODO

> 🐛 Table: Wording: `No search result in given criteria.` : should be results

Fixed

> 🐛 Table: No spacing between number and LSK

Fixed

> 🐛 Table: Click on a row in Vote mode doesn't work as check/uncheck. Why to have hover state then?

Fixed. Whole row click works now.

> 🐛 Table: Font of Rank looks too small comparing to checkbox size, other columns as well - not like in design

Fixed

> 🐛 Voting submitted: `Voting submitted You will be notified when your votes are forged.`: Why Voting not Votes. Notified how? Icon diff design

Re: Icon diff: Dominic changed the success/error icons: https://projects.invisionapp.com/d/main#/console/17570736/365994789/preview

'Voting submitted' is approved copy, IMO fine, so I kept it.

> 🐛 Tooltips: why do we have only for 'Total actions' and 'Transaction fee'?

Fixed. Added tooltip for "Productivity" and "Vote weight"

> 1. Why we have delegateOnboarding as localStorage key, not like settings object key like for the other flags?

I guess it's because then the onboarding component needs HOC to connect to store and toolbox components shouldn't need HOC.

> 2. Should we support launch protocol voting as well?

Launch protocol should work.

Efefefef:
🏅

PR #2085, opened 2019-06-03
Clean code of delegates and voting components - Closes #2081

### LINKED ISSUE #2081: Clean code of delegates and voting components
This is a follow-up to https://github.com/LiskHQ/lisk-hub/pull/2080

- `votingV2` should be renamed to `voting`, `src/components/delegatesV2` -> `src/components/delegates` etc.
- `src/components/votingListViewV2/` should be moved inside src/components/delegates
- `goToNextStep` property of `votePlaced` action, should be renamed to `callback`
- There should be a new `VoteList` component to extract duplicated code from `votingSummary` and `voteUrlProcessor`
- any other code quality issue encountered when editing voting and delegates code.

slaweet:
### What issue have I solved?

#2081

- `votingV2` was renamed to `voting`
- `src/components/delegatesV2` -> `src/components/delegates`
- Change votingResult component to generic transactionResult so that it can be reused in other transaction types
- Created a new `VoteList` component to extract duplicated code from `votingSummary` and `voteUrlProcessor`
- `goToNextStep` property of `votePlaced` action, renamed to `callback` because the action should be generic and not assume it's called from a certain context.

#### TODO
I created a follow-up ticket https://github.com/LiskHQ/lisk-hub/issues/2088 to take care of rest.

Also this ticket https://github.com/LiskHQ/lisk-hub/issues/2089 depends on the code from this pr.

### How has this been tested?

Check voting and delegates e2e tests pass.

### Review checklist


src/components/voting/voteList.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import styles from './voteList.css';
+
+const VoteList = ({
+  list, title, className, votes,
+}) => (
+  <React.Fragment>
+  {list.length > 0 ?
```
osvaldovega: you can do in here  list.length don't need the > 0

src/components/transactionResult/index.js
```diff
@@ -2,15 +2,15 @@ import React from 'react';
 import { PrimaryButtonV2, TertiaryButtonV2 } from '../toolbox/buttons/button';
 import Illustration from '../toolbox/illustration';
 
-import styles from './votingResult.css';
+import styles from './transactionResult.css';
 
 const getErrorReportMailto = (error) => {
```
osvaldovega: I think we are using this in other files too, so maybe we can have this function in a utils file where we can use it and just pass the error message

src/actions/voting.js
```diff
@@ -90,7 +90,7 @@ const handleVoteError = ({ error }) => {
  * cleans the pending state
  */
 export const votePlaced = ({
-  passphrase, account, votes, secondPassphrase, goToNextStep,
+  passphrase, account, votes, secondPassphrase, callback,
```
osvaldovega: I know this was already like this, but could be a way to split this function?, right now the code looks good but it has so many lines of code now, that start to looks bad

slaweet: I agree this needs a clean-up and it will be done in https://github.com/LiskHQ/lisk-hub/pull/2099

src/components/voting/voteList.js
slaweet: https://github.com/LiskHQ/lisk-hub/pull/2085/commits/920d8064b3c6e76670313b6d7b8e8100bf754956

src/components/transactionResult/index.js
slaweet: Done in https://github.com/LiskHQ/lisk-hub/pull/2085/commits/708ab8212f457057074fccc3d59c00d2506616e4

Issue #2043, opened 2019-05-20
Implement the new design of Bookmarks component in dashboard


reyraa:
After renamed the Followed accounts to Bookmarks, we can update the design of this component to comply with the new branding language. The design are already in InVision. There are separated design guides for [full](https://projects.invisionapp.com/d/main#/console/17570736/364218743/preview) and [empty](https://projects.invisionapp.com/d/main#/console/17570736/364218738/preview) screens.

The design includes tabs to filter the bookmarks based on the token. Changing the tab must not affect the active token. it only filters the bookmarks of this component.

Both _View all_ and _Search for accounts_ buttons navigate to the bookmarks page.

Please make sure the font details, sizing and colors are imported from a commonly used filed and not generated on a component by component basis.

PR #2087, opened 2019-06-03
Implement new design of bookmarks component in dashboard - closes 2043

osvaldovega:
### What issue have I solved?

-- #2043 

This PR contains the new design for the Bookmarks component in the dashboard page.
The new behavior is based on the active token, so if the user has selected LSK as active token for the app, thee bookmarks will show the last 5 bookmarks, if the user switch to BTC, then will show the BTC bookmarks.

In case of no saved bookmarks and empty state will be display.

### How has this been tested?

For see this dashboard with the new bookmarks design, please log in into the app and go to dashboard v2, ilke this.

1-. <hub url>/#/dashboardV2
2. Then you will see the list of bookmarks in case of have it.
3. If there is no bookmarks, then save some in the send component or wallet component and go back to dashboardV2
4. For switch to BTC a action needs to be dispatch in the redux development tool in the browser.
For LSK
```
{ type: 'SETTINGS_UPDATED', data: { token: { active: 'LSK' } } }
```
For BTC
```
{ type: 'SETTINGS_UPDATED', data: { token: { active: 'BTC' } } }
```

**NOTE:** The bookmark main page it is not implemented so the view all button at this point it is not redirecting to the correct page until this be implemented, once this be implemented the view all button will redirect to this page, same as the empty state.

```
<div>
  <Link to={'#'}>
    <SecondaryButtonV2>{t('Search Accounts')}</SecondaryButtonV2>
  </Link>
</div>
```
and 
```
<Link to={'#'}>
  <SecondaryButtonV2>{t('View All')}</SecondaryButtonV2>
</Link>
```

### Review checklist


src/components/bookmarksList/bookmarksList.js
```diff
@@ -0,0 +1,96 @@
...
+  onBookmarkSelected(address) {
+    const { history } = this.props;
+    history.push(`${routes.explorer.path}/accounts/${address}`);
+  }
+
+  getBookmarkListBasedOnSelectedToken() {
+    const { bookmarks, token } = this.props;
+
+    const actualBookmarks = { ...bookmarks };
+    return actualBookmarks[token.active].length > 5
```
massao: Don't need to check the length of the bookmarks here, you can just `return bookmarks[token.active].slice(0, 5);`

src/components/bookmarksList/bookmarksList.js
```diff
@@ -0,0 +1,96 @@
...
+                <span>{bookmark.title}</span>
+                <span>{this.displayAddressBasedOnSelectedToken(bookmark.address)}</span>
+              </div>
+            </div>)
+          : <EmptyState>
+              <img src={svg.bookmarksIconEmptyState} />
+              <h1>{t('No Bookmarks added yet')}</h1>
+              <p>{t('Start adding some addresses to bookmarks, to keep track of them.')}</p>
+              <div>
+                <Link to={'#'}>
```
massao: I would add a `// TODO` comment to be easier to find this on the code.

src/components/bookmarksList/bookmarksList.js
```diff
@@ -0,0 +1,96 @@
...
+                  <SecondaryButtonV2>{t('Search Accounts')}</SecondaryButtonV2>
+                </Link>
+              </div>
+          </EmptyState>
+        }
+        {
+          selectedBookmarks.length
+          ? <div className={styles.footer}>
+              <Link to={'#'}>
```
massao: Would add a `// TODO` comment here too

src/components/bookmarksList/bookmarksList.js
```diff
@@ -0,0 +1,96 @@
...
+        </div>
+      </Box>
+    );
+  }
+}
+
+BookmarksList.propTypes = {
+  bookmarks: PropTypes.object.isRequired,
+  history: PropTypes.object.isRequired,
+};
```
massao: Since the props come from the redux store, I'm not sure if it's needed to put them at propTypes, since it wouldn't be possible to not pass them to the component.

src/components/dashboardV2/recentTransactions/recentTransactions.js
```diff
@@ -45,14 +46,22 @@ class RecentTransactions extends Component {
...
-            btnText={t('Learn more')}
-            btnExternalUrl={links.outgoingTransactions}
-          />
+        : <EmptyState>
+            <img src={svg.icon_empty_recent_transactions} />
+            <h1>{t('No Transactions Yet')}</h1>
+            <p>{t('A great way to start is to top up your account with some {{value}} tokens.', { value: activeToken.key })}</p>
+            <div>
+            {
+              activeToken.key === 'LSK' // TODO this validation should be remove once we have the external link for BTC
```
massao: Should be using tokenMap constant

src/components/bookmarksList/bookmarksList.js
```diff
@@ -0,0 +1,96 @@
...
+          <h2>Bookmarks</h2>
+        </header>
+        <div className={`${styles.bookmarkList} bookmark-list-container`}>
+        {
+          selectedBookmarks.length
+          ? selectedBookmarks.map(bookmark =>
+            <div
+              key={bookmark.address}
+              className={`${styles.row} bookmark-list-row`}
+              onClick={() => this.onBookmarkSelected(bookmark.address)}>
```
massao: Isn't it possible to instead of having the `onClick`, to do a history.push, wrap this inside a `<Link>`?

PR #2101, opened 2019-06-04
Fix account to update in voting - Closes #2096

### LINKED ISSUE #2096: Voting: Balance change after login is not taken into account for Vote tx fee calculation
### Steps to reproduce

1. Login with fresh/empty acc to devnet
1. Send some money to the account
1. Vote for somebody

### Actual result

Not enough LSK error

### Expected result

I have the money so I should be able to send/vote

slaweet:
### What issue have I solved?

#2096

The problem was that voting was not getting account based on the active token. So I used a util to make sure that happens.

### How has this been tested?

Follow steps in #2096

### Review checklist

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -185,7 +185,11 @@ export const accountDataUpdated = ({
 }) =>
   (dispatch, getState) => {
     const networkConfig = getState().network;
-    getAccount({ networkConfig, address: account.address }).then((result) => {
+    getAccount({
+      networkConfig,
+      address: account.address,
+      publicKey: account.publicKey,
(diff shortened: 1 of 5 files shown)


slaweet:
I can't believe it, but the problem with hardware wallet was caused by an eslint rule automatically fixed with `eslint --fix`. More details in https://github.com/LiskHQ/lisk-hub/pull/2101/commits/c0eab578d03bd4ff3d31fc91f691a0f400f186c2

Should be good now.

Efefefef:
Looks like we have to wait before restarting the build
image

PR #2102, opened 2019-06-04
Hide "Edit voting" button for hardware wallets - Closes #2097

### LINKED ISSUE #2097: Voting: HW wallet: Edit voting link is enabled
### Steps to reproduce

1. Login with HW wallet
1. Check delegates and proceed to Confirmation page

### Actual result

Edit voting link is present

### Expected result

It should be disabled, as we should Confirm/Cancel tx on HW wallet.
Clicking Edit is not canceling tx on HW wallet but create a mess

image

slaweet:
### What issue have I solved?

#2097

I changed the condition to hide not only the primary but also the secondary button in the case of a hardware wallet.

### How has this been tested?

Try to vote with a hardware wallet.

### Review checklist

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -58,7 +58,7 @@
   "Auto-logout": "Auto-logout",
   "Back": "Back",
   "Back to Delegates": "Back to Delegates",
-  "Back to Voting": "Back to Voting",
+  "Back to Voting Table": "Back to Voting Table",
   "Back to wallet": "Back to wallet",
   "Balance": "Balance",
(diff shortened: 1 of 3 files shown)


slaweet:
@Efefefef good point, I'll change the button to "Back to Voting". 

I would take it even one step further and do the same also for a non-hwWallet account, to keep things consistent as in that case an error also likely means I need to edit the voting. @reyraa is it ok?

slaweet:
Label updated @yasharAyari please approve the PR.

Efefefef:
👍 Approved

PR #2105, opened 2019-06-04
Implement new design of sign message - Closes #2052

### LINKED ISSUE #2052: Implement the sign message page
This feature existed in Nano, due some reasons we decided to remove it. There's a[ new design](https://projects.invisionapp.com/d/main#/console/17570736/364220681/preview) and wee can add it back.

massao:
### What issue have I solved?

#2052 

- Cleaned code from sign message component;
- Updated to new design;
- Adjusted some mixins to have correct values;
- Updated `textarea` component to have size props as the `input` also have.

### How has this been tested?

1. Go to `/sign-message` and input any message;
    - Clicking on `Go back` uses `history.goBack`;
2. The next screen should show the signed message;

It's possible to use the query params to prefill a message:
`/sign-message?message=Any message here`

### Review checklist


src/components/app/variablesV2.css
```diff
@@ -101,6 +101,8 @@ or "warn/action" ineastd of "red/green"
   /*************************
       Other Colors
   *************************/
+  --color-console-background: #08223c;
+  --color-console-text: #9bf436;
```
osvaldovega: Probably this colors codes will be only use eon that screen but if not maybe you can rename the variables to use the color name so that will be more generic.
You can use a tool like this for get the names
http://chir.ag/projects/name-that-color/#9BF436

massao:
> everything looks good, just one minor change.
> Question, I see in the last screen there is no Go Back button, however if I want to try something else how can I go to that page without use the URL again.
> Could be possible to add the Go Back button to that screen or it is intended on that way?

Added go to dashboard link on the last screen, since if the user wants to sign another message the user should use the url again.

PR #2110, opened 2019-06-05
New designs for modals for timeout and sign out - closes #2066

### LINKED ISSUE #2066: Create the modals for timeout and signet
We have new designs for two modals:
 - 1 minute before the time is over, [a popup](https://projects.invisionapp.com/d/main#/console/17570736/364220520/preview) opens, requiring an action before the times up.
 - When the time is up, [a popup](https://projects.invisionapp.com/d/main#/console/17570736/364220519/preview) opens, requiring the user to sign in or continue as guest.

The users don't see the timer all the time. they'll only be notified one minute before the time is over, letting them reset the timer and once when the time is over allowing them to decide the next action.

This change means that we don't need the footer anymore, considering we've already moved the network identifier to the [header](https://github.com/LiskHQ/lisk-hub/issues/2053).

osvaldovega:
### What issue have I solved?

-- #2066

New design for the modal when is running out of time when the auto-logout option is enabled.
Update css file to match new design

### Review checklist

### DIFF
--- a/src/components/dialog/dialog.css
+++ b/src/components/dialog/dialog.css
@@ -1,15 +1,27 @@
+@import '../app/variablesV2.css';
+@import '../app/mixins.css';
+
 .dialog {
+  border-radius: 6px;
+
+  & .title {
+    @mixin headingNormalIntermediate;
+
+    color: var(--color-maastricht-blue);
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
+    margin: 25px 0;
+  }
+
   & .x-button {
     right: -20px;
 
     & span {
       color: rgba(255, 255, 255, 0.87);
     }
-  }
-
-  & p {
-    color: rgba(0, 0, 0, 0.87);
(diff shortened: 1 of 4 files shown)


src/components/dialog/dialog.css
```diff
@@ -1,4 +1,21 @@
+@import '../app/variablesV2.css';
+@import '../app/mixins.css';
+
 .dialog {
+  border-radius: 6px;
+
+  & .title {
+    @mixin headingNormalIntermediate;
```
massao: I just checked with Domenico, and we shouldn't need this mixin, since it's not on the styleguide, we should or use the large or the normal, just check with him in each case.

src/components/dialog/dialog.css
```diff
@@ -1,4 +1,21 @@
+@import '../app/variablesV2.css';
+@import '../app/mixins.css';
+
 .dialog {
+  border-radius: 6px;
```
massao: Also check with Domenico, if we could use the same border as in the other parts of the application.

PR #2142, opened 2019-06-17
Implement the new design of the Register second passphrase page - Closes #2136

### LINKED ISSUE #2136: Implement the new design of the Register second passphrase page
This ticket aims to implement the new design of the second passphrase page.
The latest design to be provided, here's some older version https://projects.invisionapp.com/d/main#/console/17570736/366378003/comments

slaweet:
### What issue have I solved?

#2136

- [x] Refactored  first passphrase backup step to have to common behaviour in [passphraseBackup](https://github.com/LiskHQ/lisk-hub/tree/70a4e4a0cb647fa8864dd4d05a30c8c9ad4ba422/src/components/passphraseBackup)
- [x] Implemented the backup passphrase step (using `PassphraseBackup` component) https://projects.invisionapp.com/d/main#/console/17570736/366378003/inspect
- [x] Implemented summary step  (using `TransactionSummary` Component) https://projects.invisionapp.com/d/main#/console/17570736/366378003/inspect Note that it is without box, as all other transaction types don't have the box yet and should be updated to use the BoxV2 in the summary step all at once in one ticket.
- [x] Implemented transaction result step
- [x] Implemented error toast insufficient balance (same as in voting)
- [x] Updated e2e tests
- [x] Deleted v1 1st and 2nd passphrase components as no-longer-needed
- [x] Deleted secondPassphrase redux as no-longer-needed

### How has this been tested?

- Go to settings 
- try to register 2nd passphrase
- follow the steps

### Review checklist


src/components/passphraseBackup/index.js
```diff
@@ -0,0 +1,136 @@
...
+                    {t('You can print your passphrase to store in a safe place. ')}
+                    {t('It is highly recommended to delete the PDF file and remove it from your Trash Folder too after printing it.')}
+                  </p>
+                </Tooltip>
+              </h2>
+              <div style={{ display: 'none' }} ref={this.setCanvasRef}>
+                <QRCode value={account.passphrase} />
+              </div>
+              <div className={styles.downloadLisk}>
+                <img src={svgIcons.fileOutline} />
```
osvaldovega: If I am not wrong I think we can use here the component that you use for manage the svgs

src/components/registerV2/backupPassphrase.js
```diff
@@ -1,65 +1,16 @@
-import React from 'react';
-import QRCode from 'qrcode.react';
-import { CopyToClipboard } from 'react-copy-to-clipboard';
 import { translate } from 'react-i18next';
-import moment from 'moment';
+import React from 'react';
 import grid from 'flexboxgrid/dist/flexboxgrid.css';
+
```
osvaldovega: I dont know if this new line was intended or not, but if not maybe you can fix it

src/components/registerV2/backupPassphrase.js
```diff
@@ -1,65 +1,16 @@
...
-import links from '../../constants/externalLinks';
-import Tooltip from '../toolbox/tooltip/tooltip';
-import { InputV2 } from '../toolbox/inputsV2';
-import renderPaperwallet from '../../utils/paperwallet';
-import svgIcons from '../../utils/svgIcons';
+import PassphraseBackup from '../passphraseBackup';
 import registerStyles from './registerV2.css';
-import styles from './backupPassphrase.css';
 
 class BackupPassphrase extends React.Component {
```
osvaldovega: This component now looks like just a presentational component so maybe instead of a class you can use a stateless component function and then just make the destructuring in the parameters and not use the class.
I think looks better what do you think?

src/components/secondPassphrase/secondPassphrase.js
```diff
@@ -1,19 +1,31 @@
...
 import routes from '../../constants/routes';
+import styles from './secondPassphrase.css';
 
 class SecondPassphrase extends React.Component {
+  constructor() {
+    super();
+
+    // istanbul ignore next
+    const crypotObj = window.crypto || window.msCrypto;
+    this.secondPassphrase = generatePassphrase({
```
osvaldovega: I think this function will looks better in a utils file and then just called in here

src/actions/account.js
```diff
@@ -75,27 +74,33 @@ export const passphraseUsed = data => ({
 /**
  *
  */
-export const secondPassphraseRegistered = ({ secondPassphrase, account, passphrase }) =>
+export const secondPassphraseRegistered = ({
+  secondPassphrase, account, passphrase, callback,
+}) =>
   (dispatch, getState) => {
+    if (account.balance < Fees.setSecondPassphrase) {
```
osvaldovega: I think this validation should be done outside this thunk and be called before this function.
This function should be only on charge of register the second pass, but I think will be better to avoid that one function manage more than one responsibility. What do you think?

src/components/secondPassphrase/secondPassphrase.js
slaweet: Good catch. Fixed.

src/components/registerV2/backupPassphrase.js
slaweet: It was inserted by [import-js](https://github.com/Galooshi/import-js) which I use since a few weeks to automatically generate imports for me. I can recommend the tool for you and @massao too, it has plugins for various editors.

 I changed config not to insert the empty line between internal and external imports https://github.com/LiskHQ/lisk-hub/pull/2142/commits/7fc14b1e61daf3616c3450d032c12e4ff68bd60e

src/actions/account.js
slaweet: I checked with Dominic and implemented checking the balance already on settings page.

slaweet:
Renamed https://github.com/LiskHQ/lisk-hub/pull/2142/commits/c4848aaa449cf87f6045ad8c226e3d392e647385 @yasharAyari

slaweet:
Please check again @Efefefef 
All issues should be fixed except:
- Button widths
Screenshot 2019-06-21 at 10 59 11
- Box in summary page - this one will be updated in all transaction types together to keep consistency, send and vote summary still have the design without box.

slaweet:
Fixed all except this one:

> 🐛 Wrong size
> image

Invissison says 15px and it is 15px. But I noticed the color wasn't correct, so I fixed that.

Efefefef:
Hmm height is 18 in Invision, and 20 in app

Looks like same problem here plus text should be vertically aligned to center 
image
Same here
image

slaweet:
- Fixed vertical alignment
- the `OR` line is `rgba(0,0,0,0.15);` in design. Now I used it, but we don't have such color in the styleguide https://projects.invisionapp.com/d/main#/console/17570736/365298851/inspect so should be updated in design.

Issue #2127, opened 2019-06-11
Refactor transactions actions not to store pending BTC transactions 


slaweet:
## Current behaviour
When a new transaction is emitted to the netwrork, it is added to redux  `state.transactions.pending` no matter what the token is. 

## Expected behaviour
When a new transaction is emitted to the netwrork, it is added to `state.transactions.pending` if the token is LSK, but not if the token is BTC. Because the BTC API response contains both pending and confirmed transactions.

PR #2150, opened 2019-06-18
Enable store and send BTC feature - Closes #1893

### LINKED ISSUE #1893: Enable BTC wallet to end users
This is the final subtask of #1831.

The goal is to enable the functionality implemented in the previous subtasks to be visible by default, not just by a local storage flag.

slaweet:
### What issue have I solved?

#1893

- [x] Removed all the `localStorage.getItem('btc')` conditions
- [x] Updated unit tests to pass
- [x] Fix some styles on the transaction page, especially in smaller breapoints
- [x] Fix `onClick` of `CopyToClibpboard`
- [x] Fix tabs styles for when first tabs is also last
- [x] Transaction list in wallet gets sometimes LSK transactions even if BTC is active
- [x] Prevent an error on wallet after sending BTC
- [x] Adjust "Empty account" wallet module to adjust to active token.
- [x] Update `src/utils/api/btc/transactions.js` to work also with BTC mainnet

### How has this been tested?

Try to
- Send BTC
- Switch token
- Enable/disable BTC
- View BTC transactions in wallet and dashboard

### Review checklist


src/actions/account.js
```diff
@@ -229,27 +229,22 @@ export const login = ({ passphrase, publicKey, hwInfo }) => async (dispatch, get
   }).then(async (accountData) => {
     const expireTime = (passphrase && getState().settings.autoLog) ?
       Date.now() + accountConfig.lockDuration : 0;
+    let btcAccountData;
+    if (getState().settings.token.list.BTC) {
```
massao: Is it possible to merge this with with the other `getState` call on this function, something like `const { network: networkConfig, settings } = getState()`, so there is no need to call it again?

src/actions/account.js
```diff
@@ -229,27 +229,22 @@ export const login = ({ passphrase, publicKey, hwInfo }) => async (dispatch, get
...
+      await getAccount({
+        token: tokenMap.BTC.key, networkConfig, passphrase,
+      }).then((response) => { btcAccountData = response; });
+    }
```
massao: IMO, this would look better, as there would be no mutation:
```js
const btcAccountData = getState().settings.token.list.BTC && await getAccount(...);
```
And in case the token is disabled, the value would just be false.

slaweet:
Thank you for the feedback @massao, good ideas. Fixes pushed, please check again.

Efefefef:
@yasharAyari Should we show Network Indicator when in BTC mode?
image

🐛 Balance chart?
http://localhost:8080/#/explorer/accounts/n45uoyzDvep8cwgkfxq3H3te1ujWyu1kkB
image

🐛 When more that 1 tx is pending, not all of them appears on tx list which is very very misleading

🐛 I think the balance value is also glitchy, but because of the above one it's hard to test it

slaweet:
- Fixed network indicator to say "Testnet" for BTC custom node
- Fixed balance chart on other account page
- There is some delay before pending txs apear, @yasharAyari says it's normal and ok for BTC (also on other BTC apps)

Efefefef:
🐛 Login with HW wallet doesn't work

slaweet:
@Efefefef ready to check:
- Disable account initialization for BTC
- Disable BTC on HW wallet Login 
- Redirect from non-BTC pages to Dashboard on token switch

@yasharAyari what should happen if btc-test.lisk.io is unavailable on login? Login and disable BTC?

slaweet:
Jenkins green now. We tested my send with hw wallet fix with @yasharAyari 
Anything else @Efefefef ?

Efefefef:
@slaweet Yes, I'm waiting for the answers from @yasharAyari for questions above

slaweet:
> @slaweet For me it's still unclear. Why when I make BTC transfer then go to wallet and cant see transaction in there for a whole minute. Cant we show it regardless of data fetched?

@Efefefef, We can show it. I also think it's better to show it. It should be as easy as removing this `if` https://github.com/LiskHQ/lisk-hub/blob/ff8e5e8a1a84b7e7aecde71fee1301d8bb2feff8/src/actions/transactions.js#L370 but @yasharAyari explicitly asked me to create a ticket https://github.com/LiskHQ/lisk-hub/issues/2127 to put that condition in place.

Both @yasharAyari and @reyraa yesterday confirmed that not handling pending transactions locally is expected behavior for BTC.

yasharAyari:
@Efefefef about the first question, we can't keep the PR open because of a bug in the BTC-bridge. As soon as they fix the issue, we will patch the lisk-hub to work with the changes.
for the second problem please open an issue because it is not related to the context of this PR.

Efefefef:
@yasharAyari Question is if we can release BTC feature with this bug, not about keeping PR open. How we are going to track the fix?
For the second one I created a ticket https://github.com/LiskHQ/lisk-hub/issues/2197

Efefefef:
Ok then, the last question would be about Search functionality in BTC mode.
I could not find a single word about it in the epic. Should we disable it?

slaweet:
Checked with @yasharAyari:
- Search icon in top bar should be hidden for BTC (commit coming soon)

- the bug in bitcoin bridge happens for lisk-mobile, so it’s ok to release

And I add that, the user needs to have over 30 transactions for the bug to happen, so it's not that likely to happen.

yasharAyari:
@Efefefef `BTC-bridge` is stabled enough to be used in production.it has used in the `lisk-mobile` already.

Efefefef:
@slaweet Please make the LSK accounts, transactions, and the rest unavailable and redirected under active BTC

PR #2179, opened 2019-06-21
Adjust paperwallet design - Closes #2161

### LINKED ISSUE #2161: Align the design of paperwallet
Invision: 
https://projects.invisionapp.com/d/main#/console/17570736/369317665/preview

massao:
### What issue have I solved?

#2161

- Updated font sizes and spacing in paper wallet;

**OBS:** Passphrase doesn't look exactly like the design:
- The font on the design is not used anywhere else on the application, so it was decided to keep the Gilroy;
- It's not possible with the library we are using to justify the text in a way to use all the available space;

### How has this been tested?

- Go to `/register` select any avatar and continue;
- Click on the `Download` button for the paperwallet;
- Design should be similar to the design;

### Review checklist

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -218,7 +218,7 @@
   "High": "High",
   "How do you like the hub?": "How do you like the hub?",
   "How is Lisk transparent?": "How is Lisk transparent?",
-  "How we recommend to store it": "How we recommend to store it",
+  "How we recommend to store it.": "How we recommend to store it.",
   "How would you call it?": "How would you call it?",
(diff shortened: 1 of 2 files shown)


massao:
> 🐛 "Paper" and "Encrypted wallet" should be in bold

@Efefefef that's not possible with the implementation we have at the moment, since all the tests are positioned in an absolute way.

massao:
> 🐛 "Paper" and "Encrypted wallet" should be in bold

@Efefefef It's not possible with the implementation we have right now, we would need to update how the paper wallet is being rendered, to use an HTML as base, and we would have some others limitations with that.

PR #2186, opened 2019-06-24
Remove unnecessary functions from delegate actions and reducers - closes #2185

### LINKED ISSUE #2185: Remove unnecessary functions from delegate actions and reducers

The delegate actions file right now is has some action creators and thunk that are being using for check if an desire nickname for a new delegate is already registered or not, when could be a simple call with the typed nickname and then validate if is valid or not.

### Suggested solution

Remove this function from the redux store and just make the call from any function in the component and wait for the response.
So no thunk, action creator and reducer needs to be use for this.

osvaldovega:
### What issue have I solved?

-- #2185 

Update delegate actions files removing unnecessary function.
Update delegate reducer removing unnecessary switch cases.
update delegate registration logic for validate typed nickname
update transaction creation function in summary component
update transaction broadcast function in status component

### How has this been tested?

Run the app
Go to delegate registration
type a valid and not used nickname
Then type an used nickname like genesis_1
check that the code is working properly making the corresponding validations and set the correct error messages

### Review checklist


osvaldovega:
**NOTE:** All the functionality is working as should be the only thing to take in count is that there are 2 unit test files that need to improve the coverage, for this there are a follow up ticket #2199

osvaldovega:
I fixed @Efefefef I missed on level for proper validation, is ready now

Efefefef:
Now it is same as before. You fail the transaction, then retry. Transaction is submitted but you get the failed message :(

osvaldovega:
Sorry @Efefefef I missed a little detail in the reducer, now should be working

PR #2187, opened 2019-06-25
Implement bookmarks page - Closes #2111

slaweet:
### What issue have I solved?

#2111

### How have I implemented/fixed it?

- [x] Implement the bookmarks page with a list of bookmarks
- [x] Implement filtering bookmarks
- [x] Implement editing bookmark
- [x] Implement removing bookmark
- [x] Implement empty state
- [x] Implement empty state when filtering by title
- [x] Enable button on dashboard bookmarks module that leads to the bookmarks page
- [x] Enable button on dashboard bookmarks module empty state that leads to the add bookmark page
- [x] Enable button on add bookmarks page that leads to the bookmarks page

The "Undo" functionality of removing bookmarks was decided in the last design feedback meeting to be implemented sometime later.

### How has this been tested?

- go to http://localhost:8080/#/bookmarks and try to use it :-) 

### Review checklist
- The PR follows our [Test guide](/LiskHQ/lisk-hub/blob/development/docs/TEST_GUIDE.md)
- The PR follows our [CSS guide](/LiskHQ/lisk-hub/blob/development/docs/CSS_GUIDE.md)

PR #2192, opened 2019-06-26
Implement add to bookmark page - Closes #2056

### LINKED ISSUE #2056: Implement the the add to bookmarks page
This page enables users to add new bookmarks. It's a subpage os the bookmarks page. The address validator is based on the active token. so the user can not add BTC bookmarks when the active token is LSK.

massao:
### What issue have I solved?

#2056

- Created new route `/bookmarks/add-bookmark`;
- Created form for adding a new bookmark;
- Validation of address is token based;
- Avatar is only generated for valid addresses, and is token based;
- Should show error if address is invalid or already bookmarked;
- Label only have validation of length, must be <= 20 character;
- Was decided with the design team, that we don't need the spinner and checkmark on the inputs on this page;
- If token is LSK and address belongs to a delegate, username is fetched and label field gets updated and becomes readonly;
- Switching token should clear the fields;

Obs: After successfully adding a new bookmark, it gets redirected to`/dashboard` page, due to the `/bookmarks` page being implemented in another PR: #2187 

### How has this been tested?

- Go to `/bookmarks/add-bookmark`;
- Fill the inputs with any desired data, it should have  the validations on both fields;
  - If address belongs to a delegate, label should be auto filled and readonly;
- After adding the bookmark, go to `/dashboard` to check  that the bookmark was added, since we don't have the bookmarks page yet;

### Review checklist


src/components/bookmarks/addBookmark/addBookmark.css
```diff
@@ -0,0 +1,87 @@
...
+    & .input[name="address"] {
+      padding-left: 48px;
+    }
+
+    & .input:focus ~ .feedback {
+      opacity: 1 !important;
+    }
+  }
+
+  & .feedback {
```
slaweet: Can at least some of these styles be inside the `Feedback` component?

src/components/bookmarks/addBookmark/addBookmark.css
```diff
@@ -0,0 +1,87 @@
+@import '../../app/variablesV2.css';
+@import '../../app/mixins.css';
+
+.wrapper {
+  display: flex;
+  justify-content: center;
+  width: 100%;
+
+  & .content {
+    width: 720px;
```
slaweet: Would be nice to have a global variable for this

src/components/bookmarks/addBookmark/addBookmark.css
massao: Maybe it's better to create a ticket to update places using the feedback component, since it should be consistent across all the application, and I'm worried of adding the rules to the component and breaking the component  in another place.

slaweet: ok

src/components/bookmarks/addBookmark/addBookmark.js
```diff
@@ -0,0 +1,280 @@
...
+                    <InputV2
+                      error={fields[field.name].error}
+                      className={styles.input}
+                      value={fields[field.name].value}
+                      onChange={this.onInputChange[field.name]}
+                      name={field.name}
+                      placeholder={field.placeholder}
+                      readOnly={fields[field.name].readonly}
+                      size={'l'}
+                      autoComplete={'off'}
```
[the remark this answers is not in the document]
massao: I updated this one with double quotes, due to: https://github.com/airbnb/javascript/tree/master/react#quotes

massao:
> * 🐛 There is too much margin around the inputs and button compared to the design https://projects.invisionapp.com/d/main#/console/17570736/368615413/preview
Updated to have less space between elements, others spacing are the same as in the design
> * 🐛 The "Bookmark added" step is missing https://projects.invisionapp.com/d/main#/console/17570736/368615410/preview
It was decided  with @reyraa and @yasharAyari  that since it doesn't rely on a API call, it should just redirect to the bookmarks list.

src/components/bookmarks/addBookmark/addBookmark.js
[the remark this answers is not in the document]
slaweet: Ah.. you're right 👍

Efefefef:
Can you please rebase on development so i can switch tokens?
🐛 You dont clean the readonly input field on change of the address (when there was delegate before)

Efefefef:
Approved, as the last comment is not actual

PR #2216, opened 2019-07-03
Fix the tooltip mouseleave event bug - Closes #2168

### LINKED ISSUE #2168: Tooltips are not gone immediately as I move mouse away
### Steps to reproduce

1. Hover on tooltip
1. Move mouse away from tooltip

### Actual result

Tooltip stays for some time

### Expected result

Tooltip should be gone immediately

yasharAyari:
### What issue have I solved?
It is described in #2168.

### Review checklist

### DIFF
--- a/src/components/toolbox/tooltip/tooltip.js
+++ b/src/components/toolbox/tooltip/tooltip.js
@@ -39,9 +39,9 @@ class Tooltip extends React.Component {
 
   handleMouseLeave() {
     this.setState({
-      timeoutObj: !this.state.clicked && this.state.showTooltip
-        && setTimeout(this.handleClose, 1000),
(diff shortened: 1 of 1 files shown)


src/components/toolbox/tooltip/tooltip.js
```diff
@@ -39,9 +39,9 @@ class Tooltip extends React.Component {
 
   handleMouseLeave() {
     this.setState({
-      timeoutObj: !this.state.clicked && this.state.showTooltip
-        && setTimeout(this.handleClose, 1000),
+      timeoutObj: !this.state.clicked && this.state.showTooltip,
```
osvaldovega: The solution work properly, however now that you remove the setTimeout then you need to remove the clearTimeout too, as we don't need it any more in the file

PR #2225, opened 2019-07-10
Fix leftovers in bookmarks - Closes #2206

### LINKED ISSUE #2206: Leftovers problems for bookmarks
🐛 'Filter by name' is wrong because filtering works both for name and address
🐛 While editing label you can save with empty string
🐛 Bookmarks header is jumping when go to Add a new bookmark

slaweet:
### What issue have I solved?

#2206

- [x] Prevent Bookmarks header from jumping when go to Add a new bookmark. By creating a toolbox `PageHeader` component and use it in both pages.
- [x] 'Filter by name' is wrong because filtering works both for name and address - restrict to name only
- [x] Prevent while editing label to save with empty string by disabling the `Save changes` button

### How has this been tested?

- Go from bookmarks page to Add a new bookmark page
- Try filtering by address
- Try saving with empty title

### Review checklist

### DIFF
--- a/src/components/toolbox/pageHeader/index.js
+++ b/src/components/toolbox/pageHeader/index.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import styles from './pageHeader.css';
+
+const PageHeader = ({
+  title, subtitle, children,
+}) => (
+  <header className={styles.wrapper}>
+    <span>
+      <h1>{title}</h1>
+      <div className={styles.subtitle}>{subtitle}</div>
+    </span>
+    <span>
+      {children}
+    </span>
+  </header>
+);
+
+export default PageHeader;
--- a/src/components/toolbox/pageHeader/index.js
+++ b/src/components/toolbox/pageHeader/index.js
@@ -5,13 +5,13 @@ const PageHeader = ({
   title, subtitle, children,
 }) => (
   <header className={styles.wrapper}>
-    <span>
+    <div>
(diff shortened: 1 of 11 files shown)


src/components/toolbox/pageHeader/index.js
```diff
@@ -0,0 +1,18 @@
...
+import styles from './pageHeader.css';
+
+const PageHeader = ({
+  title, subtitle, children,
+}) => (
+  <header className={styles.wrapper}>
+    <span>
+      <h1>{title}</h1>
+      <div className={styles.subtitle}>{subtitle}</div>
+    </span>
```
osvaldovega: hey Vit I was checking this and I know that works, and it is not a big deal, will not be better to wrap the title and subtitle in the div tag and then the subtitle to be a span?

slaweet: What I was trying to express is that there are multiple `span`s in a row, but `div` has its own row in its container.  But I don't have a strong opinion about this. So I changed it to your suggestion in https://github.com/LiskHQ/lisk-hub/pull/2225/commits/a0e2be96cf87f1e8f509a18b4f00ef739b6bd733

slaweet:
- updated filter placeholder to say '... and address ...', which made it longer then the width of the input, so I also increased the width to 200px
- updated edit placeholder to be same as in https://projects.invisionapp.com/d/main?origin=v7#/console/17570736/368615414/preview

PR #2258, opened 2019-07-22
Refactor search bar to use withData HOC - First part of #2025

### LINKED ISSUE #2025: Replace search actions and reducers with withData HOC
### Descriptions
This is a child ticket for #2020 .
With a short review over the code, we can find several redundant or obsolete actions. accordingly there are cases defined in the reducers switch-case statement which can be refactored or even removed. 
The naming of the actions seem to need revision.

slaweet:
### What issue have I solved?

The first part of #2025

- [x] Create withData HOC 
- [x] Use withData HOC for search suggestions to showcase it works
- The code contains TODO comments that explain what I intend to do in the follow-up PR that will finish #2025. 

### How has this been tested?

Check that all tests still pass. 
Check search bar works.

### Review checklist

### DIFF
--- a/src/utils/withData.js
+++ b/src/utils/withData.js
@@ -1,3 +1,5 @@
+/* istanbul ignore file */
+// TODO create unit tests for this file
 import { connect } from 'react-redux';
 import React from 'react';
 import { getAPIClient } from './api/network';
--- a/src/utils/withData.js
+++ b/src/utils/withData.js
@@ -2,46 +2,70 @@ import { connect } from 'react-redux';
 import React from 'react';
 import { getAPIClient } from './api/network';
 
-function withData({
-  autoload = true, propName, apiUtil, getApiParams,
-}) {
+function withData(apis = {}) {
   return function (WrappedComponent) {
     class DataProvider extends React.Component {
       constructor() {
         super();
 
         this.state = {
-          data: [],
-          error: '',
-          isLoading: false,
+          apis: Object.keys(apis).reduce((acc, key) => {
+            acc[key] = {
+              data: [],
+              error: '',
+              isLoading: false,
+            };
+            return acc;
+          }, {}),
         };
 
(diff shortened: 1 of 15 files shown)


src/utils/withData.js
```diff
@@ -0,0 +1,91 @@
...
+      apiParams: Object.keys(apis).reduce((acc, key) => {
+        acc[key] = apis[key].getApiParams(state, ownProps);
+        return acc;
+      }, {}),
+    });
+
+    return connect(
+      mapStateToProps,
+    )(DataProvider);
+  };
```
massao: IMO, should set the displayName.
https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

src/utils/withData.js
```diff
@@ -0,0 +1,91 @@
...
+          this.setApiState(key, { error, isLoading: false });
+        });
+      }
+
+      render() {
+        const { apiClient, apiParams, ...restOfProps } = this.props;
+        return (
+          <WrappedComponent
+            {...{
+              ...Object.keys(this.state.apis).reduce((acc, key) => {
```
massao: Since it uses the `Object.keys` multiple times, isn't  it possible to set this into some variable on the constructor so there is no  need to fetch the keys from the Object again?

src/utils/withData.js
slaweet: Good point. Fixed https://github.com/LiskHQ/lisk-hub/pull/2258/commits/fbfc624bdaff2a2d6656af62edbc191cff6356ed

src/utils/withData.js
slaweet: Fixed in https://github.com/LiskHQ/lisk-hub/pull/2258/commits/e383e75411c439b01f35abcd7d6cea6d804d2aae

PR #2260, opened 2019-07-22
Update account initialization banner to the new design - closes #2257

### LINKED ISSUE #2257: Update account initialization banner to the new design
### Description
As a new user, that login for first time into the application, show be able to see a small banner that let the user know that needs to initialize the account as soon as possible, before start doing more transactions.

This is the design that must be follow, [design](https://projects.invisionapp.com/d/main/?origin=v7#/console/17570736/374296680/preview)

### Acceptance criteria

- The banner should display every time that user login into the application when the account it is not initialize.

- The banner should be display in all pages if the account it is not initialize.

- The banner should contain a button that will redirect to the send component to initialize the account.

- The banner should include a link with the copy of Learn more that should redirect to this page [link](https://lisk.io/help-center/account-security/should-i-initialize-my-lisk-account)

- After account was initialize the banner should go away and not show it any more in any page

osvaldovega:
### What issue have I solved?

-- #2257 

The PR contains a new feature element that has a few details/component. Those component are:
- icon
- text to display/show
- link that will redirect to an external page (lisk documentation)
- button to init the account (on click will redirect to the send component to init the account)

This component looks like the design and it is intended to be only display when user create an account for first time and there is no transactions at all.

### How has this been tested?

The tests that I made to check this is working properly are the follow:

- login with an account already init, then the banner should NOT be visible.
- login with an uninitialize account, then the banner should be visible in all pages after login, if logout banner will NOT be display.
- login with an uninitialize account, then do click in the Initialize account button in the banner and send the first transaction to init the account after this the banner should disappear. After this logout and login with the same account and the banner should NOT show up at all.

### Review checklist


src/components/toolbox/flashMessage/flashMessage.js
```diff
@@ -0,0 +1,66 @@
...
+          >
+            {buttonText}
+          </SecondaryButtonV2>
+        )
+        : null
+    }
+  </div>
+);
+
+FlashMessage.defaultProps = {
```
massao: Since  you are using the defaultProps, should be using also the propTypes, to define the type of each props, and also if it's required or not, required props, doesn't need a default value.

PR #2281, opened 2019-07-26
Remove the peers reducer and use network reducer - closes #2024

### LINKED ISSUE #2024: Remove the peers reducer and use network reducer instead.
### Descriptions
This is a child ticket for #2020 .
With a short review over the code, we can find several redundant or obsolete actions. accordingly there are cases defined in the reducers switch-case statement which can be refactored or even removed. 

- https://github.com/LiskHQ/lisk-hub/blob/development/src/actions/peers.js and https://github.com/LiskHQ/lisk-hub/blob/development/src/store/reducers/peers.js should be removed 
- all 254 places in the code that were using it need to be updated to use https://github.com/LiskHQ/lisk-hub/blob/development/src/store/reducers/network.js instead
```
$ git grep -i peer | wc -l
     254
```
The grep above should find nothing after all is done.

osvaldovega:
### What issue have I solved?

-- #2024 

In this PR all the peers code (reducers, actions and functions) were remove and use network instead, for this some fo the logic had to be update as some of the functionality for get the liskAPIClient works different than peers.
In consequence, unit tests were update too.  

**NOTE:** About the E2E tests the only update that was needed it was a wait function, the reason for this wait function is because the test is doing the assertions for some information that is not ready and because of that the assertion failed, after checked this behavior with @Efefefef he decide to use this wait function with a few ms to see that will give the enough time to the test for make the assertion.

### How has this been tested?

For test the new functionality I had to do a manual test of all the app and check that everything still works as expected, once that I check all the app manually, I run jest and cypress to ensure that all functionality works as before.

### Review checklist


test/cypress/features/delegates/delegates.js
```diff
@@ -9,6 +9,7 @@ import urls from '../../../constants/urls';
 const txConfirmationTimeout = 20000;
 
 Then(/^I see (\d+) delegates on page$/, function (number) {
+  cy.wait(100);
```
slaweet: @osvaldovega Any idea why is this wait needed and it wasn't needed before?
@Efefefef any idea for how to not need the wait?

Same for the other 3 `cy.wait(100);` in https://github.com/LiskHQ/lisk-hub/pull/2281/files

Efefefef:
@Efefefef Test HW wallet

Efefefef:
@slaweet I suppose failures are not related to this pr and started in development branch as well. I am going to fix all of this failures in separate PR asap

Issue #2159, opened 2019-06-20
Refactor send form component - part 1


slaweet:
## Current behaviour
https://github.com/LiskHQ/lisk-hub/blob/development/src/components/sendV2/form/form.js is over 600 lines of code

it's one of the worst in the whole project, you can get the longest files by runing:
```
wc -l `find src/* -type f | grep -v assets | xargs` | sort -n
```

## Expected behaviour

https://github.com/LiskHQ/lisk-hub/blob/356c80034da1e6b7500c4d2ce68b986f10a229a0/src/components/send/form/formBase.js#L2-L7

and there will be a follow-up https://github.com/LiskHQ/lisk-hub/issues/2375

PR #2310, opened 2019-08-02
Send form refactor proposal - Closes #2269

### LINKED ISSUE #2269: Investigate the potential improvements required for the Send form component
As described in #2159 this component contains 600 lines of code. we need to investigate and define the desired improvements. This will help us have a clear understanding of the required actions and increase the accuracy of estimations and acceptance criteria.

slaweet:
### What issue have I solved?

#2269

This is a PR for Spike ticket, so it contains only the proposal for the refactoring. There is a separate ticket for the implementation https://github.com/LiskHQ/lisk-hub/issues/2159

### How has this been tested?

There are no changes that need to be tested.

### Review checklist

### DIFF
--- a/src/components/send/form/form.js
+++ b/src/components/send/form/form.js
@@ -1,4 +1,10 @@
 // eslint-disable-line max-lines
+// TODO Split this Send form component into following 4:
+// - FormWrapper - decides to render FormLSK or FormBTC based on activeToken
+// - FormLSK - Contains all LSK-specific functionality (message, static fee) and renders FormBase
+// - FormBTC - Contains all BTC-specific functionality (dynamic fee selection) and renders FormBase
+// - FormBase - Contains all functionality common to LSK and BTC (Box, address input,
+//     amount input, confirm button)
 import React from 'react';
 import Converter from '../../converter';
 import { PrimaryButton } from '../../toolbox/buttons/button';
@@ -132,6 +138,7 @@ class Form extends React.Component {
     }
   }
 
+  // TODO move `state.fields` into parent send component and ifDataFromPrevState can be deleted
   ifDataFromPrevState() {
     const { prevState } = this.props;
     // istanbul ignore if
@@ -144,6 +151,7 @@ class Form extends React.Component {
     }
   }
 
+  // TODO move `state.fields` into parent send component and ifDataFromUrl can be deleted
   // istanbul ignore next
   ifDataFromUrl() {
     const { fields = {} } = this.props;
@@ -200,6 +208,7 @@ class Form extends React.Component {
     }));
   }
 
+  // TODO move bookmark validation into a separate util or component
   // eslint-disable-next-line max-statements
   validateBookmark() {
     const { token, networkConfig } = this.props;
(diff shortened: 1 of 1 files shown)


src/components/send/form/form.js
```diff
@@ -1,4 +1,10 @@
 // eslint-disable-line max-lines
+// TODO Split this Send form component into following 4:
+// - FormWrapper - decides to render FormLSK or FormBTC based on activeToken
+// - FormLSK - Contains all LSK-specific functionality (message, static fee) and renders FormBase
+// - FormBTC - Contains all BTC-specific functionality (dynamic fee selection) and renders FormBase
+// - FormBase - Contains all functionality common to LSK and BTC (Box, address input,
+//     amount input, confirm button)
```
osvaldovega: @slaweet I understand the component for FormLSK and FormBTC but the FormBase looks like the the parent component and then I don't understand the FormWrapper.

If I understand right will be something like.

- FormBase
   - FormWrapper
      - FormLSK/FormBTC

In any case I think that most important is the FormLSK and FormBTC what will be very good for the send.

slaweet: What I meant is something like this (very simplified):

```
const FormWrapper = ({ activeToken, ...props }) => (
  activeToken === 'BTC' ? <FormBTC {...props} /> : <FormLSK {...props} />
);
```
```
const FormLSK = (props) => (
  <FormBase {...props} >
    <StaticFee />
    <MessageInput />
  </FormBase>
);
```
```
const FormBTC = (props) => (
  <FormBase {...props} >
    <DynamicFee />
  </FormBase>
);
```
```
const FormBase = ({ children }) => (
  <Box>
    <AddressInput />
    <AmountInput />
     {children}
   <ActionButtons />
  </Box>
);
```

I used a similar pattern in Request https://github.com/LiskHQ/lisk-hub/blob/development/src/components/request but there, `RequestWrapper` corresponds to `FormBase` in this proposal and `index.js`(`Request`) is proposed `FormWrapper`. So I would change the naming here to be consistent with the request.

reyraa: I'd go with dividing the components based on their purpose rather than the values they use. An amount input, is gonna validate the amount and show the fiat equivalent not matter the token. the difference will only be the pure validation function.
Same goes for recipient input. Address validation and fetching the bookmarks is exactly the same. I'd suggest not duplicate this functionality in two components (FormBTC and FormLSK). Instead, better have it central with a pure validation function dedicated for each. I assume with proper architecture, fetching bookmarks must be no different other than passing different parameters to the same function.
On thing is to decide to show or hide the Message, which is as simple as it can get.

slaweet: @reyraa I think you miss-read what I proposed above. 

AddressInput and AmountInput is in the FormBase component, so it's common for BTC and LSK. 

What is not described above is how they get the token-specific validation functions. Token-specific component can easily pass a token-specific function for that into the FormBase, like `validateAddress` below:

```
const FormBase = ({ children, validateAddress }) => (
  <Box>
    <AddressInput validateAddress={validateAddress} />
    <AmountInput />
     {children}
   <ActionButtons />
  </Box>
);
```

slaweet: And also I didn't mean that `AddressInput` and `AmountInput` necessarily need to be separate components (depends on their size). I just put them to illustrate what functionality will be in which component.

reyraa: Ok, I got you now. I agree with what proposed then.

yasharAyari: I think it is better to have on parent form and define the differences inside the form component like `AddressInput` and `AmountInput`. Then you can move the validation methods to a utility.

PR #2321, opened 2019-08-07
Network Switcher - Remaining Parts - Closes #1982

### LINKED ISSUE #1982: Network Switcher - Remaining Parts
1. Used custom node address is saved only after login, should be saved on successful connect
1. Would be nice to have Enter working as Connect click in custom node address field
1. What should happen on app start if the last used network was custom node which is not available at the moment and connection failed?

osvaldovega:
### What issue have I solved?

-- #1982 

The fix in this PR is implement it in header/signInHeader/signInHeader.js file.

- [x] Validate when the network is change and if it custom node change only if address is entered.
- [x] User be able to submit the custom url address pressing the enter button.
- [x] If custom node network is selected but no url address is entered then keep the previous network.

### How has this been tested?

1. Run the app.
2. Go to the Sign In page.
3. Open the developers tools and check the network value in the settings in local storage.
4. Then change the network from mainnet to testnet.
5. Then change network selecting only custom network (settings shouldn't be update)
6. Enter custom node URL address.
7. press enter (network should be change now).

### Review checklist

### DIFF
--- a/src/components/header/signInHeader/signInHeader.js
+++ b/src/components/header/signInHeader/signInHeader.js
@@ -8,7 +8,6 @@ import { Input } from '../../toolbox/inputs';
 import { addHttp, getAutoLogInData, findMatchingLoginNetwork } from '../../../utils/login';
 import getNetwork, { getNetworksList } from '../../../utils/getNetwork';
 import { parseSearchParams } from '../../../utils/searchParams';
-
 import Icon from '../../toolbox/icon';
 import UserAccount from '../../topBar/accountMenu/userAccount';
 import networks from '../../../constants/networks';
@@ -16,6 +15,7 @@ import styles from './signInHeader.css';
 import formStyles from '../../send/form/form.css';
 import Dropdown from '../../toolbox/dropdown/dropdown';
 import Spinner from '../../spinner/spinner';
+import keyCodes from '../../../constants/keyCodes';
 import svg from '../../../utils/svgIcons';
 
 class Header extends React.Component {
@@ -40,6 +40,7 @@ class Header extends React.Component {
     };
     this.toggleDropdown = this.toggleDropdown.bind(this);
     this.handleSettingsToggle = this.handleSettingsToggle.bind(this);
+    this.onConnectToCustomNode = this.onConnectToCustomNode.bind(this);
   }
 
   // eslint-disable-next-line class-methods-use-this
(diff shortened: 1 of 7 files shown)


osvaldovega:
thx @slaweet I just update the code with a minor change and then the error that you mentioned before is fixed

src/components/header/signInHeader/signInHeader.js
```diff
@@ -8,38 +9,50 @@ import { Input } from '../../toolbox/inputs';
...
     this.toggleDropdown = this.toggleDropdown.bind(this);
     this.handleSettingsToggle = this.handleSettingsToggle.bind(this);
+    this.onConnectToCustomNode = this.onConnectToCustomNode.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps) {
+    const { address } = this.props;
+    if (address !== nextProps.address) {
+      this.setState({ address: nextProps.address });
+      return false;
```
slaweet: Why component shouldn't update in this case? I think it should always update, so I think it should use `componentDidUpdate` instead of `shouldComponentUpdate`. Maybe I'm missing something...

osvaldovega: you are totally right @slaweet I was using the incorrect lifecycle, I just made the change and is better now

osvaldovega:
hey @Efefefef I just fixed the code after your review.
-[x] I update the success message text, I consulted this with Matt and he gave me the correct text.
-[x] I fixed the bug about save the custom network when clicked in the input.
-[x] Fixed the duplicated toaster for error message.
-[x] Fixed the network switcher error message if user is coming from another page. However everything else about how the app should interact when user is in different pages I think is out of the scope of this PR, so in case you find another bug related other pages and not this one in specific I can check it in a different PR.

Efefefef:
Great! Two more things
🐛 We don't need to show connection success toaster on every /signin page open. As we assume we stay connected to the same network after signing out it is needed only on app start.
🐛 We don't show testnet/mainnet connection failed toaster on app start in case of any problems with connection to them. However, we started to show the succesfull connection toaster on app start. What do you think does it matter or it's a rare case and showing error toaster on clicking sign in is enough? In that case it might make sence to remove successfull connection toaster as well

osvaldovega:
thx @Efefefef for the review, you are correct, I just removed it.
Everything works exactly the same just without the success toast for the connection

Efefefef:
Great 👍

PR #2346, opened 2019-08-13
switching mainnet and testnet is mismatched - closes #2030

### LINKED ISSUE #2030: switching mainnet and testnet is mismatched
### Steps to reproduce

1.Sign-in to TestNet on LiskHub 1.16.0
2.Setting
3.Switch network(Main-/Testnet,Custom) to OFF
4.Dashboard
5.logout
6.Sign-in 
7.Unfortunately,sign-in as Testnet,not Mainnet

### Actual result
switching mainnet and testnet is mismacched

It remembers the network which you have logged onto last time
So if you switch network in the UI, but dont actually log on, it won't remember that you switched

### Expected result
I think the next sign-in should be Mainnet when switched network(Main-/Testnet,Custom) to OFF.

### Version
LiskHub 1.16.0

none

osvaldovega:
### What issue have I solved?

-- #2030 

The solution consists in check in the settings if the network switcher is on or off and based on that at the time of logout update the settings to switch to Mainnet in case the actual network is Testnet or Custom Node.

So in other words.

1. If network switcher is off the current network is Mainnet in this case there is nothing to change.
2. If network switcher is on and current network is Mainnet in this case there is nothing to change.
3. If network switcher is on and current network is Testnet or Custom Node, in this case there is nothing to change.
4. If network switcher is on and current network is Testnet or Custom Node, and use login and from settings page disable (OFF) the network switcher, then when he logout the code check this and if actual network it is different from Mainnet and network switcher if OFF then update the settings and network to default on (Mainnet).

### How has this been tested?

1. Sign-in to TestNet
2. Go o Setting
3. Disabled network switcher.
5. Logout
6. Sign-in again
7. Actual network should be Mainnet

### Review checklist

### DIFF
--- a/src/components/setting/index.js
+++ b/src/components/setting/index.js
@@ -5,6 +5,7 @@ import Setting from './setting';
 import { settingsUpdated } from '../../actions/settings';
 import { toastDisplayed } from '../../actions/toaster';
 import { accountUpdated } from '../../actions/account';
+import { networkSet } from '../../actions/network';
 import { getActiveTokenAccount } from '../../utils/account';
 
 const mapStateToProps = state => ({
@@ -17,6 +18,7 @@ const mapStateToProps = state => ({
 
 const mapDispatchToProps = {
   accountUpdated,
+  networkSet,
   settingsUpdated,
   toastDisplayed,
 };
--- a/src/components/setting/setting.js
+++ b/src/components/setting/setting.js
@@ -9,7 +9,6 @@ import links from '../../constants/externalLinks';
 import settingsConst from '../../constants/settings';
 import styles from './setting.css';
 import txTypes from '../../constants/transactionTypes';
-import networks from '../../constants/networks';
 import SecondPassphraseSetting from './secondPassphrase';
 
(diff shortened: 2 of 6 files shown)


osvaldovega:
good catch @Efefefef I just made the change. You can check it again

Efefefef:
🐛 The network is changed to Mainnet right after logging out if you switched off the network switcher
Do you think it is the right thing to do? I may want to stay in the last used network. It is not said anything about implicitly changing your network: 'Enable a network switcher that lets you select testnet or custom node when logging in.'

Another thing: If you turn off the network switcher and reload the app the network stays the same. It is contrary to the logic described above. What should happen in that case?

I don't like the idea of this ticket in the first place, so please let @reyraa decide

osvaldovega:
Hey @Efefefef I just update the PR, I made the fix based on your comments... in other hand well yes the idea is that if there is no network switcher active then the app should be working only with Mainnet expect when you are logged, then after logout this should change to mainnet

Efefefef:
Cool! Now its consistent with new logic

PR #2359, opened 2019-08-15
Floating point must match selected language/locale - Closes #2353

### LINKED ISSUE #2353: Amount floating point must match the selected locale/language
As user; I should see the floating point must match my language of choice. 

### Acceptance criteria 
 - No thousand separator is required.
 - The digits must be separated by dot (.) in English and comma(,) in German.
 - If the user pastes a value with thousand separator and floating point, the input shows error.
 - The inserted value is assumed to be in the current language of Lisk Hub. hence, if the separator doesn't match the current language, the input shows error. for example, 1.12 is an invalid input if the current language is German.

massao:
### What issue have I solved?

#2353

- [x] Converter component;
- Amount input validations;
  - [x] Request;
  - [x] Send;
- [x] Send Form -> Fee values;
- [x] Send Summary -> values;
- [x] Delegate information (vote weight, productivity);

### How has this been tested?

It's needed to set a `feature-flag-language` into localStorage to see the option in settings. 
```js
localStorage.setItem('feature-flag-language', true);
```

- Go to Settings
- Change language
- All the values should be updated  to reflect the selected language, and also the validation of input fields.

### Review checklist

### DIFF
--- a/src/components/request/request.test.js
+++ b/src/components/request/request.test.js
@@ -3,11 +3,6 @@ import { mount } from 'enzyme';
 import Request from '.';
 import accounts from '../../../test/constants/accounts';
 
-jest.mock('../../i18n', () => ({
-  t: v => v,
-  language: 'en',
-}));
-
 jest.mock('../converter', () => (
   function ConverterMock() {
     return <span className="converted-price" />;
--- a/src/components/request/request.test.js
+++ b/src/components/request/request.test.js
@@ -3,6 +3,11 @@ import { mount } from 'enzyme';
 import Request from '.';
 import accounts from '../../../test/constants/accounts';
 
+jest.mock('../../i18n', () => ({
+  t: v => v,
+  language: 'en',
+}));
+
 jest.mock('../converter', () => (
   function ConverterMock() {
     return <span className="converted-price" />;
--- a/src/components/votes/votesTab.js
+++ b/src/components/votes/votesTab.js
@@ -9,7 +9,7 @@ import { Input } from '../toolbox/inputs';
 import LiskAmount from '../liskAmount';
 import routes from '../../constants/routes';
 import styles from './votesTab.css';
-import { formatBasedOnLocale } from '../../utils/lsk';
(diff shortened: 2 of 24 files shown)


src/components/request/request.test.js
```diff
@@ -3,6 +3,11 @@ import { mount } from 'enzyme';
 import Request from '.';
 import accounts from '../../../test/constants/accounts';
 
+jest.mock('../../i18n', () => ({
+  t: v => v,
+  language: 'en',
+}));
+
```
slaweet: This looks like something that might be worth to do globally in `config/setupJest.js`. We are already mocking `i18next.t` in `config/setupJest.js:26`, so you just can mock also `i18next.init` that is used in `i18n/i18n-setup.js` to return objects with the language you want.

src/components/votes/votesTab.js
```diff
@@ -154,7 +156,10 @@ class VotesTab extends React.Component {
                     : '-'}
                 </div>
                 <div className={`${grid['col-sm-2']} ${grid['col-lg-1']}`}>
-                  {vote.productivity !== undefined ? `${vote.productivity}%` : '-'}
+                  {vote.productivity !== undefined
+                    ? `${formatAmountBasedOnLocale({ value: vote.productivity, locale: i18n.language })}%`
```
slaweet: Passing `locale` into `formatAmountBasedOnLocale` in all places doesn't look very DRY. 
I would `import i18n from '../../i18n';` in `src/utils/formattedNumber.js` and default to `i18n.language` if `locale` is not provided. 

I know it makes `formatAmountBasedOnLocale` less pure, but I think it's worth it.

massao:
> The story was changed after our discussion.
> Please check the updated story and implement what is needed 👍

@Efefefef Everything is already as described on the story, the only point that should be noticed is the first one: 
```
- No thousand separator is required.
```
Should only be related to the input field, the story should be updated to reflect the conclusion of the discussion of breakout session. @reyraa

PR #2363, opened 2019-08-15
Add discreet toggle to the sign-in page - Closes #2332

### LINKED ISSUE #2151: Discreet mode
As a user, I'd like to hide my sensitive information in public places so that my private informations are safeguarded.

### Description
To give users an option to easily open their Hub application in front of others without exposing their sensitive information, we can introduce a feature that allows users to hide and show the **amount of transactions** and the **account balance** at any time.
This can be limited to the information of the account that the user has signed in with. 

Since often when users want to run the application in a public environment they want their information to be hidden from the start, we can add an option in the sign in page which enables users to sign in in discreet mode.

Users must be able to enable and disable this functionality before signing in and afterward.
This option must be available dedicated to the settings page.

### Acceptance criteria
1. Allow the user to discreetly sign in.
2. After a successful sign in, transaction amounts and the account balance should be blurred out.
3. When in discreet-mode, users should not see their sensitive information after leaving the screen and coming back to any screen with sensitive information.
4. Users should have an option to enable/disable the discreet mode in settings page.
5. This feature only should affect the sensitive information of the signed-in account.
6. By the default this feature must be disabled.

### LINKED ISSUE #2332: Add discreet toggle to the sign-in page
### Description
We want to add a switch/ an option in sign-in page so the users be able to toggle the discreet mode while in the sign-in page. The designs are available on [InVision](https://projects.invisionapp.com/d/main#/console/18170373/379164944/preview).

osvaldovega:
### What issue have I solved?

Closes #2332 
Closes #2151

This PR contains the following implementations:

- [x] Create DiscreetModeToggle component.
- [x] Added into the Sign in page.
- [x] Remove the flags in settings component as the implementation is finished.
- [x] Unit test created for the DiscreetModeToggle component

**NOTE:**The discreet mode element was added in the Sign In page centered and between the passphrase input and footer buttons, as in the design show up as part of the passphrase element.
This was discussed with Dom and they are still working to see where will be the final design for all the arrange of elements in this page. So in summary they agree in have the discreet mode element where it is right now in this PR.

### How has this been tested?

1. Run the app.
2. Go to the Sign In page, enable the discreet mode
3. Sign in
4. All the values in the app should be blur (all values related to the owner account).
5. Go to setting and the Discreet mode option should be enabled
6. Then disabled the option
7. Check the values again in the app, everything should be visible.
8. Sign out.
9. Go to Sign in page again, discreet mode should be disabled.

Discreet Mode could be enable/disable in Sign In page and Settings page

### Review checklist


src/components/toolbox/icon/index.js
```diff
@@ -80,6 +82,8 @@ export const icons = {
   dashboardIconActive,
   delegatesIcon,
   delegatesIconActive,
+  discreetModeOff,
+  discreetModeOn,
```
slaweet: Keeping the alphabetic order. Great 👍

src/components/login/login.js
```diff
@@ -187,8 +188,15 @@ class Login extends React.Component {
                   maxInputsLength={24}
                   onFill={this.checkPassphrase}
                 />
-              </div>
 
+                <DiscreetModeToggle
+                  className={styles.discreetMode}
+                  isEnable={settings.discreetMode}
+                  onClick={() => settingsUpdated({ discreetMode: !settings.discreetMode })}
```
slaweet: IMO `onClick` and `isEnable` should be only inside the `DiscreetModeToggle` component, because no matter where we use the toggle, this behavior should always be the same and we would be duplicating the code in all pages that use the toggle. This implies `DiscreetModeToggle` needs `HOC` similar to https://github.com/LiskHQ/lisk-hub/blob/dc58d3862d30e2c1b9be4c3e49d5478a6563cd47/src/components/discreetMode/index.js

osvaldovega:
thx @slaweet you are right, I forget that part I made the changes, so it is ready.
please check it again

PR #2366, opened 2019-08-19
Create a refactor proposal for logic and structure of hwWallet in electron and react - closes #1954

### LINKED ISSUE #1954: Create a refactor proposal for logic and structure of hwWallet in electron and react

At this moment the hwWallet code is split in **hwWallet middleware** functions in **utils** and **utils/api**, in these 3 places the hwWallets are being check and based on an action this trigger an specific event, some information is being set in the localStorage and in the same time in the redux store.

So the problem at this moment is that some data could be duplicated in the project and store in localStorage and this could be not necessary.

### Suggested solution

Create a generic/modular solution that helps to handle any hwWallet device detection and interaction for dont have to change or add files in the future to handle a new hardware wallet device, so the communication and interaction could be the same in the react project and then in the electron part implement the logic for the specific device.

So for the react part we can update or modify the actions, reducers, local storage and then middlewares and find the better place to handle the ipc messages and the dispatch of the actions, then check if the actual structure that we have for this can be redefine and then remove it from the local storage

osvaldovega:
### What issue have I solved?

-- #1954 

The main idea of this proposal is to implement a better way to handle all the functions related to hardware wallets, it doesn't matter what type of hardware wallet, the idea can be split in 3 parts.

#### 1. Hardware Wallet manager (library/module).
This is the wrapper code that will can create and will be expose all the functions related any hardware wallet device, so in other words those expose function that this module will have will bee the same for all hardware wallets and these are the ones that we will use en electron.

**NOTE:** The idea of this is NOT to create our own HW implementation, is just create a wrapper file that help us to expose the functions that we need in a easy way, so if in the future something change about trezor connect library or ledger library should be update and handle here no in electron or react. If we need to remove support for any device just need to remove the hw library deepency or add a new one in case of support other hw device.
There is a possibility to be use in Lisk Mobile too.

#### 2. Electron side
In electron side there is no logic only a file that will be use to communicate the react side request with the library/module all this communication will be using the ipc messages (check win file under app/src/module/win.js).
So once this file will be implemented there is no need in the future to change or update, this will ONLY work for communicate the send/requests from the react side to the HW devices.

Screen Shot 2019-08-19 at 13 18 42

#### 3. React side
Right now there is a hwManger file under utils and contains a lot of logic for send/request messages to electron part and then to the hw devices. This file should be very simple and small and contain only functions to request data/singtx/etc.. from a specific device and then wait for the answer.
**NOTE:** If any of the received answers will be store in REDUX then have more logic for this file to be a middleware.

Screen Shot 2019-08-19 at 13 32 27

### DIFF
--- a/app/src/hwManager.js
+++ b/app/src/hwManager.js
@@ -1,4 +1,19 @@
 /* eslint-disable */
+/**
+ * TODO - remove the logic of this file and use it in a external library/module
+ * This file will be on charge of communicate the electron app with the HW module/library
+ * this file should only containe a few functions.
+ * 
+ * Init - Function - For observ/subscribe or listen to changes.
+ * getDevices - function - that retturn all connected devices.
+ * signInTX - function - sign in any kind of tx 
+ * getPublicKey - function - get public key for other users like getAccounts (if the public key is used for retrive other information then dont need this func)
+ * getAccounts - function - retrive all accounts from device
+ * ping - function - function that check device status
+ * 
+ * All of the above should be use for any HW device that will be implement in the future, right now
+ * only Ledger Nano S, Ledger Nano X, Trezor Model T and Trezor One.
+ **/
 
 import win from './modules/win';
 import { createCommand } from './utils';
--- a/app/src/ledger.js
+++ b/app/src/ledger.js
@@ -1,3 +1,9 @@
+/**
+ * TODO - This file doesnt need to be here, this logic needs to be impleement in the module/library
+ * and how this works should be unknown for the electron part.
+ * Needs to remove duplicated code and improve logic for be more generic
+ */
+
 import { app, ipcMain } from 'electron'; // eslint-disable-line import/no-extraneous-dependencies
 import Lisk from '@liskhq/lisk-client'; // eslint-disable-line import/no-extraneous-dependencies
 import { LedgerAccount, SupportedCoin, DposLedger } from 'dpos-ledger-api'; // eslint-disable-line import/no-extraneous-dependencies
--- a/app/src/trezor.js
+++ b/app/src/trezor.js
@@ -1,6 +1,12 @@
 /* eslint-disable no-bitwise */
 /* eslint-disable */
 
+/**
+ * TODO - This file doesnt need to be here, this logic needs to be impleement in the module/library
+ * and how this works should be unknown for the electron part.
+ * Needs to remove duplicated code and improve logic for be more generic
+ */
+
 import { ipcMain } from 'electron'; // eslint-disable-line import/no-extraneous-dependencies
 import win from './modules/win';
 import {
--- a/src/utils/api/hwWallet.js
+++ b/src/utils/api/hwWallet.js
@@ -1,4 +1,15 @@
 /* eslint-disable */
+
+/**
+ * TODO - This file contains a lot of unnecessary logic and duplicated functions, this file should
+ * work as a communication center file for send and receive messages (data) from electron using the 
+ * IPC command.
+ * If the data will be store in REDUX then this needs to be a middleware and once that receive the 
+ * data submit the properly action.
+ * How the information is retrive from electron part should be unknown for this side of the app.
+ * Have this kind of file with this logic is necessary ONLY if we  will support HW devices in web platform
+ */
+
 import TransportU2F from '@ledgerhq/hw-transport-u2f';
 import TrezorConnect from 'trezor-connect';
 import i18next from 'i18next';


slaweet:
I think the proposal is a good starting point, but IMO there is still a lot more to cover in the proposal, I have the following suggestions.


## Move hwManager from Electron to React
I think that most of the logic doesn't have to go through electron. I think we need the electron only to get a signal when hw wallet is connected/disconnected, but for other stuff (`getAccounts`, `getAddress`, `signTx`, `getPublicKey`) we don't have to go through the electron layer.  IMO It would be much simpler if the HW Manager is on the React app side, communicates directly with the hardware wallets and uses electron only when necessary.

For example, `signTx(account, transaction)` should be all on React side and should call:
- for Ledger:
https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/utils/api/hwWallet.js#L94-L95 where `liskLedger` comes from https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/utils/api/hwWallet.js#L78 where `DposLedger` comes from https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/utils/api/hwWallet.js#L16

- for Trezor:
https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/utils/api/hwWallet.js#L139
where `TrezorConnect` comes from https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/utils/api/hwWallet.js#L14

## More detailed descriptions of HwManager methods
I think this proposal PR should contain similar explanation like I provided above for `signTx(account, transaction)` also for all other proposed methods of `HwManager` (`getAccounts`, `getAddress`, `getPublicKey`) so that it's clear that the external dependencies contain functions for what we need.

## Other things I miss in this proposal:
- describe what parameters accept functions `getAccounts`, `getAddress`, `getPublicKey`, etc.
- describe the structure of what should be stored in the account object in redux. Currently, it is: https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/actions/account.js#L195-L196
- Other functions needed in hwManager not mentioned so far, e.g.:
  - something to handle wallet connected/disconnected, 
  - getDeviceList https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/utils/api/hwWallet.js#L51
  - signMessage https://github.com/LiskHQ/lisk-hub/blob/4c68d889038ae8102160b887e5f934edc115b133/src/utils/api/hwWallet.js#L134

slaweet:
I would also like to write some code to check we have the same idea for what the hwManager should look like in mind. I think it should be something like this: 

```js
// this is /src/utils/hwManager.js
import { getAccount } from './api/lsk/account';
import ledgerManager from './ledgerManager';
import trezorManager from './trezorManager';

const managers = {
  'ledger': ledgerManager,
  'trezor': trezorManager,
}

function getVendorHwManager(loginType) {
  return managers[loginType];
}

export async function signTx(account, transaction) {
  const vendorHwManager = getVendorHwManager(account.loginType);
  return await vendorHwManager.signTx(account, transaction)
}

export async function getAccount(liskAPIClient, loginType, deviceId, accountIndex) {
  const vendorHwManager = getVendorHwManager(loginType);
  const publicKey =  await vendorHwManager.getPublicKey(deviceId, accountIndex);
  return await getAccount({ liskAPIClient, publicKey });
}

// TODO add the implementation of other functions, like getDeviceList, getAccounts, getAddress, getPublicKey, signMessage, ...
```

osvaldovega:
Thx @slaweet you have a good point I think we can use the HWManager directly in react, will be even easier.

About your questions and suggestions I will do the following.

1. I will check all the ledger and trezor functions to see those that are in common (talking about functionality) and see what params they receive once that I get that I will post it.

2. About the response of those functions well at this point I don't know because I need to try both and see what they return for any specific function and see how I can normalize the data, after normalize the data all the responses will be the same it doesnt matter the device and then we will see how the data will looks in the redux store.

3. About your last example, well yes is something like that, is a common interface well you will get one common function name for all the hw devices instead a specific one depending the device. Your examples is exactly what I was thinking.

```

// -------- Trezor.js  ----------
function init() {
  //login
}

function getAccounts() {
  //login
}

function singTx() {
  //login
}

// -------- ledger.js -----------
function init() {
  //login
}

function getAccounts() {
  //login
}

function singTx() {
  //login
}

// ------- index.js --------
export default {
  trezor,
  ledger,
}

// --------- HWManager ---------
import hwVendor from './index';

function initialize() {
  trezor.init();
  ledger.init();
}

function getAccounts(deviceType, data) {
  return hwVendor[deviceType].getAccounts();
}
```

massao:
I like the idea of having the hwManager as a separate library, that abstract the main functions and normalize the necessary data. Also that would enable that just by changing the transport, the same library being useable on the lisk-mobile in the future.  

Also since some hw libraries can be only node based, it may not be possible to use it directly from the React App. but it would be interesting if it was possible to use the hwManager more like a normalizer in-between the hw library and the React App, and not having to use Electron at all.  
Otherwise we could use the hwManager as a normalizer and it sends to electron the correct information, so in the React App we have the same API independent of the hwWallet. And we would have scoped inside the library all calls to Electron, instead of using the IPC directly from React.  

Regarding the proposal, IMHO since this is a big reestructuring of the hardware wallet, I would like to have a proper `md` file, with a more complete explanation of what is the purpose, and also how the new structure should behave, also working as a kind of documentation in the future, instead of the `TODO` comments on the code.

PR #2387, opened 2019-08-22
Enabling BTC shouldn't require app restart or logout/login - Closes #2211

### LINKED ISSUE #2211: BTC: You need to restart the app after enabling to make it availible
### Steps to reproduce

1. Login
1. Enable BTC
1. Check out the available tokens

### Actual result

No BTC

### Expected result

BTC is there

image

massao:
### What issue have I solved?

#2211

Added action to get necessary accounts when account list changes.  
Added middleware to accounts middleware, to trigger the above action when the `settings.token.list` changes.

### How has this been tested?

1. Make sure that BTC is disabled on settings;
2. Login, should only have the LSK account listed;
3. Go to settings and enable BTC;
4. It should show BTC on the token switcher and dashboard.

### Review checklist

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -164,26 +164,17 @@ async function getAccounts(tokens, options) {
   }, Promise.resolve({}));
 }
 
-export const updatedLoggedinAddresses = tokens => async (dispatch, getState) => {
-  const { network: networkConfig, account, settings } = getState();
-  const newTokens = (Array.isArray(tokens)
-    ? tokens
-    : [tokens]
-  ).filter(key => settings.token.list[key]);
-  if (!newTokens.length) return;
-
-  const [error, info] = await to(getAccounts(newTokens, {
-    networkConfig, passphrase: account.passphrase,
+export const updateEnabledTokensAccounts = token => async (dispatch, getState) => {
+  const { network: networkConfig, account } = getState();
+  const [error, result] = await to(getAccount({
+    token,
+    networkConfig,
+    passphrase: account.passphrase,
   }));
   if (error) {
     dispatch(errorToastDisplayed({ label: getConnectionErrorMessage(error) }));
   } else {
-    dispatch(accountUpdated({
-      info: {
-        ...account.info,
-        ...info,
-      },
-    }));
+    dispatch(accountUpdated(result));
   }
 };
 
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -164,6 +164,29 @@ async function getAccounts(tokens, options) {
   }, Promise.resolve({}));
 }
 
+export const updatedLoggedinAddresses = tokens => async (dispatch, getState) => {
+  const { network: networkConfig, account, settings } = getState();
+  const newTokens = (Array.isArray(tokens)
+    ? tokens
+    : [tokens]
+  ).filter(key => settings.token.list[key]);
+  if (!newTokens.length) return;
+
+  const [error, info] = await to(getAccounts(newTokens, {
+    networkConfig, passphrase: account.passphrase,
+  }));
+  if (error) {
+    dispatch(errorToastDisplayed({ label: getConnectionErrorMessage(error) }));
+  } else {
+    dispatch(accountUpdated({
+      info: {
+        ...account.info,
+        ...info,
+      },
+    }));
+  }
+};
+
 /**
  * This action is used on login to fetch account info for all enabled token
  *
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -164,7 +164,7 @@ async function getAccounts(tokens, options) {
   }, Promise.resolve({}));
 }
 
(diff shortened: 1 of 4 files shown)


src/actions/account.js
```diff
@@ -164,6 +164,29 @@ async function getAccounts(tokens, options) {
...
+
+  const [error, info] = await to(getAccounts(newTokens, {
+    networkConfig, passphrase: account.passphrase,
+  }));
+  if (error) {
+    dispatch(errorToastDisplayed({ label: getConnectionErrorMessage(error) }));
+  } else {
+    dispatch(accountUpdated({
+      info: {
+        ...account.info,
```
slaweet: Is this line needed?

src/actions/account.js
```diff
@@ -164,6 +164,29 @@ async function getAccounts(tokens, options) {
   }, Promise.resolve({}));
 }
 
+export const updatedLoggedinAddresses = tokens => async (dispatch, getState) => {
+  const { network: networkConfig, account, settings } = getState();
+  const newTokens = (Array.isArray(tokens)
+    ? tokens
+    : [tokens]
+  ).filter(key => settings.token.list[key]);
```
slaweet: Is it necessary to pass `tokens` into this action if you're reading `settings.token.list` inside anyway? Wouldn't be this enough? :

```suggestion
  Object.keys(settings.token.list).filter(key => settings.token.list[key]);
```

Currently, it feels like there's a part of the token filtering in the middleware and part in the action. IMO it's cleaner to put all token handling in one place, either here and don't pass any param, or in the middleware. and don't filter it here.

src/actions/account.js
```diff
@@ -164,6 +164,29 @@ async function getAccounts(tokens, options) {
   }, Promise.resolve({}));
 }
 
+export const updatedLoggedinAddresses = tokens => async (dispatch, getState) => {
```
slaweet: I think a more descriptive name would be
```suggestion
export const updateEnabledTokensAccounts = tokens => async (dispatch, getState) => {
```

src/actions/account.js
massao: Actually, since the function only fetches the accounts that were changes to true, it's needed so it doesn't erase the previous fetched accounts

slaweet: Ah got it. I was thinking that in https://github.com/LiskHQ/lisk-hub/blob/ae144e2546cec21d90ca9ebb3632b3150c37c890/src/actions/account.js#L148 it's not needed either, but there it's just account of one token, here it can be multiple...

Should this really allow to handle multiple tokens changed? I think even in the future if we have more tokens, the user will still enable one at a time. 

I'm asking because overall this new action feels like many of the actions we were removing in the refactoring of redux.

massao: Yeah I updated it to just fetch the updated token, and with that it's not needed to deestructure anymore.

src/actions/account.js
massao: Moved the filtering to the middleware, so the action is only called if the token was enabled, instead of triggering the action even if the token is being disabled.

src/store/middlewares/account.js
```diff
@@ -220,6 +221,14 @@ const accountMiddleware = store => next => (action) => {
     case actionTypes.accountLoggedOut:
       store.dispatch(emptyTransactionsData());
       break;
+    case actionTypes.settingsUpdated: {
+      const tokensList = action.data.token && action.data.token.list;
+      const token = tokensList && Object.keys(tokensList)[0];
+      if (tokensList[token]) {
```
slaweet: Now changing other settings (e.g. network switcher) causes an error in console because `tokensList` is not defined then. 
```suggestion
      if (tokensList && tokensList[token]) {
```

src/store/middlewares/account.js
```diff
@@ -220,6 +221,14 @@ const accountMiddleware = store => next => (action) => {
     case actionTypes.accountLoggedOut:
       store.dispatch(emptyTransactionsData());
       break;
+    case actionTypes.settingsUpdated: {
+      const tokensList = action.data.token && action.data.token.list;
+      const token = tokensList && Object.keys(tokensList)[0];
```
slaweet: `[0]` is quite hacky here and works only because we define 'BTC' before 'LSK'`.

```suggestion
      const token = tokensList && Object.keys(tokensList).find(token =>  tokensList[token]);
```

should work even if we have more tokens and enabling not just the first one.

src/actions/account.js
```diff
@@ -164,6 +164,20 @@ async function getAccounts(tokens, options) {
   }, Promise.resolve({}));
 }
 
+export const updateEnabledTokensAccounts = token => async (dispatch, getState) => {
```
slaweet: Now the name (plural) doesn't match the implementation (singular)
```suggestion
export const updateEnabledTokenAccount = token => async (dispatch, getState) => {
```

PR #2405, opened 2019-09-02
Implement send functionality with hw Manager - closes #2385

### LINKED ISSUE #2385: Implement send functionality with HWManager
The functionality of this task is to be able to send txs using the hd wallet and be able to confirm/decline the tx with the device.

osvaldovega:
### What issue have I solved?

-- #2385

This PR includes the functionality for sign tx with hardware wallets (trezor and ledger) when the user do a send action, because the reutilization of code this sign tx function is used for voting too, what cause that right now voting be possible to do as the implementation works for both events.

### How has this been tested?

**For Submit a Tx**
1. Build the app.
2. Run the app.
3. Login with Ledger or Trezor device.
4. Go to to Wallet.
5. Go to send.
6. Fill the form for send a tx.
7. Do click in the go to confirmation button.
9. Check the HW Wallet and see that all the values are correct.
10. Approve the tx.
11. Go back to Wallet and check the last tx.

**For Cancel a Tx**
1. Build the app.
2. Run the app.
3. Login with Ledger or Trezor device.
4. Go to to Wallet.
5. Go to send.
6. Fill the form for send a tx.
7. Do click in the go to confirmation button.
9. Check the HW Wallet and see that all the values are correct.
10. Cancel the tx.
11. Go back to Wallet and check that there is no submitted tx.

### Review checklist


libs/hwManager/manufacturers/trezor.js
```diff
@@ -1,39 +1,73 @@
...
 
+// TODO move this to utils file
 const hardeningConstant = 0x80000000;
 const getHardenedPath = index => [
   (44 | hardeningConstant) >>> 0,
   (134 | hardeningConstant) >>> 0,
   (index | hardeningConstant) >>> 0,
 ];
 
+// eslint-disable-next-line max-statements
```
slaweet: I think now is a good time to simplify this function and remove `eslint-disable` from here. I'd consider one with object destructuring, something like:
```js
const toTrezorGrammar = ({senderPublicKey, requesterPublicKey, recipientId, ...rest }) => {
const tx = {
  ...rest,
  recipient_id: recipientId,
  sender_public_key: senderPublicKey,
  requester_public_key: requesterPublicKey,
  ...
}
Object.keys(tx).forEach((key) => (!tx[key]) && delete tx[key]);

return tx;
```

libs/hwManager/manufacturers/trezor.js
```diff
@@ -46,24 +80,44 @@ const executeCommand = (transporter, {
...
-            },
-          );
-          return resolve(message.public_key);
-        }
-        default: {
-          // eslint-disable-next-line no-console
-          console.log(`No action created for: ${device.manufactor}.${action}`);
-          return reject(new Error(`No action created for: ${device.manufactor}.${action}`));
+      try {
+        switch (action) {
```
slaweet: Why do we need `HwManager.executeCommand` and `switch (action)`? IMO a cleaner API would be to have a method for each command:`HwManager.getPublicKey`, `HwManager.signTx`... And it's a matter of how to implement these methods in a way that they are DRY and don't repeat the same logic. Otherwise, this `switch` will be overgrown.

massao: That was intended to keep backward compatibility due to the currents PRs only changing the `"provider side"` of the hwManager, but I agree that it could be done in a way that for now it would just forward the call to the correct hw manager function, and when the `"consumer side"` is updated it can communicate directly to the correct function.

slaweet: Ah, ok. Got it. 
Good to hear from you @massao, BTW 🙂

libs/hwManager/manufacturers/ledger.js
```diff
@@ -34,13 +68,25 @@ const listener = (transport, actions) => {
   }
 };
 
+/**
+ * geteLedgerAccount - function - Check if account exist for selected coin
+ * @param {number} index - indeex for the desire account, if not value is provide use default (0)
```
slaweet: Typos
```suggestion
 * @param {number} index - index for the desired account, if no value is provided use default (0)
```

libs/hwManager/manufacturers/ledger.js
```diff
@@ -34,13 +68,25 @@ const listener = (transport, actions) => {
...
+ */
 const getLedgerAccount = (index = 0) => {
   const ledgerAccount = new LedgerAccount();
   ledgerAccount.coinIndex(SupportedCoin.LISK);
   ledgerAccount.account(index);
   return ledgerAccount;
 };
 
+/**
+ * checkIfInsideLiskApp - function - Validate if after use the pin to unblock ldeger device
```
slaweet: Typo
```suggestion
 * checkIfInsideLiskApp - function - Validate if after using the pin to unblock ledger device
```

libs/hwManager/manufacturers/ledger.js
```diff
@@ -1,31 +1,65 @@
...
+    .forEach(device => remove(device.path));
+  devices = devices.filter(device => connectedPaths.includes(device.path));
 };
 
+/**
+ * listener - function - Always listen for new messages for connect or disconnect devices.
+ * @param {object} transport - Library use for handle the ledger devices.
+ * @param {object} actions - Contains 2 functions: add and remove.
+ * @param {function} actions.add - Function for add a new device to the main list.
+ * @param {function} actions.remove - Function for remove a device to the main list.
```
slaweet:
```suggestion
 * @param {function} actions.remove - Function for remove a device from the main list.
```

libs/hwManager/manufacturers/ledger.js
```diff
@@ -1,31 +1,65 @@
...
 
+// ============================================ //
+//              DEVICES LIST
+// ============================================ //
 let devices = [];
-const clearDevices = async (transport, { remove }) => {
+
+/**
+ * addDevice - function - Add a new device to the devices list.
+ * @param {object} device - Device object comming from the ledeger library
```
slaweet: Typos
```suggestion
 * @param {object} device - Device object coming from the ledger library
```

libs/hwManager/manufacturers/trezor.js
```diff
@@ -1,39 +1,58 @@
...
 
-  process.on('exit', () => {
-    transport.onbeforeunload();
-  });
+/**
+ * listener - function - Always listen for new messages for connect or disconnect devices.
+ * @param {object} transport - Library use for handle the trezor devices.
+ * @param {object} actions - Contains 2 functions: add and remove.
+ * @param {function} actions.add - Function for add a new device to the main list.
+ * @param {function} actions.remove - Function for remove a device to the main list.
```
slaweet:
```suggestion
 * @param {function} actions.remove - Function for remove a device from the main list.
```

app/src/main.js
```diff
@@ -12,8 +12,8 @@ import updateChecker from './modules/autoUpdater';
 import server from '../server';
 
 require('babel-polyfill'); // eslint-disable-line import/no-extraneous-dependencies
```
yasharAyari: Please user ES6 import instead of `require`.

app/src/main.js
```diff
@@ -12,8 +12,8 @@ import updateChecker from './modules/autoUpdater';
...
-require('./ledger');
-// require('./modules/hwManager');
+// require('./ledger');
+require('./modules/hwManager');
```
yasharAyari: same here.

libs/hwManager/manufacturers/index.js
```diff
@@ -1,7 +1,7 @@
-import ledger from './ledger';
-import trezor from './trezor';
+import Ledger from './ledger';
```
yasharAyari: You don't need this file. Whenever you need to import both files in another file just import the whole folder like this 
`import * as libs from 'libs/hwManager/manufacturers'`.

libs/hwManager/manufacturers/trezor.js
osvaldovega: Agree, I made a few changes to work properly but yes this was need it.

libs/hwManager/manufacturers/ledger.js
```diff
@@ -34,13 +68,25 @@ const listener = (transport, actions) => {
   }
 };
 
+/**
+ * geteLedgerAccount - function - Check if account exist for selected coin
```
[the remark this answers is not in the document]
osvaldovega: done

libs/hwManager/utils/utils.js
```diff
@@ -45,8 +47,57 @@ export const subscribe = (receiver, { event, action }) => {
   return true;
 };
 
+// ===================================================== //
+//                   LEDGER UTILS
```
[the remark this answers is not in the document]
osvaldovega: agree, I just made the update in the file structure for hwManager

libs/hwManager/utils/constants.js
```diff
@@ -0,0 +1,25 @@
+
```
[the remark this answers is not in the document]
osvaldovega: done

libs/hwManager/manufacturers/ledger.js
osvaldovega: done

libs/hwManager/manufacturers/ledger.js
osvaldovega: done

libs/hwManager/manufacturers/ledger.js
osvaldovega: done

libs/hwManager/manufacturers/ledger.js
osvaldovega: done

libs/hwManager/manufacturers/trezor.js
osvaldovega: done

app/src/main.js
osvaldovega: thanks for pointing this Yashar, I was doing some tests and I can confirm that is totally necessary, however I remove it from this file and put it in the correct place.

app/src/main.js
osvaldovega: done

libs/hwManager/manufacturers/index.js
osvaldovega: base on the new structure we need it, so at least for now is better in the way it is, in case this will change we can use this solution, thx anyway

osvaldovega:
I just created the issue related the problem in the send component, we need to address that as soon as possible, the issue # with the description is #2414

libs/hwManager/constants.js
```diff
@@ -0,0 +1,31 @@
...
+      Ledeger_Nano_X: 'Ledeger Nano X',
+    },
+  },
+  Trezor: {
+    name: 'Trezor',
+    models: {
+      Trezor_Model_T: 'Trezor Model T',
+      Trezor_Model_One: 'Trezor Model One',
+    },
+  },
```
slaweet: By the architecture of putting manufacturer-specific code to each manufacturer, `MANUFACTURERS` should be IMO moved the the `ledger` folder for ledger and `trezor` folder for trezor. 

The only question is if there should be a `libs/hwManager/manufacturers/ledger/constants.js` file or if it can be part of the object in `libs/hwManager/manufacturers/ledger/index.js`

libs/hwManager/manufacturers/trezor/utils.js
```diff
@@ -0,0 +1,40 @@
...
+    fee: parseInt(fee, 10),
+    recipient_id: recipientId,
+    sender_public_key: senderPublicKey,
+    requester_public_key: requesterPublicKey,
+    ...rest,
+  };
+
+  Object.keys(transaction).forEach((key) => {
+    if (transaction[key] === 0) return key;
+    return (!transaction[key]) && delete transaction[key];
```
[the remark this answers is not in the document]
osvaldovega: my bad, I fixed

libs/hwManager/constants.js
osvaldovega: thinking about this I will say easy way is just use a object in each index file for trezor and ledger but for consistency and thinking in more constants in future I add a constants file for each manufacture, looks more organized

PR #2430, opened 2019-09-10
Fix send with hw wallet after one tx was rejected - Closes #2414

### LINKED ISSUE #2414: Application don't wait for user to accept/reject the tx for second time in the Send component
### Problem

The send component is not using the same structure as delegate (voting) and because of this actual logic, if any send tx fail, or is cancel in case of hw wallet, make that the second time user try to submit a tx this skip the summary step and go directly to the result page with error/fail status, so the user does not have time to accept or cancel the tx.

The problem is related with the transactions reducer, once any tx fail or be cancel, this is being store in the reducer of transactions in the **broadcastedTransactionsError** array, so when user try a second time this array still has the error message of the previous tx and because of this and the logic that is implemented in the send-summary component cause that problem.

### Steps to reproduce.

**With Hardware Wallet.**

1. Login with any hw device, trezor or ledger
2. Go to send.
3. fill the form
4. do click in the button to proceed.
5. using the hw dvice, cancel the tx.
6. transaction is cancel (what it is ok).
7. try to send any amount one more time.
8. after you fill the information in the form and do click you will see the problem.

### Possible Solution

Use the same logic as voting. What means remove the action/reducer related to this part and use the HOC that voting use.

slaweet:
### What issue have I solved?

#2414

The problem was that the redux store wasn't cleared, so I clear it when leaving the error page.

### How has this been tested?

Follow steps in #2414

### Review checklist

### DIFF
--- a/src/components/send/transactionStatus/transactionStatus.js
+++ b/src/components/send/transactionStatus/transactionStatus.js
@@ -20,6 +20,10 @@ class TransactionStatus extends React.Component {
     this.onRetry = this.onRetry.bind(this);
   }
 
+  componentWillUnmount() {
+    this.props.resetTransactionResult();
+  }
+
   componentDidMount() {
     const { recipientAccount, fields } = this.props;
     recipientAccount.loadData({ address: fields.recipient.address });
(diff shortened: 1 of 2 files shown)


Efefefef:
Thank you

Issue #2409, opened 2019-09-02
Enable users to send the entire balance of their account


reyraa:
This is more of a UX question:

IF you copy your balance and paste it in the amount input in Send form, you'll see an error message for not having enough balance.
Seems like we should provide a way for users to easily transfer all funds.
I've seen a few solutions:
 - We can have a button labeled _Transfer all funds_ which filled balance minus fee in the mount field.
 - We can change the Send form to allow us input the value and then it shows _your recipient will receive: $$$ LSK/BTC_. in which $$$ is equal the value you entered minus fee. 

Which one o you prefer? do you have any solution in mind?

PR #2433, opened 2019-09-11
:seedling: Add "Send entire balance" button

slaweet:
### What issue have I solved?

#2409

Work in progress

### How has this been tested?

- Go to send
- Click the "Send entire balance"

Do it for LSK, BTC, German, empty account.

### Review checklist


slaweet:
> 🐛 Input invalid value -> Send all -> Validation error still present

Fixed

> We don't have anything about send all for BTC in the story or in designs. If we want it then it's not clear how to calculate the right amount given that we can choose the fee

I think this was discussed in the grooming meeting when estimating this story. It should set the amount  based on the currently selected fee (High/Low) and don't worry about the user changing the fee later.

PR #2435, opened 2019-09-11
Modals to promote the anonymous analytics plan - closes #2408

### LINKED ISSUE #2408: Modals to promote the anonymous analytics plan
We have an anonymous analytics plan which helps us get some statistics regarding usage and bugs. Meaning that, using [Matomo](https://matomo.org/), we can get the number of people click on certain button or visiting certain page, as well as details of the bugs Lisk Hub throws.

Since Lisk Hub is a wallet, we can have this feature enabled by default. There is an option to enable it in settings, but it's not prominent and thus only <5% of our users have enabled it.

We can show a dialog to promote this feature. It should be shown after then user navigates to the Wallet page for the first time.
If the user neglects the banner, it will be visible at all times. 
Every time the user discards the banner, it'll show up again afetr 1 week. 
The banner will never show up again if the user has opted in.
The banner shows up again, when the user opts out of the plan.

We can also show this when an error happens. We can show the detail of the error and ask the user to help us improve the app by reporting a. this bug, b. all bugs.

We should make it clear that a. this is centralized and not over blockchain. b. it's anonymous and c. it has nothing to do with their sensitive information.

There designs are ready on [InVision](https://projects.invisionapp.com/d/main#/console/18170373/378557455/preview).

### Acceptance criteria
 - Users should be notified about anonymous analytics plan without blocking their normal actions.
 - Users should be able to see more details about anonymous analytics plan.
- Users must be recommended to opt in to the plan when an error occurs.

osvaldovega:
### What issue have I solved?

-- #2408 

This PR includes the implementation of the following things:

- [x] FlashMessage for Analytics.
- [x] Dialog Container for Analytics.
- [x] Analytics util file for manage the interaction.
- [x] Unit test for each of these files

### How has this been tested?

1. Run the application
2. After the app is running the banner will show up
3. If do click in Read More will open a dialog.
4. If in the dialog do click in cancel, this will show the dialog and falshMessage and then will show up after x amount of minutes.
5. If do click in accept will save the changes and will never show up again only if user disabled the monitoring again.

### New dependencies
- jest-date-mock: https://github.com/hustcc/jest-date-mock/blob/master/LICENSE

### Review checklist


Efefefef:
@reyraa  The logic is absent from the story. 

1. Can you clarify the exact conditions for modal to appear (place, time, actions, conditions)? 
1. What is X amount of minutes exactly? How exactly timeout works? When the timeout restarts? Is there a way to get rid of the modal forever?

@osvaldovega 
1. 🐛  Margins, button size, textarea size are not aligned with design
1. It is not clear what you describe in How to test #4. There is no dialog or message when you click in Cancel
1. Where I can modify timeout in order to test it

osvaldovega:
thx @Efefefef I just made the fixes to the css to looks as the design, I used pixel perfect and now looks almost like the design with a little different of the letters but that is because how it renders the text.

about your questions, let me explain better how this works.

After user install the new version 3 things will happen.

1.  First time.
Well you run the app the flash message banner will NOT show up unless you **SignIn** and then go to the **Wallet** page, this is the ONLY way to see the banner for first time, be in the wallet page.

2. Once you be in Wallet page the banner will remain open in ALL the pages until user do click in Read More and take any action. In other words, this banner will always be visible unless the user do click in cancel or accept in the dialog.
So if you don't do nothing just ignore the banner, next time you run the app this will show up as soon the app is up.

3. If the user do a click in Read More link and after the banner show up the user do click in CANCEL this will close the dialog and close the banner, and after 7 days will show up again, so that means that after do click in cancel the app will remaining you 7 days later, this is when you will see the banner again. You can adjust the computer date to trigger this scenario after 7 days.

Hope this be more useful for you for test.

reyraa:
According to the ticket

If the user neglects the banner, it will be visible at all times.
Every time the user discards the banner, it'll show up again after 1 week.
The banner will never show up again if the user has opted in.
The banner shows up again, when the user opts out of the plan.

PR #2464, opened 2019-09-17
Finalize hw manager - Closes #2446

### LINKED ISSUE #1954: Create a refactor proposal for logic and structure of hwWallet in electron and react

### LINKED ISSUE #2446: Clean-up and finalize hwManager refactoring
This is the last subtask of #1954 

The goal is to delete all unnecessary code and merge the feature branch `1954-hardware-wallet-manager` into `development,

Files to delete:
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/utils/hwWallet.js
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/utils/api/hwWallet.js
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/utils/ledger.js
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/store/reducers/hwWallets.test.js
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/store/reducers/hwWallets.js
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/actions/hwWallets.test.js
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/actions/hwWallets.js

Files to check for what can be removed:
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/constants/hwConstants.js
- https://github.com/LiskHQ/lisk-hub/blob/75ee20a2d923da35cdf40ee6c446b2fee1dc5569/src/store/middlewares/hwWallet.js

Other clean-up that might involve other files:
- remove all uses of `isHarwareWalletConnected` (i.e.: `git grep  isHarwareWalletConnected` should return empty)

slaweet:
### What issue have I solved?

Fixes #2446 
Fixes #1954

- [x] 🔥 Remove hwWallets reducer
- [x] 🔥 Remove isHarwareWalletConnected from settings store
- [x] Delete other obsolete files
- [x] Remove unncessary code from existing hwManager files
- [x] :bug: Fix voting cancelation

Note that this branch contains all changes from the whole epic. To see just changes new in this PR, see https://github.com/LiskHQ/lisk-hub/compare/1954-hardware-wallet-manager...2446-finalize-hw-manager

### How has this been tested?

Check for regressions in supported hardware wallets:
 - connect (also without open app for Ledger)
 - login
- send (including cancel and retry)
- vote (including cancel)
- disconnect

Check all of the above with:
- Ledger Nano S
- Ledger Nano X
- Trezor Model T

### Review checklist


src/store/middlewares/hwManager.js
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * subscribeToDeviceDisonnceted - Function -> To detect any hw wallet disconnection
+     * @param {fn} function - callback function to execute toast dispatch after receive the data
+     * and in case user is SignIn trigger the logout Dialog and toast message.
+     */
+    subscribeToDeviceDisonnceted((response) => {
+      const { account, settings } = store.getState();
+      const activeToken = settings.token.active || 'LSK';
+
+      // Check if user is SignedIns
```
osvaldovega: I think this was my bad, but there is a typo in here SignedIn without s

src/utils/hwManager.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+  return accounts;
+};
+
+/**
+ * signSendTransaction - Function.
+ * This function is used for sign a send transaction.
+ */
+// eslint-disable-next-line max-statements
```
osvaldovega: This line can be remove

libs/hwManager/utils.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Subscribe to a event throught the receiver
+ * @param {any} receiver - Subscriber that should be subscribed to the the event
+ * @param {object} data - Object containing the event and the action
+ * @param {string} data.event - Event name to be subscribed to thoguht the receiver
+ * @param {function} data.action - Function that should be executed when event is triggered
+ */
+export const subscribe = (receiver, { event, action }) => {
+  if (!receiver) return false;
```
yasharAyari: Please use `?` operator instead of the `if`. Then you can have one return.

slaweet: @yasharAyari What do you mean exactly? 
I can do something like this, but it IMO doesn't look very readable:
```js
export const subscribe = (receiver, { event, action }) => {
    return !receiver ? false
      : createCommand(receiver, {
        command: event,
        fn: action,
      }) || true;
  };
```

libs/hwManager/manufacturers/trezor/index.js
```diff
@@ -0,0 +1,106 @@
...
+        );
+        return resolve(message.public_key);
+      } catch (err) {
+        return reject();
+      }
+    });
+  });
+};
+
+const signTransaction = async (transporter, { device, data }) => {
```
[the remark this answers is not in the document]
slaweet: Fixed

libs/hwManager/utils.js
yasharAyari: I meant some thing like below. It fulfils your need and I think it is readable enough because it is just a simple condition.
```
const subscribe = (receiver, { event, action }) => !receiver ? receiver
      : createCommand(receiver, {
        command: event,
        fn: action,
      });
```

slaweet: Ok. change pushed. Please check again @yasharAyari

PR #2477, opened 2019-09-19
Solve warnings in console after upgrading React - Closes #2476

### LINKED ISSUE #2476: Solve warnings in console after upgrading React
### Expected behaviour
When I go to http://localhost:8080/ and open the browser console, there are no errors or warnings
### Current behaviour
When I go to http://localhost:8080/ and open the browser console, there are many warnings and some errors:
Screenshot 2019-09-19 at 15 34 29

slaweet:
### What issue have I solved?

#2476

To prevent many variations of `Warning: componentWillReceiveProps has been renamed`:
- [x] Update react-redux@7.1.1 and update code and tests where needed
- [x] Update react-i18next@10.12.5 i18next@17.0.16 and update code and tests where needed
- [x] Update enzyme@3.10.0 enzyme-adapter-react-16@1.14.0 and update code and tests where needed
- [x] Fix other broken unit tests
- [x] Fix other warnings coming from our components

Note that there are still warnings coming from [react-toolbox](https://www.npmjs.com/package/react-toolbox) but this dependency doesn't have a newer version, so we need to decide if we can get rid of it and should be done in a separate ticket. React-toolbox is causing warnings like: `Please update the following components:  ThemedRippledComponent, ThemedTBButton, ThemedThemed, ThemedProgressBar`

### How has this been tested?

Open the console on http://localhost:8080/#/dashboard and check it's empty

### Review checklist


slaweet:
> I still see the following error in the console when I open it from /dashboard:
> 
> ```
> Warning: DropdownButton: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)
> ```

Good catch, I had that commit but forgot to push it 😊 Now pushed: https://github.com/LiskHQ/lisk-hub/pull/2477/commits/e6f459695babddd6beb8390fd1a9c9dd866c7ae8

PR #2478, opened 2019-09-20
Implement login with trezor one - closes #1844

### LINKED ISSUE #1844: Implement login with Trezor One
This is a subtask of #1319. 
In this task, it should not be available to end-users yet (e.g. use a feature branch). There will be a final task to enable it.

The goal is to implement login with Trezor One. 
Design: https://projects.invisionapp.com/d/main#/console/17570736/369060560/preview

### LINKED ISSUE #1884: Implement vote with Trezor One
This is a subtask of #1319. 
In this task, it should not be available to end-users yet (e.g. use a feature branch). There will be a final task to enable it.

The goal is to implement voting with Trezor One. That means voting in https://ci.lisk.io/test/lisk-hub/development/#/delegates should work with Trezor One exactly the same as with other hardware wallets.

### LINKED ISSUE #1885: Implement send LSK with Trezor One
This is a subtask of #1319. 
In this task, it should not be available to end-users yet (e.g. use a feature branch). There will be a final task to enable it.

The goal is to implement send LSK with Trezor One. That means voting in https://ci.lisk.io/test/lisk-hub/development/#/wallet/send should work with Trezor One exactly the same as with other hardware wallets.

### LINKED ISSUE #1886: Enable Trezor One to end users
This is one of the two final subtasks of #1319. 

The goal is to enable Trezor One to end users, as in the previous tasks it was only available after setting a local storage flag.

osvaldovega:
### What issue have I solved?

-- #1844 

Also:
- closes #1884 
- closes #1885
- closes #1886

This PR include the implementation of the PIN page for the Trezor Model One, this implementation is ONLY for unlock the device for FIRST time after that device is unlock (enter the correct PIN), all other functions can be handle by device.

Implementation:
- [x] Add listener in HWManager for trezor.
- [x] Add a function to listen from trezor library for PIN on success or error in HWManager/index.js
- [x] Import and export function for requestPIN in HWManager/communication.js file
- [x] Implementation of the RequestPin component with logic
- [x] create a new input component called MaskInput for be use and interact in the same way as design.

#### How it works?
The trezor one has a particular way to unlock the device to be use properly, after the user connect the device to the PC the **ONLY** way to get any information from the device is introducing the secret PIN (same as trezor model T).

So can I get the numeric grid, in the device display to see the numbers and later use that grid in the application?
After the device is connected to the PC you just need to call the function **getPublickey**, this will trigger the grid in the device display and if the PIN is correct will return the correct public key, if the PIN is incorrect then will not return the public key.

So the interaction will be.

1. Call get public key.
2. Grid is display in the device screen.
3. Use the app grid to introduce the PIN according to the device grid order.
4. Submit the PIN.
5. The check PIN function then will be waiting to get the PIN value.
6. If correct the get public key function will resolve, if not will reject
7. In any case if PIN is valid or not, the grid will be disappear.
8. For trigger again the grid you need to call the get public key function again 

### How has this been tested?

1. Run the app.
2. Connect the trezor model T and model One.
3. In the splashscreen select the login with HW.
4. Select Trezor model T.
5. If the device is unlock you will see the loading accounts.
6. Do click in go back.
7. Now do the same steps and select trezor model One.
8. The Request PIN should be display.
9. Introduce the PIN and do click in submit.
10. Now the select account page should be display.
11. If do click in back button and try to login again with Model One the loading accounts should be display directly without as for PIN

**NOTE:** Send and Voting was already tested with the Trezor Model One and it is working properly.

### Review checklist


osvaldovega:
rebase with base branch to merge with development in the future

slaweet:
I went through the code and pushed some refactor commits (see commits for more details).

I couldn't test with the device as I don't have one with me. @colado please test it with the Trezor One device.

src/components/hwWalletLogin/requestPin.js
```diff
@@ -0,0 +1,125 @@
...
+              </div>
+            </div>
+
+            <div className={styles.buttonsContainer}>
+              <PrimaryButton
+                className="primary-btn"
+                onClick={this.onSubmitPin}
+                disabled={error}
+              >
+                {t('Enter PIN')}
```
yasharAyari: I think it is better to use 'unlock' as the label of the button.

src/components/hwWalletLogin/unlockDevice.js
```diff
@@ -71,7 +70,7 @@ class UnlockDevice extends React.Component {
           </a>
         </p>
         <Illustration name="ledgerNanoLight" />
-        <TertiaryButton onClick={() => { history.push(routes.splashscreen.path); }}>
+        <TertiaryButton onClick={goBack}>
```
yasharAyari: `goBack` button doesn't work.

slaweet:
All your comments implemented @yasharAyari. Please check again.

PR #2492, opened 2019-09-25
Fix i18n in electron app -  Closes #2486 #2490 #2487 #2489

slaweet:
### What issue have I solved?

Fixes #2486
Fixes #2490
Fixes #2487
Fixes #2489

### How have I implemented/fixed it?

I updated what I forgot in https://github.com/LiskHQ/lisk-hub/commit/e57a6f2c233c6b6683261ba73edd2bef530b6937 

The change was because i18next was upgraded and the new version init returns a promise.

The warning described in #2486 was causing all the bugs linked above.

Not that the PR also contains https://github.com/LiskHQ/lisk-hub/pull/2492/commits/5acad2ea57b8fde1de17c814330dcc6df622dcdf which is not essential to fix the bugs but it helped with debugging and thus is included as it will be useful in future debugging as well.

### How has this been tested?

Follow the steps from the tickets above.

### Review checklist
- The PR follows our [Test guide](/LiskHQ/lisk-hub/blob/development/docs/TEST_GUIDE.md)
- The PR follows our [CSS guide](/LiskHQ/lisk-hub/blob/development/docs/CSS_GUIDE.md)

PR #2502, opened 2019-09-27
Fix updating transaction list with the first one - Closes #2488

### LINKED ISSUE #2488: HW: Transaction list doesn't update on receiving the first one
### Steps to reproduce

1. Login with HW as untouched acc with empty state transaction list
1. Send some LSK to it
1. Check out the transaction list

### Actual result

Transaction list is not updated

### Expected result

Tx list displays the first transaction

slaweet:
### What issue have I solved?

Fixes #2488

I simplified the logic for updating the transaction list on a new block to use the same condition for transactions as was used for the account, because account balance was updated correctly.

Note that https://github.com/LiskHQ/lisk-hub/blob/de4443ec01e545bf1757c95de624ba0e58bb572d/src/actions/account.js#L106 was just a leftover from Lisk Nano, where we had confirmations in transaction list, so we had to update the list just because of confirmations.

### How has this been tested?

Follow steps in #2488

### Review checklist

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -8,7 +8,6 @@ import { getConnectionErrorMessage } from './network/lsk';
 import { getTimeOffset } from '../utils/hacks';
 import { loginType } from '../constants/hwConstants';
 import { networkStatusUpdated } from './network';
-import { updateTransactions } from './transactions';
 import accountConfig from '../constants/account';
 import actionTypes from '../constants/actions';
 import { tokenMap } from '../constants/tokens';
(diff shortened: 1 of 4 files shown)


slaweet:
> Code looks good but I'm having problems with the Electron app - it doesn't recognise any hardware wallet for me. @Efefefef can you check if it works for you?

Sergio, that problem is being fixed in another PR: https://github.com/LiskHQ/lisk-hub/pull/2492 

On this PR you can log in if you have the hw wallet connected before you start the app, but there is a bunch of other problems even then.

PR #2521, opened 2019-10-01
Generalize filter component on transactions list - Closes #2508

### LINKED ISSUE #2508: Generalize filter component on transactions list
We need to make the existing code of the filters generic so it can be used by other screens such as list of latest [blocks](https://github.com/LiskHQ/lisk-hub/issues/2455) and [transactions](https://github.com/LiskHQ/lisk-hub/issues/2456).

slaweet:
### What issue have I solved?

Closes #2508

- [x] Move `FilterBar` component from `screens/transactions/...` to `components/shared/...`
- [x] Move `FilterContainer` component from  `screens/transactions/...` to `components/shared/...` and rename it to `FilterDropdownButton`
- [x] Replace `MessageFieldGroup` component with a generic `TextFilter`
- [x] Separate `DateField` component from `DateFieldGroup`
- [x] Make `FilterDropdownButton` independant of the data structure used in transactions table

### How has this been tested?

- Check filtering on wallet (http://localhost:8080/#/wallet) and account page (http://localhost:8080/#/explorer/accounts/7955155501030618852L) for regressions

### Review checklist


src/components/shared/filterDropdownButton/dateFieldGroup.js
```diff
@@ -0,0 +1,163 @@
...
+
+      return {
+        ...acc,
+        [field]: {
+          value,
+          error,
+          loading: false,
+        },
+      };
+    }, {});
```
osvaldovega: Could be possible to refactor this part of the function to more the eslint ignore flag?
and maybe you don't need to set the error in every if you can do something like this.

Instead of 

```
feedback = t(`Date must be in ${this.dateFormat} format`);
error = true;
```

you can set the error based on the feedback

```
return {
        ...acc,
        [field]: {
          value,
          error: !!feedback,
          loading: false,
        },
      };
    }, {});
```

src/components/shared/filterDropdownButton/dateFieldGroup.js
```diff
@@ -0,0 +1,163 @@
...
+    };
+
+    clearTimeout(this.timeout);
+    this.timeout = setTimeout(() => {
+      this.validateDates(fields, selectionObj);
+    }, 300);
+
+    this.setState({ fields });
+    this.props.updateCustomFilters(fields);
+  }
```
osvaldovega: This function too refactor a little bit to remove the eslint flag

PR #2522, opened 2019-10-01
Implement the details of the block on a dedicated page - closes #2505

### LINKED ISSUE #2505: Implement the details of the block on a dedicated page
### Description
As a user; I'd like to see detailed information of any given block within a dedicated, standalone page.

### Acceptance criteria  
- Matching to [Designs](https://projects.invisionapp.com/share/ZRU30GY5NTX#/screens/385494275_Discovery_-_Blocks_Details_-_Filters_Opened)
- I should be able to copy the block ID and height
- I should be able to navigate to the delegate page who forged the block.
- Transactions list is not included in this story.

osvaldovega:
### What issue have I solved?

-- #2505 

This PR include the page to see the details about any block in the network.

- [x] Add the new route to the router file.
- [x] Create the new component.
- [x] Implement the Block details table.

This PR is only to see the first table in the Block Details page NOT the transaction list.

### How has this been tested?

To test this you need to use Mainnet or Testnet, Devnet wont work.

1. Run the app.
2. Select Testnet network.
3. In the URL write <localhost_url>/monitor/blocks
4. Then you will see the list of all blocks.
5. Select any block by doing click in the block ID.
6. The Block Details page will show up.
7. Now you can see all the block details in this page. (ONLY block details NO transaction list).

### Review checklist


src/utils/api/lsk/liskService.js
```diff
@@ -1,29 +1,63 @@
...
+    transformResponse: response => response.data,
+    searchParams: {
+      limit: 20,
+      ...searchParams,
+    },
+  }),
+
+  getBlockDetails: ({ networkConfig }, searchParams) => liskServiceGet({
+    serverUrl: getServerUrl(networkConfig),
+    path: '/api/v1/block/',
```
slaweet:
```suggestion
    path: '/api/v1/block/${id}',
```
I would put the id into the path here instead of adding additional complexity to `liskServiceGet`

src/utils/api/lsk/liskService.js
```diff
@@ -1,29 +1,63 @@
...
+    searchParams: { ...searchParams },
```
slaweet:
```suggestion
```
There's no further `searchParams` that this endpoint would accept

src/utils/api/lsk/liskService.js
```diff
@@ -1,29 +1,63 @@
...
+    serverUrl: getServerUrl(networkConfig),
+    path: '/api/v1/blocks/last',
+    transformResponse: response => response.data,
+    searchParams: {
+      limit: 20,
+      ...searchParams,
+    },
+  }),
+
+  getBlockDetails: ({ networkConfig }, searchParams) => liskServiceGet({
```
slaweet:
```suggestion
  getBlockDetails: ({ networkConfig }, { id }) => liskServiceGet({
```

src/components/screens/monitor/blockDetails/blockDetails.js
```diff
@@ -0,0 +1,119 @@
...
+                    <div className={styles.dataContainer}>
+                      <label>{t('Date')}</label>
+                      <span>
+                        <DateTimeFromTimestamp time={blockDetailsData.timestamp * 1000} token="BTC" />
+                      </span>
+                    </div>
+
+                    <div className={styles.dataContainer}>
+                      <label>{t('Generated by')}</label>
+                      <span className={styles.clickable}>
```
slaweet:
```suggestion
                      <span>
```
Not needed

src/components/screens/monitor/blockDetails/blockDetails.css
```diff
@@ -0,0 +1,32 @@
...
+    @mixin contentNormal bold;
+  }
+
+  & > span {
+    @mixin contentLargest;
+
+    margin-top: 13px;
+  }
+
+  & .clickable {
```
slaweet: this class is not needed as Link (a) already has cursor pointer.

src/components/screens/monitor/blockDetails/blockDetails.js
```diff
@@ -0,0 +1,119 @@
...
+          blockDetails.error
+            ? (
+              <Box.Content>
+                <Feedback status="error" show>{t('Failed to load block details.')}</Feedback>
+              </Box.Content>
+            )
+            : (
+              <React.Fragment>
+                <div>
+                  <TableRow className={styles.tableRow}>
```
slaweet: There is `Box.Row` component for which you don't need to override `hover` style. See Box demo: https://github.com/LiskHQ/lisk-hub/blob/dd73b2379df1e96828fdadb2097948b1bcfdcd19/src/components/toolbox/box/demo.js#L44

```suggestion
                  <Box.Row>
```

src/components/screens/monitor/blockDetails/index.js
```diff
@@ -0,0 +1,28 @@
...
+  id: ownProps.match.params.id,
+});
+
+export default compose(
+  withRouter,
+  connect(mapStateToProps),
+  withData({
+    blockDetails: {
+      apiUtil: liskService.getBlockDetails,
+      defaultData: [],
```
slaweet:
```suggestion
```
IMO `defaultData` for block should be `{}` (just update liskService.getBlockDetails), 
`{}` is the default in https://github.com/LiskHQ/lisk-hub/blob/dd73b2379df1e96828fdadb2097948b1bcfdcd19/src/utils/withData.js#L14 so it's not needed here.

src/utils/api/lsk/liskService.js
```diff
@@ -1,29 +1,63 @@
...
+    serverUrl: getServerUrl(networkConfig),
```
slaweet:
```suggestion
    serverUrl: getServerUrl(networkConfig),
    transformResponse: response => response.data[0],
```
I think you always want the first block, so transforming it here will save you extra logic in https://github.com/LiskHQ/lisk-hub/blob/09bba34631fe65bd7f0bb71aaae44a6ed1cc3d6e/src/components/screens/monitor/blockDetails/blockDetails.js#L14

src/components/screens/monitor/blockDetails/blockDetails.css
```diff
@@ -0,0 +1,32 @@
+@import '../../../../app/variables.css';
```
slaweet:
```suggestion
```
This import is not needed, you don't use any variables in this file

src/components/screens/monitor/blockDetails/blockDetails.css
```diff
@@ -0,0 +1,32 @@
...
+  align-items: flex-start;
+  flex-grow: 1;
+  box-sizing: border-box;
+  width: calc(100% / 5);
+
+  & > label {
+    @mixin contentNormal bold;
+  }
+
+  & > span {
```
slaweet: We agreed not to use cascading css for generic elements like `div` or `span` https://trello.com/c/J5BXlkpx

In this case, you could just move the margin to the `label` and the mixin to the `dataContainer`

osvaldovega:
Thx @slaweet for the feedback, I just made the changes

osvaldovega:
**NOTE:** I spoke with Julian and Dom and they mentioned both this page it is not finish yet, so the implementation about design will change. In other words the design it is not final.
So at this point we can see that at least that data is coming properly from the endpoint.

PR #2526, opened 2019-10-02
Implement filtering blocks - Closes #2509

slaweet:
### What issue have I solved?

Closes #2509

### How have I implemented/fixed it?

- [x] Implement the block filter dropdown component
- [x] Create useFilters hook to make the state reusable
- [x] Update the FilterBar for use on blocks page

Note that there is one edge-case bug in the API that should be fixed on the backend: https://github.com/LiskHQ/lisk-service/issues/312 

### How has this been tested?

- Go to blocks page: https://jenkins.lisk.io/test/lisk-hub/PR-2526/#/monitor/blocks
- Try filtering by various options

### Review checklist
- The PR follows our [Test guide](/LiskHQ/lisk-hub/blob/development/docs/TEST_GUIDE.md)
- The PR follows our [CSS guide](/LiskHQ/lisk-hub/blob/development/docs/CSS_GUIDE.md)

PR #2540, opened 2019-10-08
Implement the filters and sorting on the latest transactions - Closes #2512

### LINKED ISSUE #2512: Implement the filters and sorting on the latest transactions

colado:
### What issue have I solved?

#2512

- [ ] Add filter dropdown.
- [ ] Make filter component expandable.
- [ ] Apply filter functionality.
- [ ] Apply sorting functionality.

### How has this been tested?
WIP
[design](https://projects.invisionapp.com/d/main?origin=v7#/console/18539694/386607815/preview )

### Review checklist

### DIFF
--- a/src/components/screens/monitor/transactions/transactions.js
+++ b/src/components/screens/monitor/transactions/transactions.js
@@ -5,6 +5,40 @@ import TransactionsTable from '../../../shared/transactionsTable';
 import PageLayout from '../../../toolbox/pageLayout';
 
 const Transactions = ({ t, transactions }) => {
+  const filters = [{
+    label: t('Date'),
+    name: 'date',
+    type: 'date-range',
+  }, {
+    label: t('Amount Range'),
+    name: 'amount',
+    type: 'number-range',
+  }, {
+    label: t('Sender'),
+    name: 'sender',
+    type: 'text',
+  }, {
+    label: t('Message'),
+    placeholder: t('Your message'),
+    name: 'message',
+    type: 'text',
+  }, {
+    label: t('Recipient'),
+    placeholder: t('Address or Public key'),
+    name: 'message',
+    type: 'text',
+  }, {
+    label: t('Type'),
+    placeholder: t('Eg. 1,5'),
+    name: 'type',
+    type: 'text',
+  }, {
+    label: t('Height'),
+    placeholder: t('Eg. 10180477'),
+    name: 'height',
+    type: 'text',
+  }];
+
   const columns = [
     { header: t('Sender'), className: `${grid['col-xs-3']}`, key: 'senderId' },
     { header: t('Recipient'), className: `${grid['col-xs-3']}`, key: 'recipientId' },
@@ -20,7 +54,7 @@ const Transactions = ({ t, transactions }) => {
 
   return (
     <PageLayout>
-      <TransactionsTable isLoadMoreEnabled columns={columns} title={t('All transactions')} transactions={transactions} />
+      <TransactionsTable isLoadMoreEnabled filters={filters} isFilterDropdownExpandable columns={columns} title={t('All transactions')} transactions={transactions} />
     </PageLayout>
   );
 };


src/components/shared/transactionsTable/index.js
```diff
@@ -143,14 +144,15 @@ class TransactionsTable extends React.Component {
 
   render() {
     const {
-      title, transactions, columns, isLoadMoreEnabled, t,
+      title, transactions, columns, isLoadMoreEnabled, t, filters, isFilterDropdownExpandable,
```
slaweet: I think `isFilterDropdownExpandable` is not needed to be passed from outside. IMO `isFilterDropdownExpandable = filters.length > 3` and is something that can be only inside the shared https://github.com/LiskHQ/lisk-hub/blob/development/src/components/shared/filterDropdownButton/index.js

src/constants/transactionTypes.js
```diff
@@ -5,6 +5,13 @@ export default {
   vote: 3,
 };
 
+export const transactionNames = {
+  0: 'Send',
+  1: 'Second passphrase registration',
+  2: 'Delegate registration',
+  3: 'Delegate vote',
```
slaweet: This is duplicating transaction names in https://github.com/LiskHQ/lisk-hub/blob/baae7626cd59f4105c739ebf19fd357fabe8f9cb/src/components/screens/wallet/transactions/address/TransactionAddress.js#L16-L33

It would be ideal to move that component to `/shared` and reuse it.

colado: I think it definitely makes sense to move `TransactionAddress` to `shared`, since it is being used both in `dashboard` and `wallet`. However those styles are a bit different than the ones being used in the `TransactionsTable`. On the `TransactionAddress` component the LSK address is not truncated and the bookmarked accounts are displayed with the address as a subtitle.
What do you think if I move the `TransactionAddress` component to `/shared` and remove the switch statement there so that it simply takes the string from the `transactionTypes` constants file?

slaweet: Sounds good. 

There's just one problem you'll face, when you'll move all transaction type labels to the const file - translation `t(...)` function

- If you don't wrap all transaction type labels with `t` in the const file, the scanner will no longer find them, and will remove them from `common.json`
- If you do wrap all transaction type labels with `t` in the const file, it will evaluate only once and won't update on language change.

There are some hacks around it to choose from. In Hub codebase, we have:
https://github.com/LiskHQ/lisk-hub/blob/development/src/constants/networks.js#L24-L25
or 
https://github.com/LiskHQ/lisk-hub/blob/55e5b7bfbf3905bba0fa988572a4460cba8f9c96/src/constants/quickTips.js#L7

src/components/shared/transactionsTable/index.js
```diff
@@ -1,208 +1,229 @@
...
+      );
+    }
+    return (
+      <div className={`${styles.address}`}>
+        <AccountVisual address={transaction[key]} size={32} />
+        <span className={styles.addressValue}>
+          {this.getTransformedAddress(transaction[key])}
+        </span>
+      </div>
+    );
```
slaweet: This is a good candidate for a shared component, as we'll have the same in other pages, e.g.:
https://projects.invisionapp.com/d/main/#/console/18539694/386733858/inspect

The name could be something like `<AccountVisualWithAddress />`

src/constants/transactionTypes.js
colado: Thank you Vit, changes implemented.

i18n/locales/en/common.json
```diff
@@ -146,6 +148,8 @@
   "Edit bookmark": "Edit bookmark",
   "Edit transaction": "Edit transaction",
   "Edit voting": "Edit voting",
+  "Eg. 1,5": "Eg. 1,5",
+  "Eg. 10180477": "Eg. 10180477",
```
slaweet: https://github.com/LiskHQ/lisk-hub/pull/2526#discussion_r331991847 and https://github.com/LiskHQ/lisk-hub/pull/2526/commits/89538d4fca12bf50971850f00f7f03f5648ceb27 could be applied here, too.

src/components/shared/transactionsTable/index.js
colado: Good idea. I've created the shared component `AccountVisualWithAddress` and moved this logic there. I've made the replacement of the address with the bookmarked name optional there, since I noticed the Delegates tab already has the `Name` column and, if bookmarked, that name would appear twice.

i18n/locales/en/common.json
colado: But doesn't it make sense to translate these placeholders too? The way to express things like "e.g." and "i.e." are different in other languages, including German @yasharAyari

slaweet: @colado I think you misunderstood what should be done, please check the links above again.

colado: Got it! :) Changed now.

slaweet:
- :bug: On smaller breakpoint, the calendar widget goes out of view in this case. IMO should be aligned to the right of the input in that case.

Screenshot 2019-10-15 at 13 43 46
- :bug: Message filter should not be displayed before "More filters" is clicked, according to [design](https://projects.invisionapp.com/d/main/#/console/18539694/386607816/comments)
Screenshot 2019-10-15 at 13 48 45

- :books: Message is currently not supported on API, IMO should be removed from the UI until https://github.com/LiskHQ/lisk-service/issues/316 is resolved or there should be a mention somewhere about when it is planned to be solved.

src/components/shared/accountVisualWithAddress/index.js
```diff
@@ -0,0 +1,58 @@
...
+import transactionTypes, { transactionNames } from '../../../constants/transactionTypes';
+import AccountVisual from '../../toolbox/accountVisual';
+import regex from '../../../utils/regex';
+
+class AccountVisualWithAddress extends React.Component {
+  getTransformedAddress(address) {
+    const { isMediumViewPort, bookmarks, showBookmarkedAddress } = this.props;
+
+    if (showBookmarkedAddress) {
+      const bookmarkedAddress = bookmarks.LSK.find(element => element.address === address);
```
slaweet: IMO it's more future prone to use active token from store `settings.token.active` here:
```suggestion
      const bookmarkedAddress = bookmarks[this.props.token.active].find(element => element.address === address);
```

src/components/shared/filterDropdownButton/addressFilter.js
```diff
@@ -0,0 +1,95 @@
...
+  validateAmountField(fieldsObj) {
+    const { t, name } = this.props;
+    let feedback = '';
+
+    const fields = Object.keys(fieldsObj).reduce((acc, field) => {
+      const value = fieldsObj[field].value || '';
+      let error = false;
+
+      if (validateAddress(tokenMap.LSK.key, value) !== 0 && validateLSKPublicKey(value) !== 0 && value !== '') {
+        feedback = t('Invalid address');
```
slaweet:
```suggestion
        feedback = t('Invalid address or public key');
```

src/components/shared/filterDropdownButton/filterContainer.css
```diff
@@ -12,3 +24,14 @@
...
+.actionable {
+  color: var(--color-primary-standard);
+  cursor: pointer;
+  font-weight: var(--font-weight-bold);
+  text-align: right;
+  position: relative;
+  top: 10px;
+  font-size: 12px;
+  margin: 3px 0;
+}
```
slaweet: Did you consider using `<TertiaryButton size="xs">` for "More/Less filters"? It will need some custom css for positioning, but hopefully less than this.
https://github.com/LiskHQ/lisk-hub/blob/4cdc501a6630b1124afdc97d50439d8e2cac1222/src/components/toolbox/buttons/demo.js#L17

src/components/shared/filterDropdownButton/index.js
```diff
@@ -95,31 +145,21 @@ class FilterDropdownButton extends React.Component {
...
-            type="submit"
-            size="s"
-          >
-            {t('Apply Filters')}
-          </PrimaryButton>
+        <form onSubmit={this.applyFilters} className={`${styles.form} filter-container`}>
+          <div className={`${styles.container} ${areFiltersExtended && styles.extendedContainer}`}>
+            {primaryFilters.map(({
+              name, label, placeholder, valueFormatter, type,
+            }) => this.renderFields(name, label, placeholder, valueFormatter, type))}
```
slaweet: Why not make params of `renderFields` to be already in the format you have in `primaryFilters`? So that you can call just
```js
     {primaryFilters.map(this.renderFields)}
```

src/components/shared/filterDropdownButton/index.js
```diff
@@ -95,31 +145,21 @@ class FilterDropdownButton extends React.Component {
...
+            {!areFiltersExtended && this.renderFooter()}
+          </div>
+          {areFiltersExtended && (
+          <div className={styles.container}>
+            {secondaryFilters.map(({
+              name, label, placeholder, valueFormatter, type,
+            }) => this.renderFields(name, label, placeholder, valueFormatter, type))}
```
slaweet: Same `renderFields` comment here as above

src/components/shared/transactionAddress/TransactionAddress.js
```diff
@@ -0,0 +1,35 @@
...
+  address, bookmarks, transactionType, t, token,
+}) => {
+  const account = [...bookmarks.LSK, ...bookmarks.BTC].filter(acc => acc.address === address);
+
+  const renderAddress = () => {
+    if (token === tokenMap.LSK.key) {
+      return account.length ? account[0].title : address;
+    }
+    return account.length ? account[0].title : address.replace(regex.btcAddressTrunk, '$1...$3');
+  };
```
slaweet: Having `return account.length ? account[0].title : ...` twice here doesn't seem DRY.

Could be DRYed with something like this:
```js
const formatter = (token === tokenMap.LSK.key) 
  ? value => value 
  : (value => value.replace(regex.btcAddressTrunk, '$1...$3'));
return account.length ? account[0].title : formatter(address);
```

src/components/toolbox/table/table.css
```diff
@@ -97,3 +97,35 @@
 .sortAsc::after {
   transform: rotate(180deg);
 }
+
+.titleWrapper {
+  position: relative;
+}
+
+.sortInactive {
```
slaweet: This class replaces `sortable` on line 79 in this file, so it can be removed there.

I would comment directly on line 79, but I cannot as it wasn't changed.

src/components/shared/accountVisualWithAddress/index.js
colado: ✅

src/components/shared/filterDropdownButton/addressFilter.js
colado: ✅

src/components/shared/filterDropdownButton/filterContainer.css
colado: I tried this but the `TertiaryButton's` padding and its text alignment make it also a bit difficult to adjust to the design requirements.

src/components/shared/filterDropdownButton/index.js
colado: ✅

src/components/shared/filterDropdownButton/index.js
colado: ✅

src/components/shared/transactionAddress/TransactionAddress.js
colado: ✅

src/components/toolbox/table/table.css
colado: ✅

colado:
@slaweet I've also adjusted the positioning of the calendar dropdown when filters are extended and disabled the message input until the API accepts the message param.

src/components/shared/filterDropdownButton/filterContainer.css
slaweet: Got it. Thanks for clarification 👍

src/components/screens/monitor/transactions/transactions.js
```diff
@@ -1,26 +1,61 @@
...
+    placeholder: 'Address or Public key',
+    name: 'sender',
+    type: 'address',
+  }, {
+    label: t('Message'),
+    placeholder: t('Your message'),
+    name: 'message',
+    type: 'text',
+    // TODO: Enable field when API allows message param
+    disabled: true,
```
slaweet: The disabled field without any explanation feels broken. I would comment out the whole object here.

Efefefef:
As we have not implemented filtering by multiple transaction types, I request Dominic to change placeholder value to single transaction type
image

colado:
@Efefefef @imdominic As discussed, a follow-up ticket will be created to implement the changes on the `Type` field.

PR #2556, opened 2019-10-16
Resolve the UX issues of the registration page - Closes #2465

colado:
### What issue have I solved?

#2465

### How have I implemented/fixed it?
- Implemented all changes on [new registration screens](https://projects.invisionapp.com/d/main?origin=v7#/console/18244706/380009763/preview), except for multiStep, which will be implemented on a different issue.
- The progress bar and multiStep component are being implemented on #2578. This will replace the steps on top of each screen.

### How has this been tested?
Made sure registration and passphrase generation was done successfully and UI aligns with new design.

### Review checklist
- The PR follows our [Test guide](/LiskHQ/lisk-hub/blob/development/docs/TEST_GUIDE.md)
- The PR follows our [CSS guide](/LiskHQ/lisk-hub/blob/development/docs/CSS_GUIDE.md)

PR #2575, opened 2019-10-23
Update second passphrase UI to align with sign in and registration - Closes #2537

### LINKED ISSUE #2537: Update second passphrase UI to align with sign in and registration
### Description
Since we want to change Sign in and Registration pages according to #2412 and #2465, we should update second passphrase inputs to look alike the same input in other screens.
Designs are available on [InVision](https://projects.invisionapp.com/share/PEU4S8GY7TZ#/screens/386125435_6-2_Register_2nd_Passphrase).

colado:
### What issue have I solved?

#2537

- Refactored `passphraseRenderer` to accept passphrase confirmation logic.
- Applied passphrase confirmation logic to 2nd passphrase.
- The last step of 2nd passphrase registration does not contain the box since this component will be changed in a separate issue.

### How has this been tested?
Made sure passphrase creation is successful and screens aligned with design

### Review checklist


src/components/screens/secondPassphrase/confirmPassphrase.js
```diff
@@ -0,0 +1,28 @@
...
+}) => (
+  <Box className={styles.passphraseConfirmation}>
+    <Box.Header>
+      <h2>{t('Confirm your 2nd passphrase')}</h2>
+    </Box.Header>
+    <Box.Content>
+      <div className={styles.passphraseConfirmationContainer}>
+        <p className={styles.info}>{t('Based on your passphrase that was generated in the previous step, select the missing words below')}</p>
+        <PassphraseRenderer
+          values={passphrase.split(' ')}
```
slaweet:
```suggestion
          passphrase={passphrase}
```
💅 It would be DRYer to pass `passphrase` and call `.split(' ')` inside `PassphraseRenderer` than here and  in https://github.com/LiskHQ/lisk-hub/pull/2575/files#diff-618c1dc31da4cb19645765d9e46e985eR52

src/components/screens/secondPassphrase/secondPassphrase.css
```diff
@@ -54,3 +62,27 @@
...
+  position: absolute;
+  display: flex;
+  bottom: 20px;
+  width: 100%;
+  justify-content: center;
+}
+
+.lastStep {
+  min-width: 720px;
+}
```
slaweet: 💅 This class can be replaced by https://github.com/LiskHQ/lisk-hub/blob/1a8d62a40fbdabeab04e063189d41cb536137bc9/src/components/toolbox/box/demo.js#L19

src/components/screens/secondPassphrase/secondPassphrase.test.js
```diff
@@ -60,45 +66,54 @@ describe('SecondPassphrase', () => {
...
+      expect(wrapper.find('h2')).toHaveText('Registration completed');
+      wrapper.find('button.go-to-wallet').simulate('click');
 
-      wrapper.find('.go-to-wallet').first().simulate('click');
       expect(props.history.push).toHaveBeenCalledWith(routes.wallet.path);
     });
 
-    it('should handle registerSecondPassphrase failure', () => {
+    // TODO enable the next test when implemented
+    it.skip('should handle registerSecondPassphrase failure', () => {
```
slaweet: 🐛 Failure path not implemented

src/components/screens/secondPassphrase/summaryStep.js
```diff
@@ -22,6 +21,8 @@ const SummaryStep = ({
           passphrase: account.passphrase,
           account: account.info.LSK,
           callback: ({ success, error }) => {
+            // TODO: To be tackled during multiStep refactoring
+            /* istanbul ignore next */
```
slaweet: AFAIK this is just because of the skipped test case for failure path. Failure path should IMO be part of this PR, and then it should be fine to remove this comment.

src/components/shared/passphraseBackup/index.test.js
```diff
@@ -0,0 +1,20 @@
...
+    account: {
+      passphrase: generatePassphrase(),
+    },
+    t: jest.fn(),
+  };
+
+  it('should show tip after copying passphrase', () => {
+    wrapper = mount(<PassphraseBackup {...props} />);
+    wrapper.find('CopyToClipboard').at(0).simulate('click');
+    expect('tip').toBeTruthy();
```
slaweet: 💅 This assertion will be always true, no matter what you'll do with the component. It's not checking what's in `wrapper`

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,231 @@
...
+import styles from './passphraseRenderer.css';
+import { PrimaryButton, TertiaryButton } from '../../toolbox/buttons/button';
+
+class PassphraseRenderer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      indexes: [2, 9],
+      fieldSelected: undefined,
+      displayOptions: undefined,
```
slaweet: 🐛 (or 📚?)  Based on the designs, I think the initial state should be the first missing word selected. There is no screen in the designs where both missing words are empty, none of them is selected, and no options are displayed. In that case `state.displayOptions` is not needed, because it should depend on `state.fieldSelected`

This also applies to the previous Story https://github.com/LiskHQ/lisk-hub/pull/2556

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,231 @@
...
+      chosenWords: {},
+      options: {},
+      isCorrect: false,
+      hasErrors: false,
+      disabledButton: true,
```
slaweet: 💅 I think it's not necessary to store this value, it can be easily computed from other state values

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,231 @@
...
+    };
+
+    this.handleConfirm = this.handleConfirm.bind(this);
+  }
+
+  UNSAFE_componentWillMount() { // eslint-disable-line camelcase
```
slaweet: 💅 IMO logic in this method should be moved to `constructor` since `UNSAFE_componentWillMount` is deprecated.

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,231 @@
...
+
+    this.setState({
+      isCorrect: status,
+      hasErrors: !status,
+    });
+
+    this.timeout = setTimeout(cb, 1500);
+  }
+
+  getRandomIndexesFromPassphrase(qty) {
```
slaweet: 💅 naming a method `getSomething` suggests it will `return` something, so I'd suggest:
```suggestion
  setRandomIndexesFromPassphrase(qty) {
```

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,231 @@
...
+  componentWillUnmount() {
+    clearTimeout(this.timeout);
+  }
+
+  handleConfirm() {
+    const { chosenWords, indexes } = this.state;
+    const { values } = this.props;
+
+    const answers = Object.values(chosenWords);
+    const status = answers.filter((answer, index) => answer === values[indexes[index]])
```
slaweet: 💅 IMO more descriptive name would be
```suggestion
    const isCorrect = answers.filter((answer, index) => answer === values[indexes[index]])
```

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,231 @@
...
+    this.setState({
+      options: {
+        [indexes[0]]: options[0],
+        [indexes[1]]: options[1],
+      },
+      indexes,
+      answers: [],
+      hasErrors: false,
+      chosenWords: {},
+      fieldSelected: undefined,
```
slaweet: Here same is in the comment above

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,231 @@
...
+      },
+    });
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.hasErrors && !this.props.hasErrors) {
+      this.setState({
+        fieldSelected: undefined,
+        displayOptions: undefined,
+        chosenWords: {},
```
slaweet: I don't see why this is needed. AFAIK `hasErrors` is in `state`, not in `props`.

src/components/shared/passphraseRenderer/index.js
colado: Which comment?

slaweet: https://github.com/LiskHQ/lisk-hub/pull/2575#discussion_r339495583

src/components/screens/secondPassphrase/secondPassphrase.css
colado: ✅

src/components/screens/secondPassphrase/secondPassphrase.test.js
colado: ✅

src/components/screens/secondPassphrase/summaryStep.js
colado: I have implemented the failure path but I still need design to provide the final copy and to let me know if there is anything else there that would need to change in case of failure so I will get back to this when I hear back from them.

src/components/shared/passphraseRenderer/index.js
colado: You're right. I made the changes. It should also be possible to reselect an option once you've already selected it so I changed that also.

src/components/shared/passphraseRenderer/index.js
colado: True. I forgot to remove it after moving the logic from the screen to the component.

src/components/shared/passphraseRenderer/index.js
colado: ✅

src/components/shared/passphraseBackup/index.test.js
colado: I made the changes to check the text of the tip but somehow even when `this.state.showTip` is true, the test returns the content of the `tip` as empty. Can you check if there is something missing on the test?

src/components/shared/passphraseRenderer/index.js
colado: You're right. Although from what I see there's no reason it shouldn't work inside `componentDidMount` instead of `componentWillMount`. I moved it there, what do you think?

src/components/shared/passphraseRenderer/index.js
colado: ✅

src/components/shared/passphraseRenderer/index.js
colado: ✅

src/components/shared/passphraseRenderer/index.js
colado: ✅

colado:
Blocked until missing copy is received

src/components/screens/secondPassphrase/summaryStep.js
colado: ✅

src/components/shared/passphraseRenderer/index.js
slaweet: I think it's less code and more clear to the reader if it's all in one space the `constructor`.

src/components/shared/passphraseBackup/index.test.js
slaweet: This was the problem: https://github.com/LiskHQ/lisk-hub/pull/2575/commits/8f2431b286c478a392acd6961aa20ddf2c1eda35

I also did two more commits on this file. See commit messages for more info...

src/components/shared/passphraseRenderer/index.js
slaweet: I see you stopped using `disabledButton`, but forgot to remove this line
```suggestion
```

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,226 @@
...
+    const { chosenWords, indexes } = this.state;
+    const { values } = this.state;
+
+    const answers = Object.values(chosenWords);
+    const isCorrect = answers.filter((answer, index) => answer === values[indexes[index]])
+      .length === 2;
+
+    const cb = isCorrect
+      ? () => this.props.nextStep()
+      : () => this.setRandomIndexesFromPassphrase(2);
```
slaweet: One more thought about making this even DRYer. If you define `const qty = 2` inside `setRandomIndexesFromPassphrase`, then you don't need to create new functions here and use just:

```js
    const cb = isCorrect ? this.props.nextStep : this.setRandomIndexesFromPassphrase;
```

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,226 @@
...
+      hasErrors: !isCorrect,
+    });
+
+    this.timeout = setTimeout(cb, 1500);
+  }
+
+  setRandomIndexesFromPassphrase(qty) {
+    const { values } = this.state;
+    let idxs = values.map((w, index) => index);
+    const indexes = [...Array(qty)]
```
slaweet: the `qty` variable name is not very descriptive, could be something like:
```suggestion
    const indexes = [...Array(numberOfMissingWords)]
```

src/components/shared/passphraseRenderer/index.js
colado: ✅

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,225 @@
...
+        return index;
+      })
+      .sort((a, b) => a - b);
+    const options = this.assembleWordOptions(values, indexes);
+
+    this.setState({
+      options: {
+        [indexes[0]]: options[0],
+        [indexes[1]]: options[1],
+      },
```
slaweet: These two look very similar (not DRY):
https://github.com/LiskHQ/lisk-hub/blob/a2f966a9867a3dd7985f41d47bbe8ffbb63a04a8/src/components/shared/passphraseRenderer/index.js#L74-L80
https://github.com/LiskHQ/lisk-hub/blob/a2f966a9867a3dd7985f41d47bbe8ffbb63a04a8/src/components/shared/passphraseRenderer/index.js#L28-L34

It could be resolved if `this.assembleWordOptions` already returned `options` in the format that is used in `state`

src/components/shared/passphraseRenderer/index.js
colado: ✅

src/components/shared/passphraseRenderer/index.js
```diff
@@ -0,0 +1,206 @@
...
+
+class PassphraseRenderer extends React.Component {
+  constructor(props) {
+    super(props);
+    const initialIndexes = [2, 9];
+    this.state = {
+      indexes: initialIndexes,
+      fieldSelected: initialIndexes[0],
+      chosenWords: {},
+      options: this.assembleWordOptions(props.passphrase.split(' '), initialIndexes),
```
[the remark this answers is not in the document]
slaweet: Solved in https://github.com/LiskHQ/lisk-hub/pull/2575/commits/755574759515149c0c5cdbe22a31827a2f130591

colado:
Thank you, @Efefefef. 
- Paper wallet title changed.
- Title on summary step changed.

Regarding the final step box, I have spoken with Julian and it was agreed to leave this as it is, since they have redesigned the whole `TransactionResult` component (the one used in the last step), so it will be changed in a separate issue for all the transactions. I updated the description here for clarification.
cc @imdominic.

jooltr:
Sorry for the delay guys, I thought it was done.

🎨Please double check the height of the passphrase container, it seems a bit smaller than the one provided and it feels clustered.

🎨Please double check the spacing between the line and the numbers, it seems a bit longer than the one provided and it feels clustered.

Thanks

colado:
Thank you, @jooltr. I have added some vertical padding to the passphrase container and fixed the space between the numbers and lines. Let me know if it looks ok to you now.

jooltr:
@colado the 2nd passphrase registration looks great! Thank you.

@sridharmeganathan please note that the feedback posted above is still relevant for the  passphrase container in the `Create an Account` screen.

Thanks!!

PR #2581, opened 2019-10-24
Fix sending BTC with decimal points - Closes #2571

### LINKED ISSUE #2571: BTC: Transaction fee is invalid for decimals in Deutsch
### Steps to reproduce

1. Switch to Deutsch
1. Switch to BTC
1. Try to send some decimal amount of money

### Actual result

Impossible to send transaction

### Expected result

Fee validation understands deutsch notation

### Version

1.21, 1.22

image

slaweet:
### What issue have I solved?

Closes #2571

But it happened the same in German and English localization

Fix disabled condition of submit button in BTC send form

It was wrong because formBtc was storing state.fields.amount.error
and passing it in props.extraFields to formBase.

Solved by merging fields and extraFields so that amount from fields is used.

The other problem was that the dynamic fee was computed based on the previous state of the amount. I fixed this by moving the fee update only after amount state is updated.

### How has this been tested?

Follow steps in #2571

### Review checklist

### DIFF
--- a/src/components/screens/wallet/send/form/formBtc.js
+++ b/src/components/screens/wallet/send/form/formBtc.js
@@ -119,19 +119,15 @@ export default class FormBtc extends React.Component {
       : t('Invalid amount');
   }
 
-  // TODO move dynamic fee calculation and presentation to a separate component
   getCalculatedDynamicFee(dynamicFeePerByte) {
     const { fields: { amount }, unspentTransactionOutputs } = this.state;
-    if (amount.error) {
-      return 0;
-    }
-    const feeInSatoshis = btcTransactionsAPI.getTransactionFeeFromUnspentOutputs({
-      unspentTransactionOutputs,
-      satoshiValue: toRawLsk(amount.value),
-      dynamicFeePerByte,
-    });
-
-    return feeInSatoshis;
+    return amount.error
+      ? 0
+      : btcTransactionsAPI.getTransactionFeeFromUnspentOutputs({
+        unspentTransactionOutputs,
+        satoshiValue: toRawLsk(amount.value),
+        dynamicFeePerByte,
+      });
   }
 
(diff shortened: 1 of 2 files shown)


src/components/screens/wallet/send/form/formBtc.js
```diff
@@ -114,14 +120,14 @@ export default class FormBtc extends React.Component {
...
-  getCalculatedDynamicFee(dynamicFeePerByte, value) {
+  getCalculatedDynamicFee(dynamicFeePerByte) {
     const { fields: { amount }, unspentTransactionOutputs } = this.state;
     if (amount.error) {
       return 0;
     }
     const feeInSatoshis = btcTransactionsAPI.getTransactionFeeFromUnspentOutputs({
       unspentTransactionOutputs,
-      satoshiValue: toRawLsk(value || amount.value),
+      satoshiValue: toRawLsk(amount.value),
```
osvaldovega: 💅🏻 Vit just a minor thing, I know you didn't change the whole function but maybe you can update it to something like this.

```suggestion
      getDynamicFeeCalculation(dynamicFeePerByte) {
    const { fields: { amount }, unspentTransactionOutputs } = this.state;
    return amount.error
    ? 0;
    : btcTransactionsAPI.getTransactionFeeFromUnspentOutputs({
      unspentTransactionOutputs,
      satoshiValue: toRawLsk(amount.value),
      dynamicFeePerByte,
    });
  }
```

src/components/screens/wallet/send/form/formBtc.js
```diff
@@ -114,14 +120,14 @@ export default class FormBtc extends React.Component {
   }
 
   // TODO move dynamic fee calculation and presentation to a separate component
```
osvaldovega: 💅🏻based on our conversation this comment can be remove as the function will remain in this file

PR #2585, opened 2019-10-25
Minify the bundles in production build - Closes #2584

### LINKED ISSUE #2584: Minify the bundles for production
### Description
We should buld the app for production and minify our bundles for shipping to production.
We [used to do it](https://github.com/LiskHQ/lisk-hub/blob/454eaee37cd784418d2e6731d1b10242f910bc0e/config/webpack.config.prod.js) in the past and we [stop it](https://github.com/LiskHQ/lisk-hub/commit/efc7f37cd0df2435e15305496053ac2ee8328af1#diff-1a40e6c8e8713d4af22007160e07a446) it for some reason.

### Motivation
Explained [here](https://reactjs.org/docs/optimizing-performance.html#profiling-components-with-the-chrome-performance-tab) and [here](https://codeengineered.com/blog/why-minify-javascript/) and tones of other places.

reyraa:
### What issue have I solved?
#2584

The was an issue with exporting modules in [toolbox/box](https://github.com/LiskHQ/lisk-hub/tree/development/src/components/toolbox/box). modules lost their references due to minification. This was fine in development mode thanks to source mapping.

### How has this been tested?
- Built for development: the app should run and work.
- Build for production, the bundles should be compressed and the app should work.

### Notes
It's highly discouraged to [use displayName as a property in code logic](https://github.com/LiskHQ/lisk-hub/blob/development/src/components/toolbox/box/index.js#L18). but it's not a concern of this PR. We can do it later.

### Review checklist


package.json
```diff
@@ -147,13 +147,13 @@
     "cypress-cucumber-preprocessor": "1.12.0",
     "cypress-pipe": "1.3.3",
     "del-cli": "1.1.0",
-    "electron": "4.2.4",
-    "electron-builder": "20.38.3",
+    "electron": "7.0.0",
+    "electron-builder": "21.2.0",
```
yasharAyari: Please don't update the version of all the electron dependencies. @osvaldovega works on another issue that will take care of all of those upgrades.

package.json
```diff
@@ -147,13 +147,13 @@
...
     "electron-devtools-installer": "2.2.4",
     "electron-ipc-mock": "0.0.3",
-    "electron-json-storage": "4.1.5",
-    "electron-rebuild": "1.8.2",
-    "electron-updater": "4.0.6",
+    "electron-json-storage": "4.1.8",
+    "electron-rebuild": "1.8.6",
+    "electron-updater": "4.1.2",
```
yasharAyari: same as above.

slaweet:
Do we really need to stop using the `Box.Header` convention? I find it pretty neat, not to have to import all the sub-components.
The naming convention was inspired by react-bootstrap, e.g. see https://react-bootstrap.github.io/components/cards/ so there should be a way to keep it.
 
They use `displayName`:
https://github.com/react-bootstrap/react-bootstrap/blob/ad226f8cd1b2432e0be81eb9fd091f10ba595f66/src/createWithBsPrefix.js#L26
They use minification:
https://github.com/react-bootstrap/react-bootstrap/blob/ad226f8cd1b2432e0be81eb9fd091f10ba595f66/tools/dist.webpack.config.js#L9
and webpack 4:
https://github.com/react-bootstrap/react-bootstrap/blob/ad226f8cd1b2432e0be81eb9fd091f10ba595f66/package.json#L123

If it can work for them, it should work for us, too.

reyraa:
@slaweet I haven't read their code, but the way you did it spoils tree shaking and causes the issue explained above. 
TBH, I don't see the neatness in it, I'd rather import a component when I need, not import the whole dynasty to use the runt of the family. 
but we can consider this for the future, for now I think building in production mode and comporessing for production is more important.

PR #2598, opened 2019-10-28
Discreet mode is broken - closes #2550

### LINKED ISSUE #2550: Discreet mode is broken
### Steps to reproduce

1. Login with enabled discreet mode
1. Go to "Delegates"
1. Go to the detail of one of the delegates

### Actual result

Your balance in the header is not blurred.

### Expected result

Your balance in the header is blurred.

### Version

- 1.21.0 https://ci.lisk.io/test/lisk-hub/release-1.21.0/#/explorer/accounts/5201600508578320196L
- 1.22.0 https://ci.lisk.io/test/lisk-hub/release-1.22.0/#/explorer/accounts/5201600508578320196L

Delegates list page ✅ :
Screenshot 2019-10-14 at 11 22 19

Delegate detail page 🐛 :
Screenshot 2019-10-14 at 11 22 27

osvaldovega:
### What issue have I solved?

-- #2550 

The DiscreetMode component check if the account that is in screen is the same as the logged user based on that execute the function to blur or not.
The fix consist is pass a prop to tell the component if need to evaluate the current address or not.

### How has this been tested?

1. Login into the app.
2. Go to wallet and select any account and see that the header amount is still blur.
3. Go to Delegates page.
4. Do click on any delegate account to view the details.
5. The header account mount should be blur

### Review checklist

### DIFF
--- a/src/components/shared/discreetMode/discreetMode.js
+++ b/src/components/shared/discreetMode/discreetMode.js
@@ -20,11 +20,11 @@ class DiscreetMode extends Component {
 
   shouldEnableDiscreetMode() {
     const {
-      addresses, location, isDiscreetMode, shouldEvaluate,
+      addresses, location, isDiscreetMode, shouldEvaluateForOtherAccounts,
     } = this.props;
     if (!isDiscreetMode) return false;
 
-    if (shouldEvaluate) {
+    if (shouldEvaluateForOtherAccounts) {
       if (addresses.length && location.pathname.includes(routes.transactions.path)) {
         return this.handleBlurOnTransactionDetailsPage();
       }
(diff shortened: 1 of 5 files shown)


src/components/shared/discreetMode/discreetMode.js
```diff
@@ -38,13 +44,15 @@ class DiscreetMode extends Component {
 
 DiscreetMode.defaultProps = {
   addresses: [],
+  shouldEvaluate: false,
```
slaweet: 💅 `shouldEvaluate` is not telling me what is the flag doing.

I find it really hard to find a better name. Best I got so far is:
```suggestion
  blurShouldBeDisabledOnOtherAccounts: false,
```

If you can think of some even clearer name, I'd be happy to accept it.

osvaldovega: Thx Vit for the review, you are right the prop name is for evaluate or check that logic, so I just changed.

About your concern for the behavior of other wallet account, I checked with Ali again and yes it is fine that when you see other account you can don't see any amount blur, so the behavior is still the same.

PR #2601, opened 2019-10-29
Implement the new error style - closes #2543

### LINKED ISSUE #2543: Implement the new error style
### Description
We'd like to implement the new error style proposed by design team.

osvaldovega:
### What issue have I solved?

-- #2543

Update the Feedback component and Input component to the see the feedback information to look same as design.

The change is all about css and not js components, then confirm and check all the places where this is use continue working as before.

### How has this been tested?

1. Run the app.
2. Check all the pages where input and feedback component works.
3. Trigger the error (feedback) message depending of the input type and see that the error message show as the designs.

### Review checklist


src/components/toolbox/autoSuggest/index.js
```diff
@@ -29,6 +31,9 @@ class AutoSuggest extends React.Component {
     this.onSelectItem = this.onSelectItem.bind(this);
     this.resetListIndex = this.resetListIndex.bind(this);
     this.handleUpdateIndex = this.handleUpdateIndex.bind(this);
+    this.onBlur = this.onBlur.bind(this);
+    this.onFocus = this.onFocus.bind(this);
+    this.isOnError = this.isOnError.bind(this);
```
reyraa: If you use arrow function, you don't need these.

src/components/screens/bookmarks/bookmarksList/bookmarksList.js
```diff
@@ -92,7 +95,11 @@ class BookmarksList extends React.Component {
 
   onTitleChange({ target }) {
     this.setState({
-      eddittedTitle: target.value,
+      ...this.state,
+      eddittedTitle: {
+        value: target.value,
```
reyraa: Error and feedback are conceptually the more similar than the input value. I think they should be grouped together.

src/components/toolbox/autoSuggest/index.js
osvaldovega: hey Ali thanks for the review totally agree with you, I like more to use arrow functions in class instead this, however we did't use that standard so far.

I think you are talking about this.

```
class Name extends Component {

  onBlur = () => {
    ...
  }

  render() {
    return (
      <Component onBlur={this.onBlur} />
    )
  }
}
```

If you are talking of use the arrow functions in the render directly we try to not use unless be really necesary because cause performance issues, not that the app wont work but it is not a good practice.

src/components/screens/bookmarks/bookmarksList/bookmarksList.js
osvaldovega: I dont understand, let me check it

src/components/toolbox/autoSuggest/index.js
slaweet: I agree that arrow functions are less code than binding everything, but it terms of consistency, we use `bind` in all class-based components. So if we want to set a new standard for how to handle this, I'd say the solution should be to use react Hooks and no class-based components.

We already agreed on using hooks in the last technical meeting.

src/components/toolbox/autoSuggest/autoSuggest.test.js
```diff
@@ -83,5 +84,37 @@ describe('Recipient Input', () => {
...
+  });
+
+  it('should show error message if value is wrong and then remove it when value is correct', () => {
+    const wrongValue = { target: { name: 'recipient', value: 'HHH' } };
+    const correctValue = { target: { name: 'recipient', value: 'FRG' } };
+    wrapper = mount(<AutoSuggest {...props} />);
+    wrapper.find('input.recipient').simulate('focus');
+    wrapper.find('input.recipient').simulate('change', wrongValue);
+    wrapper.setProps({
+      ...props,
```
slaweet:
```suggestion
```

> Similar to setState, this method accepts a props object and will merge it in with the already existing props.

https://airbnb.io/enzyme/docs/api/ShallowWrapper/setProps.html

src/components/shared/searchBar/searchBar.css
```diff
@@ -22,3 +22,14 @@
     height: 3px;
   }
 }
+
+.searchInput + span {
```
slaweet: 💅 According to our CSS guide, we shouldn't use generic tags (like span or div) for CSS selectors.

src/components/shared/filterDropdownButton/dateFieldGroup.test.js
```diff
@@ -75,21 +75,21 @@ describe('DateFieldGroup', () => {
       wrapper.find('.dateToInput input').simulate('change', { target: { name: 'dateTo', value: '12.12.16' } });
       jest.advanceTimersByTime(300);
       wrapper.update();
-      expect(wrapper).toContainMatchingElement('.feedback.show');
+      expect(wrapper).toContainMatchingElement('.feedback');
```
slaweet: 💅 This is not testing anything anymore. I fixed the same problem in `amountFieldGroup.js` in https://github.com/LiskHQ/lisk-hub/pull/2601/commits/d9bc6ad522dbcc494febc3f1a41d71d246459bc6

osvaldovega:
thx @reyraa @slaweet for the review feedback, I just push the last changes, I was confirming that those new chnages dont brake anything else.

Related to the comment first comment, hooks will be a nice approach but as with that I need to move some logic out and create more file including unit test, I think we can do in a near future, at this point thx to the update of babel config file (thx Ali) I just use the recommendation from Ali.
In any case thx guys for the suggests.

src/components/screens/bookmarks/bookmarksList/bookmarksList.js
```diff
@@ -92,13 +93,15 @@ class BookmarksList extends React.Component {
 
   onTitleChange({ target }) {
     this.setState({
-      eddittedTitle: target.value,
+      ...this.state,
```
reyraa: What's this for? setState merges the given object. You don't need to add this.

src/components/screens/registerDelegate/selectName/SelectName.js
```diff
@@ -130,7 +131,12 @@ class SelectName extends React.Component {
             }
           </p>
 
-          <label className={styles.nicknameLabel}>{t('Your nickname')}</label>
+          <label className={styles.nicknameLabel}>
+            {t('Your nickname')}
+            <Tooltip classNam>
```
reyraa: Is `classNam` a typo?

src/components/toolbox/autoSuggest/index.js
```diff
@@ -11,6 +11,7 @@ class AutoSuggest extends React.Component {
     this.state = {
       dropdownIndex: 0,
       isLoading: false,
+      isOnFocus: true,
```
reyraa: `isBlurred` and `isFocused` are more popular terms.

osvaldovega: true my mistake, I just fixed

src/components/screens/registerDelegate/selectName/SelectName.js
osvaldovega: the autocomplete feature messing with me. Yes it is a typo thanks for the catchup Ali

src/components/screens/bookmarks/bookmarksList/bookmarksList.js
osvaldovega: true I forget remove it from the last change, is fixed now

Efefefef:
@reyraa @osvaldovega I need designs

PR #2616, opened 2019-11-01
Fix UX issues with voting - Resolves #2466

### LINKED ISSUE #2466: Resolve UX issue with voting
### Description
According to the user testing session, we have discovered a few UX issues affecting the experience of our users in voting page.
The design team have provided us with a new set of design guides which are created to resolve the aforementioned issues.

Acceptance criteria 
- The implementation must match the design provided on [InVision](https://projects.invisionapp.com/share/BXU5J2TS2RN#/screens/388243627).
- The white background expands to the entire screen width while the inner content is aligned within confines of grid system.
- The header must appear by replacing the existing content, the background is transparent in this position. Screenshot 2019-10-23 at 11 30 55
- When the screen is scrolled to top, the header moves along side other components. When reached the main header, it sticks to the bottom of the main header and its background turns white with a simple fade-in animation.
- When the page scrolled bottom, the sticky header starts moving down along side other components only after it overlays its initial position.
- When in sticky state, if should cover other banners such as new version or anonymous analytics plan.

A basic implementation of the transition animation explained above can be found [here](https://www.w3schools.com/howto/howto_js_sticky_header.asp).

reyraa:
### What issue have I solved?
#2466 
- Change the JSX conditions to match the flow defined in the design files.
- Updated the designs.
    - There were some inconsistencies in the design and with the other pages, I made decisions accordingly.
- Updated the translation files.

Extra:
I've renamed `verify-message` directory to `verifyMessage` to follow naming conventions.

### Review checklist


src/components/screens/delegates/votingHeader.css
```diff
@@ -2,14 +2,154 @@
...
+  }
+
+  & .bg {
+    position: fixed;
+    width: 100%;
+    height: 78px;
+    left: 0;
+    top: 73px;
+    z-index: 0;
+    border-bottom: solid 1px #e1e3eb;
```
osvaldovega: For all colors we use variables like this.

```suggestion
    border-bottom: solid 1px var(--color-platinum);
```

src/components/screens/delegates/votingHeader.css
```diff
@@ -2,14 +2,154 @@
...
+    }
+
+    & > span,
+    & h5 {
+      height: 50%;
+      margin: 0;
+      display: flex;
+      justify-content: flex-start;
+      padding-left: 15px;
+      border-right: solid 1px #e1e3eb;
```
osvaldovega: same in here

```suggestion
      border-bottom: solid 1px var(--color-platinum);
```

src/components/screens/delegates/votingHeader.css
```diff
@@ -2,14 +2,154 @@
...
+      display: flex;
+      margin: 0;
+
+      & img {
+        width: 100%;
+      }
+    }
+
+    & > span,
+    & h5 {
```
osvaldovega: Ali, I think you don't know this, but according to the css rules that we have in the docs, it is not a big deal, but just for you know about it.

So in this particular case div and span need a class but the h5 for example is ok.

> We should use cascading css for semantic elements (h1, h2, ul, li,...) but not for generic elements (div and span) - if you need to modify their styles, add a new className to them.

src/components/screens/delegates/votingHeader.css
```diff
@@ -2,14 +2,154 @@
...
+      justify-content: flex-start;
+      padding-left: 15px;
+      border-right: solid 1px #e1e3eb;
+      width: calc(100% - 20px);
+    }
+
+    & > span {
+      align-items: flex-end;
+      font-size: var(--font-size-small-secondary);
+      color: var(--color-slate-gray);
```
osvaldovega: same here

> We should use cascading css for semantic elements (h1, h2, ul, li,...) but not for generic elements (div and span) - if you need to modify their styles, add a new className to them.

src/components/screens/delegates/votingHeader.css
```diff
@@ -2,14 +2,154 @@
...
+
+    &.fee {
+      padding: 19px 11px 19px 29px;
+      width: 190px;
+
+      & > span,
+      & h5 {
+        border-right: none;
+      }
+    }
```
osvaldovega: here too

> We should use cascading css for semantic elements (h1, h2, ul, li,...) but not for generic elements (div and span) - if you need to modify their styles, add a new className to them.

src/components/screens/delegates/votingHeader.css
```diff
@@ -2,14 +2,154 @@
...
+      display: block;
+      margin: 0;
+      padding: 0;
+      height: 45px;
+
+      & span {
+        padding: 0;
+        margin-top: 6px;
+        width: auto;
+      }
```
osvaldovega: here

> We should use cascading css for semantic elements (h1, h2, ul, li,...) but not for generic elements (div and span) - if you need to modify their styles, add a new className to them.

src/components/screens/delegates/votingHeader.css
reyraa: Done

src/components/screens/delegates/votingHeader.css
reyraa: Done

src/components/screens/delegates/votingHeader.css
reyraa: Done

src/components/screens/delegates/votingHeader.css
reyraa: Done.

src/components/screens/delegates/votingHeader.css
reyraa: Done

src/components/screens/delegates/votingHeader.css
reyraa: Done

src/components/toolbox/icon/index.js
```diff
@@ -151,6 +154,9 @@ export const icons = {
   txDefault,
   txDelegate,
   txVote,
+  addedVotes,
+  removedVotes,
+  totalVotes,
```
yasharAyari: Please add these new icons to the end of the list.

reyraa:
Regarding
- 1024 there's a ticket: #2630, not in the scope of this ticket.
- _Edit votes_, we changed it based on recommendation by native speakers.
- Width of buttons, we should maintain a general rule, enforced by toolbox components. Doesn't make sense to change it here. we should only add the title and the box size get defined by the content.
- Tooltip position: Change it to bottom. looks better now. regarding other details, we have a toolbox so we don't have to re-specify the rules over and over. If there's an issue we need a dedicated ticket to fix it in toolbox and check it thought out all screens.
- Button border color: Same thing.
- The box margin top: Not touched by me, out of the scope of this ticket. Will check and create other ticket if necessary.
- Alignment of titles: Looks aligned to me
Screenshot 2019-11-08 at 15 18 46
- Error page
   - Argument must be a string: Requires a dedicate ticket #2634, please review the steps to reproduce.

PR #2619, opened 2019-11-04
Display the list of top accounts - closes #2454

### LINKED ISSUE #2454: Display the list of top accounts
### Description
As a user I'd like to see the list of accounts with the most tokens in Lisk Hub. I'd like to know what percentage of the total LSK supply they hold.

Design: https://projects.invisionapp.com/share/ZRU30GY5NTX#/screens/386073386_Discovery_-_Top_Accounts

Acceptance Criteria

- [ ] Page should  loaded with top 30 accounts initially .
- [ ] Selecting "more" button should add next 30 top accounts (as per our standard) with the existing list and this step continues 
- [ ] Clicking a row should lead to detail page of given account
- [ ] The page should match with the proposed design. Refer inVision.

osvaldovega:
### What issue have I solved?

-- #2454

For this feature, I had to do the following.

- [x] Add a new tab for Accounts in Monitor page.
- [x] Add a new component called Accounts table to show the information.
- [x] Create a screen component called Accounts under monitor directory, where AccountsTable is render.
- [x] Add a new function to fetch the top accounts in the liskService file under API directory.
- [x] Add unit test for Accounts component.

### How has this been tested?

1. Run the app.
2. Login to Mainnet or Testnet.
3. Go to Monitor page.
4. In the tabs options select Accounts (Top Accounts).
5. See that everything works as should be even Load More button

### Review checklist


slaweet:
AFAIK, "Supply" will not be provided by the Service accounts API and it is something you can compute locally, it's just `account.balance / response.data.supply`, where `response` comes  from https://testnet.lisk.io/api/node/constants

osvaldovega:
Thanks Vit for the feedback I just push the changes, you can check it again

src/components/shared/accountsTable/accountsTable.css
```diff
@@ -0,0 +1,11 @@
...
+
+.content {
+  padding: 0 !important;
+  counter-reset: section;
+}
+
+.counter::before {
+  counter-increment: section;
+  content: "#" counter(section);
+}
```
slaweet: 🏆 Cool idea 😄

src/components/shared/accountsTable/index.js
```diff
@@ -0,0 +1,135 @@
...
+                          <LiskAmount val={account.balance} roundTo={0} token={tokenMap.LSK.key} />
+                        ),
+                      },
+                      {
+                        header: t('Supply'),
+                        className: `${grid['col-xs-2']} ${grid['col-md-1']}`,
+                        id: 'supply',
+                        getValue: (account) => {
+                          const amount = new BigNumber(account.balance / supply * 100);
+                          return <span>{`${amount.toFormat(2)} %`}</span>;
```
slaweet: 🐛 The supply doesn't use the correct decimal separator in German locale

For an example of implementation, refer to https://github.com/LiskHQ/lisk-desktop/blob/c156288dbf5502e56bfe5153849ed1463147275d/src/components/shared/delegatesTable/index.js#L40

src/components/shared/accountsTable/index.js
```diff
@@ -0,0 +1,135 @@
...
+                      {
+                        header: t('Owner'),
+                        className: `${grid['col-xs-3']} ${grid['col-md-2']}`,
+                        id: 'owner',
+                        getValue: (account) => {
+                          const delegateUsername = account.delegate ? account.delegate.username : '';
+                          const text = account.knowledge
+                            ? `${account.knowledge.owner} ${account.knowledge.description}`
+                            : delegateUsername;
+                          return (<span>{text}</span>);
```
slaweet: 💅 Is the span needed? IMO less code is usually better
```suggestion
                          return text;
```

src/components/shared/accountsTable/index.js
```diff
@@ -0,0 +1,135 @@
...
+  handleLoadMore = () => {
+    const { accounts } = this.props;
+    accounts.loadData({ offset: accounts.data.length });
+  }
+
+  render() {
+    const {
+      accounts,
+      emptyStateMessage,
+      isLoadMoreEnabled,
```
slaweet:
```suggestion
```
💅 This prop is IMO not needed since it's always `true`.

src/components/shared/accountsTable/index.js
```diff
@@ -0,0 +1,135 @@
...
+class AccountsTable extends React.Component {
+  handleLoadMore = () => {
+    const { accounts } = this.props;
+    accounts.loadData({ offset: accounts.data.length });
+  }
+
+  render() {
+    const {
+      accounts,
+      emptyStateMessage,
```
slaweet: 💅 This prop is IMO not needed since it's not passed.

```suggestion
```

src/components/toolbox/table/index.js
```diff
@@ -47,9 +47,9 @@ const Table = ({
         ))}
       </TableRow>
       )}
-      {data.map(row => (
+      {data.map((row, index) => (
         <TableRow
-          key={row.id}
+          key={index}
```
slaweet: 💅  Index as a key is an anti-pattern
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

Instead, you can do something like in delegates table which also doesn't have `id` coming from the API
https://github.com/LiskHQ/lisk-desktop/blob/c156288dbf5502e56bfe5153849ed1463147275d/src/components/shared/delegatesTable/index.js#L22

Or it might be cleaner to add another prop to the `Table` component that would specify what should be the `primaryKey`. used as `key`

src/utils/api/lsk/liskService.js
```diff
@@ -131,6 +131,21 @@ const liskServiceApi = {
...
+    searchParams: {
+      limit: DEFAULT_LIMIT,
+      ...searchParams,
+    },
+  }),
+
+  getNetworkStatus: async ({ networkConfig }, searchParams) => liskServiceGet({
+    networkConfig,
+    path: '/api/v1/network/status',
+    searchParams,
```
slaweet:
```suggestion
```
💅AFAIK There are no `searchParams` that would this endpoint accept

src/components/screens/monitor/accounts/accounts.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import MonitorHeader from '../header';
+import AccountsTable from '../../../shared/accountsTable';
```
slaweet:
```suggestion
import AccountsTable from './accountsTable';
```
💡 `AccountsTable` is not really `shared`, as it's used only here and AFAIK there is no plan to use it from somewhere else. I think you copied this pattern from one of these:
- `TransactionsTable`, which is used in both `/monitor/transactions` and `/monitor/blocks/<BLOCK_ID> `
- `DelegatesTable`, which is used in both `/delegates` and `/monitor/delegates/`

In the end, it's not a big deal, so I don't insist on moving it, just wanted to mention it.

src/components/shared/accountsTable/index.js
```diff
@@ -0,0 +1,135 @@
...
+                          return <span>{`${amount.toFormat(2)} %`}</span>;
+                        },
+                      },
+                      {
+                        header: t('Owner'),
+                        className: `${grid['col-xs-3']} ${grid['col-md-2']}`,
+                        id: 'owner',
+                        getValue: (account) => {
+                          const delegateUsername = account.delegate ? account.delegate.username : '';
+                          const text = account.knowledge
```
slaweet:
```suggestion
                          const text = account.knowledge && account.knowledge.owner
```
🐛  service 1.1.0 https://testnet-service-staging.lisk.io/api/v1/accounts/top?limit=30 always has `knowledge`, but it's empty object if not available, so now it would show
Screenshot 2019-11-08 at 13 43 58

osvaldovega:
Thx Vit I checked your points and I fixed

jooltr:
All checked and good on our end. Great work!!

PR #2638, opened 2019-11-12
Make the line chart of the wallet page generic - closes #2625

### LINKED ISSUE #2625: Make the line chart of the wallet page generic

osvaldovega:
### What issue have I solved?

-- #2625

- [x] Create a new generic chart component (toolbox/Graphic).
- [x] Added some default settings for the 3 charts that we going to (line, bar and doughnut).
- [x] Added unit test for the graphicOptions.js file
- [x] Replace the existing balance chart graphic component to use the new one (Graphic). 

### How has this been tested?

1. Run the app.
2. Go to Wallet page.
3. Check the balance chart

**NOTE:** I tested the bar and doughnut chart as weell, I made some simple tests to align the settings for those chart same as the design that we have, but was only for set the settings and see if works. The implementation of these other charts will be when we implement the charts for monitor.

### Review checklist


src/components/toolbox/Graphic/graphicOptions.js
```diff
@@ -0,0 +1,206 @@
...
+      borderWidth: '0',
+      borderColor: '',
+      borderSkipped: 'bottom',
+    },
+  },
+}, baseOptions, options);
+
+
+// ========================================= //
+//                            DOUGGNUT GRAPH                                 //
```
colado: 💅 small typo: `DOUGHNUT`

osvaldovega:
Thank you for the feedback @colado I update the code based on your changes.
About the change in the X-Axis for the balance chart, you are right I changed because there was a problem when present several points, cause the label was not able to red properly, however I decided to ask and check that with Dom and see what he thinks, and he told me that for now is better has the change because is more readable but he will work on a salution for that, so probably in a near future he will change it.
Everything else done.

**NOTE:** By the way I will work on the storybook for chart as Yashar requested so I will include that in this PR

osvaldovega:
Thanks guys for the feedback I implemented that changes based on the comments, please check it again.
Sotybook is added

src/components/toolbox/Chart/chart.stories.js
```diff
@@ -0,0 +1,98 @@
...
+                  label: '123456789L',
+                  data: [35, 55, 23, 67, 98, 46, 134, 66, 70, 33, 100, 120],
+                  backgroundColor: styles.whiteSmoke,
+                  borderColor: styles.borderColor,
+                  pointBorderColor: styles.borderColor,
+                  pointBackgroundColor: styles.whiteColor,
+                  pointHoverBackgroundColor: styles.whiteColor,
+                  pointHoverBorderColor: styles.ultramarineBlue,
+                  pointHoverBorderWidth: 4,
+                  borderWidth: 2,
```
slaweet: IMO the point of creating our toolbox Chart is, that you can call it without any style properties and it looks like Lisk. Of course, any styles can be changed in a specific page, but the default values should IMO be from Lisk style guide. Currently, removing all the styles in this file makes it look like this (not looking like Lisk):
Screenshot 2019-11-18 at 08 57 03

src/components/toolbox/Chart/chart.stories.js
```diff
@@ -0,0 +1,98 @@
...
+              labels: ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Oug', 'Sep', 'Oct', 'Nov', 'Dec'],
+              datasets: [
+                {
+                  label: '123456789L',
+                  data: [35, 55, 23, 67, 98, 46, 134, 66, 70, 33, 100, 120],
+                  barPercentage: 1,
+                  categoryPercentage: 0.5,
+                  barThickness: 'flex',
+                  maxBarThickness: 14,
+                  minBarLength: 0,
```
slaweet: Similar comment to the above. Removing styles makes it look like this:
Screenshot 2019-11-18 at 08 58 15

src/components/toolbox/Chart/chart.stories.js
```diff
@@ -0,0 +1,98 @@
...
+              datasets: [
+                {
+                  label: 'ABC',
+                  data: [100, 634, 329, 98],
+                  backgroundColor: [
+                    '#68d284',
+                    '#4873ec',
+                    '#2c4a93',
+                    '#0d152c',
+                  ],
```
slaweet: Similar comment to the above. Removing styles makes it look like this:
Screenshot 2019-11-18 at 08 58 21

src/constants/chartConstants.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const typeLine = 'line';
+export const typeBar = 'bar';
+export const typeDoughnut = 'doughnut';
+
+export const GRAPH_TYPES = {
+  [typeLine]: Line,
+  [typeBar]: Bar,
+  [typeDoughnut]: Doughnut,
+};
```
slaweet: 💡 IMO this is an implementation detail of our `Chart` component. Putting it into another file and allowing anyone to import it, is IMO violation of encapsulation of our `Chart` component. 

Not to mention that a file in `constants` folder should usually be something small to import anywhere, but not importing it triggers building the whole 'react-chartjs-2', causing problems like described here: https://github.com/LiskHQ/lisk-desktop/pull/2589#pullrequestreview-307836707

src/components/toolbox/Chart/chart.stories.js
osvaldovega: @slaweet thx for the comments, about this comment and the other 2 above, I understand what you mentioned, however I was thinking in a way to do this but the only detail is that this visual especification is part of the **data** object not the **options** and for each chart we need to make a specific function to provide the data in the way we need to present the chart as design, that's why is looking like your mention

src/constants/chartConstants.js
osvaldovega: Im agree, however this was one of the request from Yashar for this file

slaweet: Got it @osvaldovega 

@yasharAyari what do you think about moving only `GRAPH_TYPES` inside `Chart`  and keeping 
```
export const typeLine = 'line';
export const typeBar = 'bar';
export const typeDoughnut = 'doughnut';
```
here in `src/constants/chartConstants.js`?

Since `GRAPH_TYPES` doesn't really contain constants, but the whole implementation of three components.

src/components/toolbox/Chart/index.js
```diff
@@ -0,0 +1,39 @@
...
+      data={data}
+      width={width}
+      height={height}
+      options={optionsByGraphic(type, options)}
+    />
+  );
+};
+
+Chart.propTypes = {
+  type: PropTypes.string,
```
slaweet:
```suggestion
  type: PropTypes.oneOf(Obejct.keys(GRAPH_TYPES)),
```

src/components/toolbox/Chart/index.js
```diff
@@ -0,0 +1,39 @@
...
+import { optionsByGraphic } from '../../../utils/chartOptions';
+
+const Chart = ({
+  type,
+  data,
+  width,
+  height,
+  options,
+}) => {
+  const Graph = type && GRAPH_TYPES[type];
```
slaweet: 💡 You don't need the condition if you apply `PropTypes.oneOf` as suggested below.
```suggestion
  const Graph = GRAPH_TYPES[type];
```

src/components/toolbox/Chart/index.js
```diff
@@ -0,0 +1,39 @@
...
+
```
slaweet:
```suggestion
const dataStylers = {
  [typeLine]: (data) => {
    return data; // TODO implement the logic to set the default Lisk style for line chart
  },
  [typeBar]: (data) => {
    return data; // TODO implement the logic to set the default Lisk style for bar chart
  },,
  [typeDoughnut]: (data) => {
    return data; // TODO implement the logic to set the default Lisk style for doughnut chart
  },,
};

data = dataStylers[type](data)
```

src/components/toolbox/Chart/chart.stories.js
slaweet: You could do something like https://github.com/LiskHQ/lisk-desktop/pull/2638#discussion_r347306539

slaweet: Or, alternatively, you could put the colors into the `chartConstants.js` 
```js
export const doughnutPrimaryColorPallete = [
                    '#68d284',
                    '#4873ec',
                    '#2c4a93',
                    '#0d152c',
                  ],
```

and here just 
```js
import { doughnutPrimaryColorPallete } from '../../constants/chartConstants.js'
```

so that we don't have the definition of the colors in all files that use a chart.

osvaldovega:
@slaweet I just pushed some of the changes, please check what I did and let me know if you have any question about it.

src/utils/chartOptions.js
```diff
@@ -171,47 +164,132 @@ export const barGraphOptions = options => merge({
...
+      barPercentage: 1,
+      categoryPercentage: 0.5,
+      barThickness: 'flex',
+      maxBarThickness: 14,
+      minBarLength: 0,
+      stack: 'dob',
+      backgroundColor: chartStyles.ufoGreen,
+    },
+  ],
+}, data);
```
slaweet: There’s just one detail to fix: it should handle different lengths of dataset, because now you cannot get rid of the undefined.
Screenshot 2019-11-18 at 15 01 14


You can do something like this:
export const barChartData = data => merge({
  datasets: (data.datasets || []).map((_, i) => ({
      barPercentage: 1,
      categoryPercentage: 0.5,
      barThickness: 'flex',
      maxBarThickness: 14,
      minBarLength: 0,
      backgroundColor: primaryColorPalete[i],
    })),
}, data);

src/utils/chartOptions.js
```diff
@@ -171,47 +164,132 @@ export const barGraphOptions = options => merge({
...
+      borderColor: chartStyles.borderColor,
+      pointBorderColor: chartStyles.borderColor,
+      pointBackgroundColor: chartStyles.whiteColor,
+      pointHoverBackgroundColor: chartStyles.whiteColor,
+      pointHoverBorderColor: chartStyles.ultramarineBlue,
+      pointHoverBorderWidth: 4,
+      borderWidth: 2,
+    },
+  ],
+}, data);
```
slaweet: This needs change similar to the one explained on `barChartData`

src/utils/chartOptions.js
```diff
@@ -171,47 +164,132 @@ export const barGraphOptions = options => merge({
...
+      barThickness: 'flex',
+      maxBarThickness: 14,
+      minBarLength: 0,
+    },
+    {
+      barPercentage: 1,
+      categoryPercentage: 0.5,
+      barThickness: 'flex',
+      maxBarThickness: 14,
+      minBarLength: 0,
```
slaweet: 🎨  I think this shouldn't be 0, according to the design https://projects.invisionapp.com/d/main?origin=v7#/console/18539694/385859990/inspect?scrollOffset=1048

osvaldovega:
awesome fixed @slaweet

src/utils/chartOptions.js
```diff
@@ -0,0 +1,284 @@
...
+ * data ONLY for Bar chart
+ * @param {object} data - More data that can be pass to the chart
+ */
+export const barChartData = data => merge({
+  datasets: (data.datasets || []).map((_, index) => ({
+    barPercentage: 1,
+    categoryPercentage: 0.5,
+    barThickness: 'flex',
+    maxBarThickness: 14,
+    minBarLength: 1,
```
slaweet: 🎨 In the design it's:
```suggestion
    minBarLength: 2,
```

src/components/screens/wallet/balanceChart.js
```diff
@@ -1,63 +1,60 @@
...
 import { withTranslation } from 'react-i18next';
-import { Line as LineChart } from 'react-chartjs-2';
 import Box from '../../toolbox/box';
 import BoxHeader from '../../toolbox/box/header';
 import BoxEmptyState from '../../toolbox/box/emptyState';
-import styles from './balanceChart.css';
 import * as ChartUtils from '../../../utils/balanceChart';
 import { tokenMap } from '../../../constants/tokens';
 import i18n from '../../../i18n';
+import Chart from '../../toolbox/Chart';
```
slaweet: 💅 Our naming convention is to start with lower-case. To be updated in all places
```suggestion
import Chart from '../../toolbox/chart';
```

src/utils/chartOptions.js
```diff
@@ -0,0 +1,284 @@
...
+ */
+export const lineChartData = data => merge({
+  datasets: (data.datasets || []).map((_, index) => ({
+    backgroundColor: chartStyles.transparent,
+    borderColor: colorPallete[index],
+    pointBorderColor: colorPallete[index],
+    pointBackgroundColor: chartStyles.whiteColor,
+    pointHoverBackgroundColor: chartStyles.whiteColor,
+    pointHoverBorderColor: colorPallete[index],
+    pointHoverBorderWidth: 4,
```
slaweet: 🎨 Looking at the designs, there should be no point without hover
https://projects.invisionapp.com/d/main?origin=v7#/console/18805079/391946699/preview?scrollOffset=0
https://projects.invisionapp.com/d/main#/console/17570736/374296680/preview
```suggestion
    pointHoverBorderWidth: 4,
    pointRadius: 0,
    pointHoverRadius: 4,
```

src/utils/chartOptions.js
```diff
@@ -0,0 +1,284 @@
...
+    case typeBar: {
+      return barChartData(data);
+    }
+
+    case typeDoughnut: {
+      return doughnutChartData(data);
+    }
+
+    default:
+      return [];
```
slaweet: Not that it has some effect now, but for the future 🙂:
```suggestion
      return data;
```

osvaldovega:
I just pushed the last changes, let me know what do you think

osvaldovega:
@slaweet I was checking the code and I didn't found the error looks like there is a problem in jenkis wih the last update becase the paths were update the code is running fine, could be something related to cache problem.

I will do a minor change and push again and see if that works but all this done and working properly

slaweet:
I think the problem was due to case-insensitive filesystem on macOS and case-sensitive filesystem on Linux.

Now it works well, as well as everything else. Thank you, Osvaldo.

PR #2640, opened 2019-11-12
Add block ID search functionality to the search bar - Closes 2567

### LINKED ISSUE #2567: Add block ID search functionality to the search bar
### Description
We'd like to be able to search for block IDs in the main search field.

The search should use Lisk Core directly to get the results. 

[Design](https://trello.com/c/ECrRwH3b/128-add-a-template-for-block-id-as-the-result-of-search) information.

colado:
### What issue have I solved?

resolves #2567

Added block search functionality.
Searches are possible typing the whole block ID, as it is with the transactions.

### How has this been tested?
Made sure that block results on searches is in line with the other results.

### Review checklist


colado:
In order to get the `generatorUsername` as the [design](https://projects.invisionapp.com/share/MEUSP2H7RWA#/screens/393007389_Search_-_Block_ID_Hover) specifies, we need to use the endpoint from Service instead of the one from Core, as it is used on the other searches.
@reyraa let me know what your final preference is regarding this: whether to follow the design and get the data from the Service endpoint as it is now, or whether another approach is preferred, i.e. using the Core endpoint and instead showing another value, like `height`.

reyraa:
I've talk to Oliver regarding this, will get back to this topic ASAP.

reyraa:
@colado we concluded to remove the extra information and only display the block Id
https://projects.invisionapp.com/d/#/console/18869943/393007389/comments/124302327

Efefefef:
@reyraa And what should happen if I search for a block in devnet?

reyraa:
Now I state that monitor features are only available for mainnet and testnet. So users can search for a blockId, but it only redirect them to a page that displays the aforementioned message.

PS. better copy on the way.

reyraa:
Aligned the changes with designers.

PR #2641, opened 2019-11-13
Display the list of next forgers - Closes #2624

### LINKED ISSUE #2624: Display the list of next forgers in delegates monitoring
### Description
As a user; I'd like to see the list of next forgers according to [designs](https://projects.invisionapp.com/d/main/#/console/18539694/389182920/preview). 

### Acceptance criteria
- the list update every 10 sec.
- Total forged value should have thousand separators.

colado:
### What issue have I solved?

#2624

Added `ForgingDetails` component to `monitor/delegates` with data from next and last forgers, as well as total forged.

Design can be found [here](https://projects.invisionapp.com/d/main?origin=v7#/console/18539694/389184881/preview?scrollOffset=1276).

### How has this been tested?
Compare data with Lisk explorer to make sure results are accurate.

### Review checklist


src/components/screens/monitor/delegates/forgingDetails.js
```diff
@@ -0,0 +1,60 @@
...
+  // TODO: Check for alternative solution
+  const delegatesList = sortDirection.includes('asc') ? [...delegates.data] : [...delegates.data].reverse();
+  const totalForged = networkStatus.data.supply - initialLSKSupply;
+
+  const lastForger = [...delegatesList].sort(
+    (a, b) =>
+      b.lastBlock && a.lastBlock && b.lastBlock.height - a.lastBlock.height,
+  )[0];
+
+  const nextForgers = [...delegatesList]
```
slaweet: 💡 I think this would solve the glitch when loading the page:
```suggestion
  const nextForgers = [...delegatesList].filter(d => d.forgingTime)
```
The problem is that `forgingTime` comes from different api request thatn the list of delegates, so at some point it's undefined

src/components/screens/monitor/delegates/forgingDetails.css
```diff
@@ -0,0 +1,46 @@
...
+
+.contentWrapper {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-evenly;
+
+  & div {
+    flex-grow: 1;
+    flex-basis: 33%;
+  }
```
slaweet: 💡 I think this should be replaced with http://flexboxgrid.com/ so as not to add more work to https://github.com/LiskHQ/lisk-desktop/issues/2630

using flexboxgrid will also ensure there is padding between the columns in cases like this:
Screenshot 2019-11-14 at 16 48 16

colado: Good idea! Replaced with `flexboxgrid`.

src/components/screens/monitor/delegates/forgingDetails.js
colado: Thank you, Vit! I used your suggestion and it works better. There is a small delay but I think it's inevitable...

src/components/screens/monitor/delegates/forgingDetails.css
```diff
@@ -0,0 +1,37 @@
...
+
+  & .totalForged {
+    @mixin contentNormal;
+  }
+
+  & .contentBody {
+    @mixin contentNormal bold;
+
+    color: var(--color-ultramarine-blue);
```
slaweet: I thought the content is blue because it's links to delegates, which already are blue and bold, so this wouldn't be needed.

src/components/screens/monitor/delegates/forgingDetails.css
```diff
@@ -0,0 +1,37 @@
+@import '../../../../app/variables.css';
+@import '../../../../app/mixins.css';
+
+.contentWrapper {
+  & h3 {
+    font-weight: 600;
+    font-size: 13px;
+    line-height: 20px;
```
slaweet: 💅 This kind of stuff should IMO come from https://github.com/LiskHQ/lisk-desktop/blob/development/src/app/mixins.css

colado: ✅

src/components/screens/monitor/delegates/forgingDetails.css
colado: You're right. I replaced those with a `Link`. However I left the styling so that the `, ` wasn't included inside the link when hovering.

src/components/screens/monitor/delegates/forgingDetails.js
```diff
@@ -0,0 +1,67 @@
...
+    .filter(d => d.forgingTime)
+    .slice(0, 10)
+    .sort((a, b) => a.forgingTime - b.forgingTime);
```
slaweet: 🐛 It's not getting the correct delegates. It should first `sort` then `slice`.

src/components/screens/monitor/delegates/forgingDetails.test.js
```diff
@@ -0,0 +1,53 @@
...
+    networkStatus: {
+      data: {
+        supply: 9000000000000000000,
+      },
+    },
+    delegates: {
+      isLoading: true,
+      data: delegates.map((delegate, i) => ({
+        ...delegate,
+        forgingTime: i,
```
slaweet: 💡 To make it more realistic You could use:
```suggestion
        forgingTime: moment().add(i * 10, 'seconds'),
```
https://momentjs.com/docs/#/manipulating/add/
 and mayby also shuffle the whole array with `.sort(() => Math.random() - 0.5)`
https://javascript.info/task/shuffle

src/components/screens/monitor/delegates/delegates.test.js
```diff
@@ -53,7 +53,7 @@ describe('Delegates monitor page', () => {
   });
 
   it('renders a page with header', () => {
-    expect(wrapper.find('Box header')).toIncludeText('Active delegates');
+    expect(wrapper.find('Box header').at(1)).toIncludeText('Active delegates');
```
slaweet: 💅 I would rather add some class to the Box. Because this way, it will be again broken when we add the box with charts.
```suggestion
    expect(wrapper.find('Box.delegate-table header')).toIncludeText('Active delegates');
```

src/components/screens/monitor/delegates/forgingDetails.test.js
colado: Thank you, Vit. Much better

src/components/screens/monitor/delegates/delegates.test.js
colado: ✅

src/components/screens/monitor/delegates/forgingDetails.js
colado: ✅

reyraa:
@colado 
@Efefefef 
@slaweet 

- The heading border bottom is not correct.
- Breakpoints are fine as implemented.
- The data should update, every 10 sec obviously.
- I would keep it blank just like the others. Would be better if we had a placeholder template, but for now just keep it consistent (blank).

colado:
Changes are implemented except for the automatic update of data. As discussed with @reyraa, the approach to use is to be decided.

slaweet:
The lines look better, but now the box content doesn't have correct padding-right/left:
Screenshot 2019-11-18 at 18 10 23

colado:
You're right, @slaweet. Fixed the alignment.

src/components/screens/monitor/delegates/forgingDetails.js
```diff
@@ -0,0 +1,70 @@
...
+import BoxContent from '../../../toolbox/box/content';
+import styles from './forgingDetails.css';
+import LiskAmount from '../../../shared/liskAmount';
+import { tokenMap, initialLSKSupply } from '../../../../constants/tokens';
+import routes from '../../../../constants/routes';
+
+const ForgingDetails = ({
+  t, delegates, sortDirection, networkStatus,
+}) => {
+  const delegatesList = sortDirection && sortDirection.includes('asc') ? [...delegates.data] : [...delegates.data].reverse();
```
slaweet: 💅 I don't see the need for reversing the order here. `delegatesList` is being sorted again when used anyway.
```suggestion
  const delegatesList = delegates.data;
```

src/components/screens/monitor/delegates/forgingDetails.js
```diff
@@ -0,0 +1,70 @@
...
+  const totalForged = networkStatus && networkStatus.data.supply - initialLSKSupply;
+
+  const lastForger = [...delegatesList].sort(
```
slaweet: 💅 I don't see why this extra spread is needed
```suggestion
  const lastForger = delegatesList.sort(
```

src/components/screens/monitor/delegates/forgingDetails.css
```diff
@@ -0,0 +1,40 @@
+@import '../../../../app/variables.css';
+@import '../../../../app/mixins.css';
+
+.contentWrapper {
+  & > div:not(:last-child) {
```
slaweet: 💅 Using `div` here is against the point 15 or our css guide: https://github.com/LiskHQ/lisk-desktop/blob/development/docs/CSS_GUIDE.md

colado: `className` added

src/components/screens/monitor/delegates/withForgingStatus.js
```diff
@@ -79,6 +79,7 @@ const withForgingStatus = delegatesKey => (ChildComponent) => {
             nextHeight: height + i + 1,
           },
         }), {}),
+        nextForgersList: nextForgers.slice(0, 10),
```
slaweet: 🐛 Storing `nextForgersList` next to `nextForgers` means that `nextForgersList` is not being updated on new block, which is happening for `nextForgers` in `componentDidUpdate`

src/components/screens/monitor/delegates/withForgingStatus.js
```diff
@@ -199,6 +200,8 @@ const withForgingStatus = delegatesKey => (ChildComponent) => {
             ...this.props[delegatesKey],
             data: this.getDelegatesData(),
           },
+          lastBlock: latestBlocks[0],
```
slaweet: 🐛 Unfortunately, this works only when the page is loaded and `latestBlocks[0]` is from Lisk Service API, but once there's a new block received from  the Lisk Core socket, this one doesn't contain `generatorUsername`

slaweet: Just fixed with https://github.com/LiskHQ/lisk-desktop/pull/2641/commits/1176f1e021ffe4825834713046d8e2a7edc06f1f

src/components/screens/monitor/delegates/withForgingStatus.js
```diff
@@ -72,35 +72,46 @@ const withForgingStatus = delegatesKey => (ChildComponent) => {
...
+          nextHeight: height + i + 1,
+          username: delegate.username,
+        })),
       });
     }
 
     componentDidUpdate(prevProps) {
       const { latestBlocks } = this.props;
       const newBlock = latestBlocks[0] || {};
+      const forgerIndex = this.state.nextForgers.findIndex(delegate =>
```
slaweet: 💡 IMO the code (here and in `getUpdatedNextForgersList`) would be more straight-forward with
```suggestion
      const forger = this.state.nextForgers.find(delegate =>
```

src/components/screens/monitor/delegates/withForgingStatus.js
```diff
@@ -72,35 +72,46 @@ const withForgingStatus = delegatesKey => (ChildComponent) => {
...
+      const nextForgers = [...this.state.nextForgers];
+
+      if (forgerIndex !== -1) {
+        const forger = {
+          ...nextForgers[forgerIndex],
+          forgingTime: moment().add(voting.numberOfActiveDelegates * 10, 'seconds'),
+          nextHeight: newBlockHeight + voting.numberOfActiveDelegates,
+        };
+        nextForgers.push(forger);
+        nextForgers.splice(0, 1);
```
slaweet: 💡 `shift` would IMO be more straight-forward way to achieve the same:
```suggestion
        nextForgers.shift();
```

colado: Thank you for the suggestions, @slaweet. Changed

slaweet:
> Current forger in the list of next forgers

I think this is happening when a delegate misses to forge a block in their slot.
Because many (right now 26) delegates on testnet are not forging, this is happening quite often but still randomly.

Since delegate mising a block is something that can always happen, `getUpdatedNextForgersList` should take it into account
https://github.com/LiskHQ/lisk-desktop/blob/a3c60280bfd81ecf7f88c044aaba5a60d651cb41/src/components/screens/monitor/delegates/withForgingStatus.js#L98-L111

It has these implications:
- I should move to the end of the list not just `forger`, but also all delegates before `forger`.
- `forgingTime` and `nextHeight` needs to be recomputed for the moved forgers

It also looks like `nextHeight` is not used anymore and can be removed

slaweet:
When thinking about https://github.com/LiskHQ/lisk-desktop/issues/2663 I realized there is another problem to fix:

- 🐛 When nearing the end of the round, the list should show fewer delegates. E.g. if there are only 5 blocks left in the round, only 5 next forgers should be displayed. 

Note that if the next delegate misses to forge, then it should be removed from the next forgers' list, but the displayed list should still have the length of 5.

src/components/screens/monitor/delegates/forgingDetails.js
```diff
@@ -0,0 +1,62 @@
...
+import routes from '../../../../constants/routes';
+
+const ForgingDetails = ({
+  t, networkStatus, latestBlock, nextForgers,
+}) => {
+  const totalForged = networkStatus && networkStatus.data.supply - initialLSKSupply;
+  const remainingBlocksInRound = latestBlock && 101 - (latestBlock.height % 101);
+  const nextForgersInRound = nextForgers.filter(
+    (forger, i) => i < 10 && i < remainingBlocksInRound,
+  );
```
slaweet: :nail_care: I think it would be simpler code and less computation to replace  `nextForgers.filter` with something `nextForgers.slice(0, Math.min(10, remainingBlocksInRound));`

src/components/screens/monitor/delegates/withForgingStatus.js
```diff
@@ -72,35 +72,40 @@ const withForgingStatus = delegatesKey => (ChildComponent) => {
...
           this.loadNextForgers(latestBlocks);
         }
       }
     }
 
+    getUpdatedNextForgersList(forgerIndex) {
+      const nextForgers = [...this.state.nextForgers];
+      nextForgers[forgerIndex] = {
+        ...nextForgers[forgerIndex],
+        forgingTime: moment().add(voting.numberOfActiveDelegates * 10, 'seconds'),
```
slaweet: - 🐛 `forgingTime` needs to be updated not only for `nextForgers[forgerIndex]`, but also for all `nextForgers` before `forgerIndex` (if there are any). 

💡 So the update can be performed with `map` on `forgersToPush`, but then the `forgingTime` needs to depend on the index.

src/components/screens/monitor/delegates/withForgingStatus.js
```diff
@@ -188,6 +193,18 @@ const withForgingStatus = delegatesKey => (ChildComponent) => {
       }
     }
 
+    ensureGeneratorUsername(block) {
+      const generator = this.state.nextForgers.find(
```
slaweet: Now that `nextForgers` is a list instead of an object it would be nice for performance:
- 🐎 to avoid the `find` in case `generator` is not needed, that is if `block.generatorUsername` is defined.
- 🐎 prevent maping on all blocks below in this file if only first is used, something like replace:
```js
latestBlock: latestBlocks.map(this.ensureGeneratorUsername.bind(this))[0]
```
with
```js
latestBlock: latestBlocks.slice(0,1).map(this.ensureGeneratorUsername.bind(this))[0]
```

colado:
Thank you, @slaweet. Last changes implemented!

reyraa:
I close this in favor of #2681.

PR #2642, opened 2019-11-14
Create a dark theme poc - Closes #2579

### LINKED ISSUE #2579: Create a PoC for introducing dark theme using PostCSS
We'd like to introduce dark theme and in order to evaluate all aspects and effects of introducing such a massive change, we suggest to introduce changes using the current technology, PostCss, on the settings page.

yasharAyari:
### What issue have I solved?
It is described in #2579.

1. replace postcss-cssnext with postcss-preset-env
2. add dark theme colors to variables.css
3. implement theme context 
4. implement withTheme HOC
5. Add ThemeContext.Provider to app component
6. Implement the dark mode version of all pages

### How has this been tested?
Make sure the dark mode color scheme is the same as final design on invision. 

### Review checklist


src/components/screens/settings/settings.js
```diff
@@ -138,6 +138,19 @@ class Settings extends React.Component {
...
+              <label className={`${styles.fieldGroup} ${styles.checkboxField}`}>
+                <CheckBox
+                  name="DarkMode"
+                  className={`${styles.checkbox} DarkMode`}
+                  checked={settings.DarkMode}
+                  onChange={this.handleCheckboxChange}
+                />
+                <div>
+                  <span className={styles.labelName}>Dark Mode</span>
+                  <p>Hide balance and transactions amounts</p>
```
slaweet: 🐛 The copy doesn't seem right

slaweet:
I fixed the e2e tests: https://github.com/LiskHQ/lisk-desktop/commit/adafda8fd0cfd73cdbb36e8643cb0f3569a020be but I see there was a reason for the `overflow: auto;` so that the dark background goes all the way to the bottom when scrolling. Can we fix this in a follow-up ticket?

test/cypress/features/send.feature
```diff
@@ -9,6 +9,7 @@ Feature: Send
     And I go to transfer confirmation
     And I confirm transfer
     And I go back to wallet
+    And I dismiss the onboarding
```
slaweet: 💡 I would consider hiding the onboarding by setting `walletOnboarding = true` to the `localStorage` in the cypress setup:
https://github.com/LiskHQ/lisk-desktop/blob/adafda8fd0cfd73cdbb36e8643cb0f3569a020be/test/cypress/support/commands.js#L36-L40

to keep the feature file simpler. Because without knowing the problem we had today, I don't really see why this step is needed.

Efefefef: I agree but Yashar has a different vision on e2e tests

PR #2654, opened 2019-11-18
Remove custom solution for cypress github check - Closes #2645

### LINKED ISSUE #2645: Replace custom e2e Github status check to Cypress provided one
In order to see the status of the e2e tests right from the PR, we've implemented the custom Github check.
Now as Cypress-Github integration was enabled we can use one that is provided.

Efefefef:
### What issue have I solved?

Resolves #2645 

In case all Cypress tests pass with no failure we won't see the github check at all ( as we are not using paid features for the main run)
In case some of them fail and we rerun tests with 'recording' we will see either succesfull or errored github check

### Review checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,9 +93,6 @@ pipeline {
 						}
 					},
 					"cypress": {
-						githubNotify context: 'Jenkins e2e tests',
-							     description: 'e2e tests in progress...',
-							     status: 'PENDING'
 						dir('lisk') {
 							checkout([$class: 'GitSCM',
 							          branches: [[name: "v${env.LISK_CORE_VERSION}" ]],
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -198,21 +187,6 @@ EOF
 				  onlyStable: false,
 				  sourceEncoding: 'ASCII'
 			junit 'coverage/jest/junit.xml'
-			script {
-				catchError {
-					if(readFile(".cypress_status").trim() == '0'){
-					    status = 'SUCCESS'
```
fchavant: Don't we need to keep this to set the Jenkins build status based on the outcome of cypress?

Jenkinsfile
```diff
@@ -141,20 +135,15 @@ EOF
...
+										if [ $ret -ne 0 ]; then
 										  FAILED_TESTS="$( awk '/Spec/{f=1}f' cypress.log |grep --only-matching '✖ .*.feature' |awk '{ print "test/cypress/features/"$2 }' |xargs| tr -s ' ' ',' )"
-										  cd $WORKSPACE/$BRANCH_NAME
-										  make coldstart
-										  export CYPRESS_coreUrl=http://127.0.0.1:$( docker-compose port lisk 4000 |cut -d ":" -f 2 )
-										  sleep 10
-										  cd -
-										  npm run cypress:run -- --record --spec $FAILED_TESTS |tee cypress.log
-										  ret=$?
-										  grep --extended-regexp --only-matching 'https://dashboard.cypress.io/#/projects/1it63b/runs/[0-9]+' cypress.log |tail --lines=1 >.cypress_url
```
fchavant: https://github.com/LiskHQ/lisk-desktop/issues/2645 only mentions removing the github notification but not the "2 stages" run

Efefefef: I moved it into the changed if instruction `if [ $ret -ne 0 ]; then` so it's there but in then instead of else

Jenkinsfile
Efefefef: I think t's only for Cypress check status. We now don't need it. See updated PR desc

PR #2665, opened 2019-11-22
 Fix monitor to work for Custom mainnet/testnet nodes - Closes #2664

### LINKED ISSUE #2664: Monitor with mainnet/testnet custom node stopped working
### Steps to reproduce

1. Set custom node, which is actually on testnet or mainnet (based on `nethash`)  Screenshot 2019-11-22 at 13 07 12
2. Go to Monitor

### Expected behavior
See data:
Screenshot 2019-11-22 at 13 03 42

### Actual behavior
See message that monitor is not available:
Screenshot 2019-11-22 at 13 05 38

Current `development` branch after merging https://github.com/LiskHQ/lisk-desktop/pull/2640

slaweet:
### What issue have I solved?

Resolves #2664

Updated the monitor pages to reuse logic from `liskService.js` to check if lisk service is available or not. It is using `getNetworkNameBasedOnNethash`.

### How has this been tested?

Follow steps to reproduce in #2664

### Review checklist

### DIFF
--- a/src/utils/api/lsk/liskService.js
+++ b/src/utils/api/lsk/liskService.js
@@ -125,6 +125,14 @@ const liskServiceApi = {
     return tabOptions[tab](network, rest);
   },
 
+  getLiskServiceUrl: (networkConfig) => {
+    try {
+      return getServerUrl(networkConfig);
+    } catch (e) {
+      return null;
+    }
+  },
+
   getNextForgers: async ({ networkConfig }, searchParams) => liskServiceGet({
     networkConfig,
     path: '/api/v1/delegates/next_forgers',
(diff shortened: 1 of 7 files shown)


src/utils/api/lsk/liskService.js
```diff
@@ -125,6 +125,14 @@ const liskServiceApi = {
...
   },
 
+  getLiskServiceUrl: (networkConfig) => {
+    try {
+      return getServerUrl(networkConfig);
+    } catch (e) {
+      return null;
+    }
+  },
+
```
reyraa: My comment is not limited to this particular method, but documenting inputs and outputs is recommended habit because it helps other developers to know what to pass and what to expect. even for testing, it makes the review process easier.
In addition, can we make sure the output of the function has a unique type. in this case we don't need to perform complex checks when using the output in conditional statements.

slaweet: I added jsDoc to this function. Please check it and let me know what you think.

reyraa: Looks good. Thanks

PR #2686, opened 2019-12-03
 Add data size to the details of a single transaction - Closes #2514

### LINKED ISSUE #2514: Add data size to the details of a single transaction
As a use; I'd like to see the size of the data field in bytes. This information must be visible in transaction details page next to the message itself.

reyraa:
### What issue have I solved?

Resolves #2514

- Added the data size next to message
Extra: Removed the steps label in registration process since designers complained about it :-)

### Review checklist

### DIFF
--- a/src/components/screens/explorer/transactions/transactionDetailView/transactionDetailView.js
+++ b/src/components/screens/explorer/transactions/transactionDetailView/transactionDetailView.js
@@ -1,5 +1,6 @@
 import { withTranslation } from 'react-i18next';
 import React from 'react';
+import { sizeOf } from '../../../../../utils/helpers';
 import AccountInfo from './accountInfo';
 import BoxRow from '../../../../toolbox/box/row';
 import TransactionTypeFigure from '../../../wallet/transactions/typeFigure/TransactionTypeFigure';
@@ -76,11 +77,17 @@ class TransactionDetailView extends React.Component {
         {children}
         { transaction.type === transactionTypes.send
           && (transaction.asset && transaction.asset.data) ? (
-            <BoxRow>
-              <div className={styles.detailsWrapper}>
+            <BoxRow className={styles.message}>
+              <div className={`${styles.detailsWrapper}`}>
                 <span className={styles.label}>{t('Message')}</span>
-                <div className={`${styles.message} tx-reference`}>
+                <div className={`${styles.value} tx-reference`}>
                   {transaction.asset.data}
+                </div>
+              </div>
+              <div className={`${styles.detailsWrapper}`}>
+                <span className={styles.label}>{t('Size')}</span>
+                <div className={`${styles.value} tx-size`}>
+                  {`${sizeOf(transaction.asset.data)} bytes`}
                 </div>
               </div>
             </BoxRow>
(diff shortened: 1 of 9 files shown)


src/components/screens/explorer/transactions/transactionDetailView/transactionDetailView.js
```diff
@@ -76,13 +77,19 @@ class TransactionDetailView extends React.Component {
         {children}
         { transaction.type === transactionTypes.send
           && (transaction.asset && transaction.asset.data) ? (
```
slaweet: - 🐛 Message and Size is not displayed if the message is empty

src/utils/helpers.js
```diff
@@ -73,3 +73,42 @@ export const flattenArray = arr =>
     (Array.isArray(item)
       ? [...acc, ...flattenArray(item)]
       : [...acc, item]), []).filter(item => !!item);
+
+// eslint-disable-next-line
```
slaweet: - 💅 please don't disable all eslint rules, but only the specific rule. Now it's not clear at all what is disabled and why is it needed and makes room for introducing more eslint violations.

src/utils/helpers.js
```diff
@@ -73,3 +73,42 @@ export const flattenArray = arr =>
...
+
+  if (typeof object === 'string') return object.length * sizes.string;
+  if (typeof object === 'boolean') return sizes.boolean;
+  if (typeof object === 'number') return sizes.number;
+  if (Array.isArray(object)) {
+    return object.map(sizeOf).reduce((acc, curr) => acc + curr, 0);
+  }
+  if (typeof object === 'object') return sizeOfObject(object, sizeOf);
+  return 0;
+};
```
slaweet: - 💅 AFAIK the data object of transaction type 0 is always a string. So I don't understand why you implemented all this extra logic for other data types that just ads dead code to our code base 🤔

reyraa: I just wasn't sure if it's gonna remain string. else we would call it "message" instead of data.

slaweet: AFAIK, there is no plan to change it in Lisk Core. I think it's called `data` just because that's how it was requested from the community in https://github.com/LiskHQ/lisk-sdk/issues/26 and nobody gave any thought to naming. 

I still don't see a good reason to keep all this extra logic in the repository, because one day it might be useful.

src/components/screens/explorer/transactions/transactionDetailView/transactionDetailView.js
reyraa: Resolved

src/utils/helpers.js
reyraa: agreed
I'll remove the other parts

src/utils/helpers.js
```diff
@@ -73,3 +73,15 @@ export const flattenArray = arr =>
...
+
+/**
+ * Returns the size of a given string in bytes
+ * If the passed parameter is not a string, it returns 0
+ *
+ * @param {string} str - a random string
+ * @returns {number} - string size for strings, 0 for others
+ */
+export const sizeOfString = (str) => {
+  if (typeof str === 'string') return str.length * 2;
```
slaweet: 🐛 The size is shown double the size
```suggestion
  if (typeof str === 'string') return str.length;
```

You can see it e.g. on this transaction https://testnet-explorer.lisk.io/tx/10611085484582831767

reyraa: This logic was wrong. we already had it in Send and Request components. I updated the utility and reused in those components

PR #2690, opened 2019-12-05
Fix the remainig issues with the dark mode - Closes #2682

### LINKED ISSUE #2682: There are still some visual issues in dark mode
### Expected behavior
Dark mode looks as in the designs https://projects.invisionapp.com/share/MTTUMGYWQD5#/screens/383200745

### Actual behavior
There are issues described in https://github.com/LiskHQ/lisk-desktop/pull/2671#pullrequestreview-326023636

reyraa:
### What issue have I solved?

Resolves #2682

- Fixed the issues with verify message.
- Fixed the delegate rank color in wallet page header.
- Fixed the canvas of all new illustrations to match those of the light ones we already had. and then removed the styles I created in my previous PR to overcome the differences.
- Improved the theme determination logic based on @slaweet 's suggestion.

Plus, I've created a ticket to fix the issues of the charts. #2692

### Review checklist


src/components/screens/verifyMessage/verifyMessage.css
```diff
@@ -18,3 +18,9 @@
     vertical-align: middle;
   }
 }
+
+.result {
+  & h1 {
+    color: var(--color-maastricht-blue);
```
slaweet: 💡  I was thinking if there is a better way to ensure that all headlines have a contrasting color to the background. 

I remember there were a lot of places where the headline was dark on dark by default, e.g.:
https://github.com/LiskHQ/lisk-desktop/blob/35a84736c4e405ea7a38762d79b2ebf9f8789270/src/components/screens/monitor/blocks/blocksOverview/blocksOverview.css#L4-L6


Maybe could be solved by setting `color: var(--color-maastricht-blue);` in the same place as where the dark background it set. In 
https://github.com/LiskHQ/lisk-desktop/blob/35a84736c4e405ea7a38762d79b2ebf9f8789270/src/components/toolbox/box/box.css#L7

or 
https://github.com/LiskHQ/lisk-desktop/blob/35a84736c4e405ea7a38762d79b2ebf9f8789270/src/app/app.css#L46

reyraa: It's good idea. I'll create a ticket to fix it after the beta release

slaweet:
@imdominic this PR should be ready for Design review

reyraa:
@slaweet Covered the remaining parts. Working on the dark mode feature, I noticed in many tests we have before hooks where we render the component, while we never use it. and this is not detected programmatically because we declare a `let` variable in the top most `describe` and assign mounted components later. so there's always at least one `it` statement in which we assigned and used the mounted component. I suggest we always define it as `const` inside the `it` statement to ensure this done only once. Concerning the code readability, if we define the assumptions such as `accountSignedIn`, `accountSignedInUsingHW`, `accountNotSingedIn` with self descriptive naming convention outside the test logic, then I assume the tests should remain neat.

PR #2714, opened 2019-12-20
Solve warnings in console caused by React-toolbox - Closes #2479

### LINKED ISSUE #2479: Solve warnings in console caused by React-toolbox
### Current behaviour
When I go to http://localhost:8080/ and open the browser console, there are warnings coming from [react-toolbox](https://www.npmjs.com/package/react-toolbox).  React-toolbox is causing warnings like: `Please update the following components:  ThemedRippledComponent, ThemedTBButton, ThemedThemed, ThemedProgressBar`
Screenshot 2019-09-23 at 14 14 47

### Expected behaviour
When I go to http://localhost:8080/ and open the browser console, there are no errors or warnings.

The problem is that this dependency doesn't have a newer version, so we need to decide if we want to replace it with other library or write the needed components ourself. We use `react-toolbox` for the following:
```sh
$ git grep "from 'react-toolbox"
src/components/toaster/toaster.js:import Snackbar from 'react-toolbox/lib/snackbar';
src/components/toolbox/buttons/button.js:import { Button as ToolBoxButton } from 'react-toolbox/lib/button';
src/components/toolbox/progressBar/progressBar.js:import ToolBoxProgressBar from 'react-toolbox/lib/progress_bar';
src/components/toolbox/timestamp/index.js:import Tooltip from 'react-toolbox/lib/tooltip';
src/components/toolbox/timestamp/index.js:import theme from 'react-toolbox/lib/tooltip/theme.css';
```

This is a follow-up to https://github.com/LiskHQ/lisk-hub/pull/2477

slaweet:
### What issue have I solved?

Resolves #2479

- [x] Remove React-toolbox dependency from Button components
- [x] Remove React-toolbox dependency from ProgessBar component
- [x] Remove React-toolbox dependency from Timestamp component
- [x] Remove React-toolbox dependency from Toaster component
- [ ] Migrate Toaster component to https://github.com/fkhadra/react-toastify

### How has this been tested?

Click through the whole app and check for regressions in the aforementioned components.

### Review checklist


slaweet:
@yasharAyari I was looking at multiple toast libraries https://npmcompare.com/compare/react-toast-component,react-toast-notifications,react-toaster,react-toaster-js,react-toastify and react-toastify is ranked the best, but I don't think it's worth to use it in our case because:

- It doesn't look as our style guide (border-radius, close icon, show animation, and much more), so we'll still need our custom toast component with all our styles.
- It has a lot of options/features we don't need, so it adds unnecessary size to our js bundle
- It has its own way of dispatching toasts so it means refactoring all 30 (counted by `git grep -l oastDisplayed | wc -l`) files in our code that dispatch a toast, which means a lot of changes that are not needed for the goal of this PR - get rid of the warnings in the console.

yasharAyari:
1. you can customize the style of the toaster you can find information in the blew link :
https://github.com/fkhadra/react-toastify#le-style
2. The only thing that we don't need is the progress bar that you can easily disable it
3. the change you mentioned is a necessary change. Because you can implement the same behavior with the above library without involving the redux store.

yasharAyari:
I have talked with Julian about the animation. he said we can replace the current animation of the toaster with the `react-toastify` animation.

slaweet:
1. I missed that option,
2. There's a lot more we won't use (just look at the home page demo https://fkhadra.github.io/react-toastify/): "All Positions expect bottom-right", "Disable auto-close", "Newest on top*", "Close on click", "Pause delay on hover", "Right to left layout", "Pause toast on visibility change", "Allow to drag and close the toast".
3. Migrating from redux is something to consider, but not a necessary part to reach the goal of this PR.

I still think that migrating to `react-toastify` is not in the scope of this PR. Let's get a third opinion. What do you think @reyraa?

reyraa:
@slaweet we discussed and compared the animations and features you mentioned with Julian. The react tostify library seems a better option. Please use it. Thanks.

reyraa:
sonarcloud detects a code smell. But the button mentioned if defined after the function expression, and expression does get hoisted, there's no problem with that chunk of code.

PR #2716, opened 2020-01-02
Implement new blocks and transactions buttons - Closes #2510 #2513

### LINKED ISSUE #2510: Implement the load new blocks button
### Description
We'd like to display a button on top of the list when there are new blocks to be shows. In order to not distract users, we don't want to update the list automatically.

### Acceptance criteria
- The button should slide down from beneath the list header when there are new blocks.
- The button should slide up to its initial state when the user clicked on it.
- The new blocks attach to the top of the list with no animation.

### LINKED ISSUE #2513: Implement the load new transactions button
### Description
We'd like to display a button on top of the list when there are new transactions to be shows. In order to not distract users, we don't want to update the list automatically.

### Acceptance criteria
- The button should slide down from beneath the list header when there are new transactions.
- The button should slide up to its initial state when the user clicked on it.
- The new transactions attach to the top of the list with no animation.

slaweet:
### What issue have I solved?

Resolves #2510
Resolves #2513

I decided to handle both tickets in one PR because once the button is implemented for Blocks, adding the button for Transactions is just 10 lines in https://github.com/LiskHQ/lisk-desktop/blob/2510-implement-load-new-blocks-button/src/components/shared/transactionsTable/index.js#L81-L90 so a separate PR would IMO be overkill.

- [x] Create `useServiceSocketUpdates` hook for listening to new blocks
- [x] Create the `LoadLatestButton` shared component that uses `useServiceSocketUpdates` hook
- [x] Implement styling of the `LoadLatestButton` component to match [design](https://projects.invisionapp.com/d/main#/console/18906983/396861464/preview)
- [x] Use `LoadLatestButton` on Blocks monitor 
- [x] Use `LoadLatestButton` on Transactions monitor 

Note that there are the following deliberate divergencies from design:
- Font-size of the button is 15px, like any other Primary button of Size L. (The design has 13px.)
- The button is placed between the box header and table header. (In design, it is between table header and table content.)
 I think this way it looks fine and the code is cleaner as I don't have to pass the button inside the toolbox Table component.

### How has this been tested?

- Go to Block Monitor
- Wait for blocks to be forged so that the button appears
- Click the button to get the new block displayed

### Review checklist


reyraa:
- The button pushes the list down, it's distracting. Specially on the blocks page which happens every 10 seconds. Some times right after you click on it, it appears again. Although not mentioned in the acceptance criteria, I think it makes sense to skip updates for at least 10 seconds once the user clicked. All for the original purpose of "not distracting the user".
- Also the button is unnecessarily wide. I asked Dom to revise this.
- We should prepend the new block not reload the list. the current implementation displays the loading overlay and blocks the list which imo is not desired. open for discussion though.

PR #2744, opened 2020-01-27
Import 2 different versions of lisk-element - Closes #2737

### LINKED ISSUE #2737: Import 2 versions of Lisk Element
### Description
We need to be able to install and import version 2.x and 3.x of `lisk-elements` library to be able to interact with those two versions of `lisk-core` when is required.

yasharAyari:
### What issue have I solved?

Resolves #2737

1. Update my npm version to 6.10.0 to be able to use npm aliases
2. Install version 3.x as lisk-client-Edge
3. Store the API version of lisk-core endpoint on networkset action
4. Create a helper which desice of lisk-client needs to be used based on the API version of redux store
5. Import lisk-client form the above helper instead of npm package
6. Create a webpack alias called 'Util' for the ease of use in the development process 

### Review checklist


reyraa:
@yasharAyari Looks good in general. I merge this so we can move forward. We'll fix the broken tests afterwards.

PR #2748, opened 2020-01-29
Align the issue template across the team

sridharmeganathan:
### What issue have I solved?

Issues templates are not aligned with other repos


sridharmeganathan:
The requested is not required. The current template is matching with required format.

PR #2777, opened 2020-02-12
Refactor delegates screen - Resolves #2774

### LINKED ISSUE #2774: The delegate search bar is hidden
### Expected behavior
This search bar is used to ease the voting process. It should be visible/usable at all times.

### Actual behavior
The search bar is hidden and can't be used.

### Steps to reproduce
- Navigate to the delegates page.

1.25.0-beta.0

reyraa:
### What was the problem?
This PR resolves #2774

### How was it solved?
Why did I hide the input in the first place?
This page had a bug since the first implementation: If you voted to standby delegates, and you didn't click on the load more button to load sufficient number of delegates, you wouldn't see all of your votes on the Voted tab. 
This issue stemmed in the component's trying to map votes on the existing delegates. Since the existing delegates didn't include the delegates to which you have voted, you couldn't see all your votes.
I tried to fix it, but since the page was refactored several times, it was tricky. clearly we needed to refactor this page.
Here's how I proceed:
-  I read the list of delegates and votes and store them on the Redux store.
- I only filter to delegates based on the tab and never mutate the list, so your votes and the list of delegates are always kept in Redux store, no matter the page or tab your on.

WIP.

### How was it tested?
- Delegates and votes should look fine on all tabs.
- You should be able to search through the list.
- If you load more delegates, change the tab and change the tab back, you should see the same number of delegates.
- you should be able to proceed to the voting summary page, and press on Edit Votes and see your previously selected votes.
- You should be able to vote using a launch protocol.


sridharmeganathan:
Tested the functionality and it is working fine.  Assigned other developer for the code review.

PR #2780, opened 2020-02-17
Fix the color of search bar results - Closes #2735

### LINKED ISSUE #2735: Darkmode: when user input block id information in generic search (Desktop header), the preview of search panel is not visible enough to end user
In dark mode in generic search, when user input the block id, the preview of search panel is not visible enough to end user

reyraa:
### What was the problem?
This PR resolves #2735

### How was it solved?
The color of the result value was not defined. so I used `maastricht-blue` which is used everywhere else.

### How was it tested?
Search for a block id in dark and light mode. the result should be readable.


sridharmeganathan:
Reviewed the functionality. It is working as expected

PR #2781, opened 2020-02-17
Align the colors of request form with the design guides - Closes #2726

### LINKED ISSUE #2726: Sharing link field design alignment in light mode
### Expected behavior

The field should look inactive (not editable), as of the user is not able to edit it but just copy the link by clicking on the action below. Currently does in the dark mode but not in the light mode.

Same as above applies to BTC, as of you are not able to edit the address that is there.

Inactive Share link field

### Actual behavior

The field looks like the ones above, so editable, when is not.

Screenshot 2020-01-10 at 17 10 22

### Steps to reproduce

Log in with your account, navigate to wallet, request LSK (same applies to BTC)

1.23.0

reyraa:
### What was the problem?
This PR resolves #2726

### How was it solved?
The background color of the testarea in light mode was defined `transparent`. I used `white-smoke` as requested.

### How was it tested?
Make sure the background color of the textarea on the request panel aligns with the design guides.

### DIFF
--- a/src/app/variables.css
+++ b/src/app/variables.css
@@ -141,7 +141,7 @@ or "warn/action" ineastd of "red/green"
   --color-strong-mystic: var(--color-mystic);
   --color-passphrase-bg: var(--color-white);
   --color-shadow-word-option: var(--color-ultramarine-blue);
-  --color-request-textarea: transparent;
+  --color-request-textarea: var(--color-white-smoke);
 
   /*************************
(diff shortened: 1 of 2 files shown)


reyraa:
@slaweet I had a similar idea as well. Couldn't figure out which textarea can have an inactive state. I can't think of a scenario in which an input can be inherently inactive.

PR #2782, opened 2020-02-17
Navigate to the wallet page when clicking on asset from menu - Closes #2723

### LINKED ISSUE #2723: Navigate to the wallet page when clicking on asset from menu
### Expected behavior

Navigate to the wallet page of the specific asset when clicking on the it from the dropdown menu on the top right corner. This applies also to the same token, if you are on delegates and from the dropdown you click on the asset you are already in, you will land on the wallet page.

### Actual behavior

Navigates to the same page where you are currently but in the other asset.

### Steps to reproduce

Sign in with an account - Click on the top right corner - Select a different asset or eventually the same one (redirects you to the wallet page if you are not already there)

1.23.0

reyraa:
### What was the problem?
This PR resolves #2723

### How was it solved?
1. I navigate to Wallet right after the token change.
2. For the sake of consistency, if the user tries to navigate to a path which is not available for the active token, say if clicked on a Vote launch protocol while BTC is active, I make it fall back to Wallet.

### How was it tested?
Change the token, it should redirect you to Wallet.

### DIFF
--- a/src/components/shared/customRoute/customRoute.js
+++ b/src/components/shared/customRoute/customRoute.js
@@ -20,7 +20,7 @@ const CustomRoute = ({
   Piwik.tracking(rest.history, settings);
 
   if (forbiddenTokens.indexOf(settings.token.active) !== -1) {
-    return <Redirect to={`${routes.dashboard.path}`} />;
+    return <Redirect to={`${routes.wallet.path}`} />;
   }
 
   return ((isPrivate && isAuthenticated) || !isPrivate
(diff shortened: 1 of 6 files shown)


sridharmeganathan:
Can we apply the same style to BTC request also ? Not sure it is valid

image

reyraa:
@slaweet I fixed the issue with e2e tests

sridharmeganathan:
Verified. Working as expected.

PR #2785, opened 2020-02-17
Support Lisk development networks - Closes #2784

### LINKED ISSUE #2784: Support Lisk development networks
### Description
Lisk Desktop can run againts:
- Mainnet
- Testnet
- Betanet
- Alphanet
- Devnet
- Local nodes

As a user I should be able to connect to the corresponding Service nodes automatically.

reyraa:
### What was the problem?
This PR resolves #2784

### How was it solved?
We had a utility to automatically connect to mainnet, testnet, betanet and nodes running on port 4000, but the devOps team have introduces new URL schema with https support. 
Here I refactored to the utility to suppoort the new schema as well.

### How was it tested?
Make sure Desktop can connect to the nodes defined in the issue #2784.

### DIFF
--- a/src/utils/api/lsk/liskService.js
+++ b/src/utils/api/lsk/liskService.js
@@ -12,52 +12,51 @@ import { adaptTransactions } from './adapters';
 import transactionTypes from '../../../constants/transactionTypes';
 
 const isStaging = () => (
-  localStorage.getItem('useLiskServiceStaging') || version.includes('beta') || version.includes('rc')
-    ? '-staging' : ''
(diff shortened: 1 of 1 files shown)


reyraa:
@yasharAyari I absolutely agree. As we predicted, every now and then there's a new url schema added, to support which we need a new update. I agree that we should add a second input so users can enter their Lisk Service URL. but this can't be done unless other developers have access to Lisk Service, e.g. it's open source. 
until then, let's keep it as implemented.

PR #2787, opened 2020-02-17
Fix desktop notifications to work with Lisk Core 3.0 - Closes #2779

### LINKED ISSUE #2779: Push notification is not working in betanet
### Expected behavior
Push notification message should be sent whenever the user receives tokens from others
Push notification should work only when user receive new tokens

### Actual behavior

- [ ] No push notification message received but my balance getting updated. This issue is only in Betanet and it is working in Testnet. 
- [ ] User who is voting delegates and user who is registering as a delegate is receiving push notification (0 tokens) which is incorrect. Refer the below image for reference
image

Betenet.

slaweet:
### What was the problem?
This PR resolves #2779 

The problem was that the transactions were not transformed by `txAdapter`.

### How was it solved?

Ensure that the logic for notification uses transaction object transformed by `txAdapter` from `/src/utils/api/lsk/adapters`

### How was it tested?

- Build the desktop app from this branch
- Open the app and connect to https://betanet.lisk.io and login to an account
- Send transactions to/from this account and see a notification is/isn't displayed as expected based on the transaction type

### DIFF
--- a/src/store/middlewares/account.js
+++ b/src/store/middlewares/account.js
@@ -95,11 +95,10 @@ const checkTransactionsAndUpdateAccount = (store, action) => {
   const { transactions, settings: { token } } = state;
   const account = getActiveTokenAccount(store.getState());
 
-  const txs = action.data.block.transactions || [];
+  const txs = (action.data.block.transactions || []).map(txAdapter);
   const blockContainsRelevantTransaction = txs.filter((transaction) => {
-    const morphedTx = txAdapter(transaction);
-    const sender = morphedTx ? morphedTx.senderId : null;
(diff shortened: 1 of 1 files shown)


reyraa:
@slaweet 
 - Built the app.
 - Connected to betanet
 - Minimized the app.
- Used the faucet to transfer tokens
 > No notifications received.

Repeated the same process with another account, as active window and in minimized mode. still no notification.

Are my test steps connect?

slaweet:
@reyraa I just checked again, following your steps (both minimized and active window) and it works for me. Do notifications work for you with testnet?

PR #2801, opened 2020-02-20
Use next_forgers to fetch the list of active delegates - Closes #2800

### LINKED ISSUE #2800: Retrieve the active delegates list from /forgers
### Expected behavior
The list of active delegates are fetched from `/delegates/forgers` which is accurate at any time.

### Actual behavior
The list of active delegates are fetched from `/delegates` which can be inaccurate if an standby delegates joins the list of active delegates.

1.25.0-beta.0

reyraa:
### What was the problem?
This PR resolves #2800

### How was it solved?
We used the `/delegates` endpoint to fetch the first 101 items using `limit=101` parameter. but we should have used `/delegates/next_forgers`.

Note:
At the moment this endpoint limits the number of delegates to 100, while we need 101. @MichalTuleja will update the Service endpoint to accept 101.

### How was it tested?
Make sure the active delegates list on the monitor section has 101 items and it matches the output of `/delegates/next_forgers`.

### DIFF
--- a/src/utils/api/lsk/liskService.js
+++ b/src/utils/api/lsk/liskService.js
@@ -124,7 +124,7 @@ const liskServiceApi = {
 
   getActiveDelegates: async ({ networkConfig }, { search = '', tab, ...searchParams }) => liskServiceGet({
     networkConfig,
-    path: '/api/v1/delegates/active',
+    path: '/api/v1/delegates/next_forgers',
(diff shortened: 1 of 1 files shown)


MichalTuleja:
The status of the API change can be tracked here:
https://github.com/LiskHQ/lisk-service/issues/461

PR #2809, opened 2020-03-02
Remove language selector - Closes 2713

### LINKED ISSUE #2713: Remove the language selector on the settings page
### Description
We'd like to remove the Language selector in the settings page while maintaining the infrastructure. 

### Acceptance Criteria
 - The language selector in the settings page is removed.
 - Remove the local files.

reyraa:
### What was the problem?
This PR resolves #2713

### How was it solved?
Removed the language selector. Lisk Desktop falls back to English if it was set to Deutsch before.

### How was it tested?
 - Tests should pass.
 - Language selector must not exist on Settings.
 - If the app was set to Deutsch, running this version should set it to English.


slaweet:
Based on reading the code, I suspect that the OS-level menu (see example in the screenshot below) will stay in German in the new version if the user set german language before. 

But `npm run dist` fails on my machine, so I’m unable to check.

Screenshot 2020-03-03 at 14 53 00

PR #2821, opened 2020-03-12
Display connected nodes - Closes #2819

### LINKED ISSUE #2819: Display network statistics (only connected peers)
### Description
Display the list of connected peers in the existing blockchain

### Motivation
Provide users with realtime stats on network status
https://projects.invisionapp.com/share/9WUY7BIJ452#/screens/395159458

### Acceptance Criteria

- [x] Adapt to the UX design
- [x] List  the connected peers with IP, port, location, version  and current block height details
- [x] Sort the list by Height & version. By default, the list should be sorted by height (descending)
- [x] Display the connected peers based on location ==> Map View.  Map should have Zoom in & Zoom out options

reyraa:
### What was the problem?
This PR resolves #2819

### How was it solved?
Fetch and display the list of connected nodes.

### How was it tested?
- Compared with design guides.
- Added unit tests.

Known differences with the design guides:
- I've put the zoom button on the top left instead of top right. The amount of added code base wasn't worth it.
- The proportion of the flag is not the same as designs, I couldn't find package that looks exactly the same as in designs.
- Map colors are different. I'm using an open map, which is not as configurable as Google maps.


reyraa:
@yasharAyari This works fine on my local, looks like the images don't get loaded properly using the production build. I'll fix it.

src/components/screens/monitor/network/flag.js
```diff
@@ -0,0 +1,109 @@
...
+  dz: 'algeria',
+  am: 'armenia',
+  at: 'austria',
+  az: 'azerbaijan',
+  bs: 'bahamas',
+  bh: 'bahrain',
+  bd: 'bangladesh',
+  bb: 'barbados',
+  be: 'belgium',
+  bj: 'benin',
```
slaweet: Why not use the ISO 3166-1 alpha-2 two-letter codes already in the `src/components/screens/monitor/network/flag.css` file? I suppose the flag.css file is generated by some tool? I think there should be a tool out there to already use ISO 3166-1 alpha-2 as the class names. 

This would bring less complexity and more countries supported out-of-the-box.

reyraa: This was an auto-generated dictionary. Since it's an object and doesn't cause big-o overload, I didn't see the need to spend time on it.

slaweet: For the computational complexity, it's fine. :+1: 

What is the requirement of how many countries should we support?

Now, this file supports about 90 countries, but AFAIK there are almost 200 countries in the world and the flag.css seems to cover all of them.

reyraa:
@yasharAyari the markers are now visible

reyraa:
@slaweet 
I made some minimal changes in the appearances of the zoom control. 
Fixed the sorting by version.
Left the initial height sort as it is.
and made Network the default tab.

sridharmeganathan:
1. Zoom out is not working as expected.  When you are continuously selecting the Zoom out, at one point we can see only the number of connected peers, the map is getting disappeared 

2. The cumulative  connected peers are incorrect when we zoom in.

PR #2823, opened 2020-03-18
Fix table load more button - Closes #2669

### LINKED ISSUE #2669: Monitoring: Load more button is not shown after several fetchings
### Expected behavior
Load more is always there
### Actual behavior
It is not shown after several uncertain number of loadings
### Steps to reproduce
1. Go to monitoring
1. Blocks
1. Click Load more until you cant see the button
1.23.0

reyraa:
### What was the problem?
This PR resolves #2669 

### How was it solved?
- Remove the %30 condition
- Control the button in parent components

### How was it tested?
Manual and using unit tests

### DIFF
--- a/src/components/screens/monitor/blockDetails/blockDetails.js
+++ b/src/components/screens/monitor/blockDetails/blockDetails.js
@@ -72,6 +72,10 @@ const BlockDetails = ({
     [id]: <LabeledValue label={label} className={styles.dataContainer}>{value}</LabeledValue>,
   }), {});
 
+  const canLoadMore = blockTransactions.meta
+    ? blockTransactions.data.length < blockTransactions.meta.total
+    : false;
+
   return (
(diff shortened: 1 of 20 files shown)


src/components/screens/monitor/blockDetails/blockDetails.js
```diff
@@ -125,6 +125,7 @@ const BlockDetails = ({
         title={t('Transactions')}
         transactions={blockTransactions}
         emptyStateMessage={t('There are no transactions for this block.')}
+        canLoadMore={false}
```
slaweet: It's an edge case, but the genesis block (height=1) has more than 30 transactions, so it might be worth to have that covered

reyraa: Thanks. I forgot about genesis block. Fixed it.

PR #2843, opened 2020-03-30
Display latest votes on delegates monitor page - Closes #2657

### LINKED ISSUE #2657: Display the list of latest votes
### Description
As a user; I'd like to see the latest votes in a dedicated tab.

### Motivation
- Help delegates have a better understanding about changes in the list of their voters.
- Help voters get notified if the community is revoking trust from a delegate or increasing attention towards another.

### Acceptance Criteria
According to designs on [InVision](https://projects.invisionapp.com/share/BQWIWV1C5A9#/screens/410025094/comments):

- This should lay under the third tab on delegates page.
- This feature is available only for mainnet, testnet & betanet  just like other monitoring features.
- The data is not live.

reyraa:
### What was the feature?
This PR resolves #2657

### How was it solved?
Added the list of latest votes.


slaweet:
I noticed that
- the table headers are misaligned on a smaller breakpoint
Screenshot from 2020-03-31 09-37-44
- loading bar missing
- somehow the latest votes view adds horizontal scroll to the whole page, even though there is just empty space on the right.

slaweet:
I also noticed something in the new design of next forgers since this is the first time I saw it in action. I know it's not related to this ticket/PR, so just mentioning it.

At first look, it's not clear that each of "Next forgers" is a link. I discovered it only by the change of cursor on hover, but even then the hover area is quite big (no margin left/right between hover area of two forgers) and when hovering in between two forgers, it's not clear if clicking will go to the one on the left or the one on the right.

src/components/screens/monitor/delegates/latestVotes/voteRow.js
```diff
@@ -0,0 +1,71 @@
...
+        />
+      </span>
+      <span className={grid['col-sm-2']}>
+        <DateTimeFromTimestamp time={data.timestamp * 1000} token="BTC" />
+      </span>
+      <span className={grid['col-sm-2']}>
+        <LiskAmount val={data.balance} token="LSK" />
+      </span>
+      <span className={`${grid['col-lg-1']} ${gridVisibility['hidden-md']}  ${gridVisibility['hidden-sm']} ${gridVisibility['hidden-xs']}`}>
+        <span>{data.height}</span>
```
slaweet, line 40: The table header says this value should be "round", not "height". AFAIK round can be computed like this:

```suggestion
        <span>{Math.ceil(data.height / 101)}</span>
```

reyraa:
Sorry @slaweet , I accidentally removed your comment. but your feedback are already applied. Thanks for the review.

package.json
```diff
@@ -66,6 +66,7 @@
     "electron-localshortcut": "3.1.0",
     "express": "4.16.4",
     "flexboxgrid": "=6.3.1",
+    "flexboxgrid-helpers": "1.1.3",
```
slaweet, line 69: This is a great helper to have on the project :+1:

src/components/screens/monitor/delegates/latestVotes/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Table from '../../../../toolbox/table';
+import VotesRow from './voteRow';
```
slaweet, line 3: This component is defined as `VoteRow` but in this file referred to as `VotesRow`. It can be changed here
```suggestion
import VoteRow from './voteRow';
```

Or in the other file, but both should use the same name

jest.config.js
```diff
@@ -91,6 +91,8 @@ module.exports = {
     'src/components/screens/monitor/delegates/delegateRow.js',
     'src/components/screens/monitor/delegates/delegates.js',
     'src/components/screens/monitor/network/map.js',
+    'src/components/screens/monitor/delegates/latestVotes',
+    'src/components/screens/monitor/delegates/delegatesTable/index.js',
```
slaweet, line 95: I'd expect some comment (here or in PR description) explaining why the test coverage is ignored. E.g. there is a follow-up ticket to write the tests, or some other reason.

reyraa: There's not much to it. These tests are extremely dull. I'll cover them later on when I have more time. plus, there's an epic for adding e2e tests soon.

PR #2873, opened 2020-05-04
Fix balance tooltip values and lables - Closes #2871

### LINKED ISSUE #2871: Data mismatch in the wallet page lisk graph
The  balance chart is missing data for X-axis and the tool tip in line graph is showing Jan 2020 always

image

reyraa:
### What was the problem?
This PR resolves #2871

### How was it solved?
- The tooltip only showed dates in January. We have updated the data and forgot to update the moment format parameters. I fixed it by updating the format string.
- Some xAxis labels are skipped. I fixed by removing redundant options that caused this problem.
- @sridharmeganathan Since we talked about using k and M signs for numbers in thousand and million range, this is not possible. Because most transaction values are significantly lower than account balances, such formatting would make the Y axis labels confusing. Assume an account has 7,987500 LSK, and the last 30 transaction are all under 100 LSK. If I shorten the balance to show 7.9k, then the graph looks like:

7.9k |...x.....................................................
7.9k |................x................................x......
7.9k |...........................x............x................
7.9k |_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.........Mon......Tue.......Wed.......Thu....Fri.....

As you can see the yAxis gives us nothing.

### How was it tested?
Visually by comparing several accounts.

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -14,7 +14,7 @@
   "Accept": "Accept",
   "Access your account by scanning the QR code below with the Lisk Mobile App:": "Access your account by scanning the QR code below with the Lisk Mobile App:",
   "Account": "Account",
-  "Account Balance:          {{balance}} {{token}}": "Account Balance:          {{balance}} {{token}}",
+  "Account Balance": "Account Balance",
   "Account name": "Account name",
   "Account nickname": "Account nickname",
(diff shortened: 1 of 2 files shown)


reyraa:
@sridharmeganathan please read the notes above your comment.

PR #2879, opened 2020-05-07
Fix issues discovered on Network Minitor page - Closes #2872

### LINKED ISSUE #2872: Data mismatch in Network monitoring page
There is a data mismatch when i compare the lisk-desktop networking monitoring page with lisk explorer monitoring page.

- Versions distribution & Height distribution should be sorted from current to previous
- Versions distribution & Height distribution tooltip is not fully visible.
- Connected peers and disconnected peers can be changed.
- Connected peers grid should be sorted by height (descending)

image

image

image

reyraa:
### What was the problem?
This PR resolves #2872 

### How was it solved?
Fixed the tooltip hiding under the map. Furthermore:

1. Added tooltipts to explain the information of this page only reflects the peers connected to the Lisk Service node.
2. Rename connected / disconnected peers to active / inactive.
3. Renamed Peers to Discoverable peers.
4. Renamed Connected peers to Active peers.
Screenshot 2020-05-11 at 10 44 04


reyraa:
@MichalTuleja 
@karmacoma 
@ishantiw

please take a look and verify the terminology changes. Thanks.

ishantiw:
LGTM, just one point, if we have already mentioned it in the tooltip that this page only reflects the peers connected to the Lisk Service node, we can keep connected/disconnected peers instead of active/inactive (unless service is going to provide a network-wide view of the nodes by fetching status of all the known peers in the network, @MichalTuleja ?)

reyraa:
@ishantiw sounds reasonable. I'll change it if @MichalTuleja is ok.

MichalTuleja:
It looks good assuming that we use the following naming convention:
- **connected/disconnected** - for real time data, returns only partial information about the network and can give an impression that some nodes are "disappearing" for a while,
- **active/inactive** - only if Lisk Service collects node information and returns data based on some interval. For example last 5 minutes and the node was visible most of the time and is classified as active. When the node availability falls down below let's say 30% threshold, it can be classified as inactive.

ishantiw:
> It looks good assuming that we use the following naming convention:
> 
> * **connected/disconnected** - for real time data, returns only partial information about the network and can give an impression that some nodes are "disappearing" for a while,
> * **active/inactive** - only if Lisk Service collects node information and returns data based on some interval. For example last 5 minutes and the node was visible most of the time and is classified as active. When the node availability falls down below let's say 30% threshold, it can be classified as inactive.

I agree with the naming convention, 
so in the screenshot of this issue, `Discoverable peers` correspond to the second naming convention `active/inactive` assuming that there is a service to provide that info?

MichalTuleja:
That would require a modification of Lisk Service. So for now it seems that `connected/disconnected` is the only proper name.

reyraa:
@ishantiw @MichalTuleja to conclude 
- Change `Discoverable peers` to `Connected peers` on graph title.
- Change `Active` and `Inactive` legends in the same graph to `connected` and `disconnected`.
- Change `Active peers` on the table title to `Connected peers`.

Just to clear out, this is basically reverting to how it was before, except for the tooltips. Correct?

ishantiw:
@reyraa Sounds good. It was my suggestion that either we change the naming convention to a more network-wide view (but for that we need extra services as @MichalTuleja stated above) or we simply mention it somewhere that this is only Lisk Service view of the network. 
So reverting to how it was before, except for the tooltips carrying info that this is Lisk Service node's view ✅

PR #2885, opened 2020-05-13
Fix the total forged value on Delegates Monitor - Closes #2884

### LINKED ISSUE #2884: The total forged value on Delegates Monitor is not correct
### Expected behavior
The value shown as total forged value on Delegates Monitor page must calculated based on the values forged by all delegates since the beginning. 

### Actual behavior
The value doesn't match.

### Steps to reproduce
Navigate to delegates monitor page.
The initial supply is 1M and the total supply is 39.6M. This doesn't match with the value shown.

reyraa:
### What was the problem?
This PR resolves #2884

### How was it solved?
I take the initial supply a constant equal to 100M LSK
and fetch the total supply from the Lisk Service `network/status` endpoint.
the total forged is displayed as `total supply` minus `initial supply`.

### DIFF
--- a/src/components/screens/monitor/delegates/forgingDetails.js
+++ b/src/components/screens/monitor/delegates/forgingDetails.js
@@ -7,6 +7,7 @@ import routes from '../../../../constants/routes';
 import Box from '../../../toolbox/box';
 import BoxHeader from '../../../toolbox/box/header';
 import styles from './overview.css';
+import LiskAmount from '../../../shared/liskAmount';
 
 const Forger = ({ forger }) => (
   <div className={`${styles.forger} forger-item`}>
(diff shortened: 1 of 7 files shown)


src/components/screens/monitor/delegates/forgingDetails.js
```diff
@@ -22,8 +24,11 @@ const Forger = ({ forger }) => (
 );
 
 const ForgingDetails = ({
-  t,
+  t, networkStatus,
 }) => {
+  const initialSupply = 10000000000000000;
```
MichalTuleja, line 29: This might make the value misleading when used with a different blockchain than mainnet. I've checked the Core API and there is no endpoint telling us what was the initial supply.

Consider making it hardcoded, based on NetworkId/nethash.

reyraa: I thought the initial supply is the same for all networks.

reyraa: The testnet supply is 139,576,098 and the mainnet supply is 139,630,112. The only difference according to Manu is 
> mainnet is burning a lot faster and consistent than testnet.

PR #2940, opened 2020-07-06
Fix visual UI Issues found during review - Closes #2935 

### LINKED ISSUE #2935: Visual issues discovered testing the new layout
### Description
The following visual issues are discovered during the testing phase:

- Change the headings font weight. (e.g. for buttons, addresses and balance should be used the regular font).
- Consolidate Lisk addresses in a single line.
- Remove the Bitcoin logo at the address block, there should be only address.
- Filters icon is broken.
- Buttons at the address block must be aligned with the “Send LSK” button.
- Place the loader spinner at the top of the block, now it crosses the tabs.  
- The old view of the “Load more” button, change to the new one.
- The _Add to bookmark_ button should open the form on the main bookmark modal.

UsamaHameed:
### What was the problem?
This PR intends to resolve the issues explained in #2935.

- [x] Change the headings font weight. (e.g. for buttons, addresses and balance should be used the regular font).
partially done. there are some visual issues with the font though. the normal font weight looks like its bold.
- [x] ~~Consolidate Lisk addresses in a single line.~~
- [x] Remove the Bitcoin logo at the address block, there should be only address.
- [ ] Filters icon is broken
investigated the issue and found that the icon is broken when resized with relative units.
- [x] Buttons at the address block must be aligned with the “Send LSK” button. 
- [x] Place the loader spinner at the top of the block, now it crosses the tabs.
- [x] The old view of the “Load more” button, change to the new one.
- [x] The Add to bookmark button should open the form on the main bookmark modal.

### How was it solved?
I made minor adjustments. 

### How was it tested?
Tested visually. I have not been able to run the unit tests though. All of them fail; seems like a setup issue to me.

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -241,7 +241,7 @@
   "Lisk {{version}}": "Lisk {{version}}",
   "Lisk {{version}} is here!": "Lisk {{version}} is here!",
   "Lisk's blockchain is based on the Delegated Proof of Stake (DPoS) consensus algorithm, in which 101 delegates are voted in by LSK holders to secure the network.": "Lisk's blockchain is based on the Delegated Proof of Stake (DPoS) consensus algorithm, in which 101 delegates are voted in by LSK holders to secure the network.",
-  "Load more": "Load more",
+  "Load more": "Load More",
   "Loading": "Loading",
   "Loading...": "Loading...",
   "Locale": "Locale",
--- a/src/components/screens/bookmarks/bookmarksList/bookmarksList.css
+++ b/src/components/screens/bookmarks/bookmarksList/bookmarksList.css
@@ -109,7 +109,6 @@
         & > button {
           min-width: initial;
           padding: 0 10px;
-          color: var(--color-white);
         }
       }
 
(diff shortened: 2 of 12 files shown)


reyraa:
Reviewed your changed, Looks good. Thank you.
Found some minor issues though:
- The balance info box doesn't seem ok, the Send button drops outside the box.
- (Additionally) Please put a white space between balance number and unit. 
- Here I think we have to make the Address smaller so it's consolidated in max 2 lines. `n1uapxtw4UhoPSEcXwcMjDpsRfGMoqyJoG` overflows to the third lines, which breaks the buttons alignment. Maybe we should make the address title smaller in general, or do it according to the address length. See which one is better.
- @ann-vashchenko  does the load more button has a hover animation?

UsamaHameed:
[x] The balance info box doesn't seem ok, the Send button drops outside the box.
[x] (Additionally) Please put a white space between balance number and unit.
[x] Here I think we have to make the Address smaller so it's consolidated in max 2 lines. n1uapxtw4UhoPSEcXwcMjDpsRfGMoqyJoG overflows to the third lines, which breaks the buttons alignment. Maybe we should make the address title smaller in general, or do it according to the address length. See which one is better.

PR #2961, opened 2020-07-15
Implement new routing changes - Closes #2956

### LINKED ISSUE #2956: Implement rendering of UI Components through routes and search params
### Description

We would like to have the flexibility of defining the UI using routes and search params in the url. 

### Motivation

This approach makes it easier to define and control the UI that the client sees. It also makes it possible to share specific states of UI using links. 

### Acceptance Criteria

- Change the current routing system and add options like `modals` and `active tabs` to the search params. These options should be extendable so that in the future we can easily add new options. 
- Whenever a relevant search param is present, that UI component should be rendered. 
- Define a way to limit the components that can be rendered through the url in every specific route.

UsamaHameed:
### What was the problem?
This PR resolves #2956 

### How was it solved?
Implement new routing system

### How was it tested?
Unit tests


src/constants/routes.js
```diff
@@ -156,6 +155,7 @@ export const modals = {
     component: VotingSummary,
     isPrivate: true,
     forbiddenTokens: [tokenMap.BTC.key],
+    allowedOnlyOnPages: ['voting'],
```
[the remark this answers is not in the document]
reyraa: In general, our routes (default export) must have the following routes
```
/wallet/transactions
/wallet/voting
/wallet/delegate-profile
/wallet/multi-signature // upcoming
/accounts/overview
/accounts/:id
/transactions/overview
/transactions/:id
/blocks/overview
/blocks/:id
/delegates/overview
/delegates/stand-by
/delegates/latest-votes
/network
/termsOfUse
/login/passphrase
/login/hw
/register
/dashboard

File: src/store/reducers/searchParams.js [hunk 2]
```diff
@@ -0,0 +1,30 @@
+import actionTypes from '../../constants/actions';
+import { parseSearchParams } from '../../utils/searchParams';
+
+
+const getInitialState = () => {
+  const { modal } = parseSearchParams(window.location.search);
+  return { modal };
```
reyraa, line 7: the `location` object has this value already and is available throughout the app, why persist on Redux store? seems redundant.

src/components/shared/customRoute/index.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
 import { Redirect, Route } from 'react-router-dom';
+
```
reyraa, line 4: Remove the extra blank line

UsamaHameed: I added to add a line between the _node_module_ imports and _local_ imports. What do you think about that?

reyraa: We didn't use to do that. but it seems a good idea. keep it please.

PR #3004, opened 2020-08-12
fix max blocks forged numbers and time

### LINKED ISSUE #3002: Incorrect round status in delegates monitoring page
### Expected behavior
In delegates page, the round status should show XX/101. 

### Actual behavior
Round status is showing XX/103 which is incorrect. We forge 101 blocks per cycle.

1.27.0

UsamaHameed:
### What was the problem?
This PR resolves #3002 

### How was it solved?
Fix the hardcoded values

### How was it tested?
NA

### DIFF
--- a/src/constants/delegates.js
+++ b/src/constants/delegates.js
@@ -1,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const MAX_BLOCKS_FORGED = 101;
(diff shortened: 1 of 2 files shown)


src/constants/delegates.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const MAX_BLOCKS_FORGED = 101;
```
MichalTuleja, line 2: This is correct for the version 3 of Lisk Core. The v4 and the v5 are going to forge 103 blocks per round.

UsamaHameed: I think yes. We are going to change this constant again when we start using lisk core v4. As far as I know, there is no api that returns this number.

MichalTuleja: That's true, there is no API for that. It can be hardcoded for now, but Lisk Service can provide that value in the future.

PR #3014, opened 2020-08-18
Allow user to bookmarks the account numbers from wallet page directly - Closes #2980

### LINKED ISSUE #2980: Allow user to bookmarks the account numbers from wallet page directly
### Expected behavior
User should access the 'Add bookmarks' window directly from wallet page instead of navigating through the search bookmarks window and also the account number should be pre-filled 

### Actual behavior
Users are accessing 'Add bookmarks' page through search bookmarks window

1.27.0 rc

isalga:
### What was the problem?
This PR resolves #2980 
### How was it solved?
- Fix failing api call
- Autofill form with query params
- Decouple bookmark modals
- Prevent from adding repeated bookmarks

### How was it tested?
manual testing


src/components/screens/bookmarks/addBookmark/addBookmark.js
```diff
@@ -156,19 +158,22 @@ class AddBookmark extends React.Component {
...
-      token: active,
-      account: {
-        title: label.value,
-        address: address.value,
-        isDelegate: !!(delegate && delegate.username),
-        publicKey,
-      },
-    });
+
+    if (!bookmarks[active].some(bookmark => bookmark.address === address.value)) {
```
isalga, line 166: @reyraa as I saw that it was possible to add multiple times the same address to bookmarks I added this code to prevent it from happening. I assume that it is not an expected feature, but let me know if it was intended to work that way.

src/components/screens/bookmarks/addBookmark/addBookmark.js
```diff
@@ -67,23 +72,36 @@ class AddBookmark extends React.Component {
...
+    };
+
+    if (account.data.delegate !== prevProps.account.data.delegate) {
+      if (account.data.delegate && account.data.delegate.username !== label.value) {
+        const data = { value: account.data.delegate.username, readonly: true };
+        newFields = { ...newFields, label: data };
+      } else if (label.readonly) {
+        newFields = { ...newFields, label: { value: '', readonly: false } };
+      }
     }
```
reyraa, line 97: This part of the code the overly imperative and the conditional statements don't seem to cover all of the possible branches. Please review.

reyraa (changes requested):
Hi Iris. 
Thanks for implementing the feature. There are one thing to consider, You are adding a new modal under `addBookmark` but you're not removing it from `bookmarks`, hence causing code/logic duplication. This is against DRY.
Please remove the `MultiStep` from `bookmarks` and keep your new modal the only screen to add/edit bookmarks.
Only use the route query params to pre-fill the bookmark form. 
Also please reinstate the button on the Send transaction status screen, to add/remove account to/from bookmarks.

src/components/screens/bookmarks/addBookmark/addBookmark.js
reyraa: Thanks for detecting the issue. I think introducing a function to check if a bookmark exist is more readable than a conditional statement.

reyraa (approved):
Looks good. Thank you

ann-vashchenko (approved):

PR #3019, opened 2020-08-19
Fix crash when access bookmarked accounts - Closes #2981

### LINKED ISSUE #2981: An error occurred when i access bookmarked accounts
An error occurred when i access bookmarked account number. This is happening in random scenarios.

isalga:
### What was the problem?
This PR resolves #2981 
When accessing a bookmarked regular(non delegate) account, from a previous bookmarked delegate account, the view 
was trying to render the delegate tab but as the account was not delegate the code was crashing when accessing `delegate.data.username` at `delegateProfile.js` because data property was `undefined`.

### How was it solved?
Prevent `delegate.data` from being `undefined` setting an empty object.

### How was it tested?
Opening bookmark menu and then open a delegate account, after that repeat the same process but opening a regular account. Repeat that alternating from delegate-to-regular and regular-to-delegate.


sridharmeganathan (dismissed):
Great job. Working as expected

src/components/screens/wallet/delegateProfile/index.js
```diff
@@ -1,47 +1,113 @@
...
-  },
-  nextForgers: {
-    apiUtil: getNextForgers,
-    defaultData: [],
-    autoload: true,
-    getApiParams: () => ({
-      limit: 101,
-    }),
-  },
+const delegateProfile = ({
```
UsamaHameed, line 14: Can you please move this component into its own file and export the component with the `withTransaction` HOC from the index file?

UsamaHameed (changes requested):

UsamaHameed (approved):

PR #3027, opened 2020-08-25
Display other accounts balance chart when discreet mode is on - Closes #2987

### LINKED ISSUE #2987: The balance chart is not visible in discreet mode for other accounts
The balance chart is not visible in discreet mode for other accounts. The balance chart should be visible to others user when the discrete mode is ON.

isalga:
### What was the problem?
This PR resolves #2987

### How was it solved?
Wallet overview was passing discreet mode to balance chart indistinctly of which account it was displaying.
I added a condition that just makes that prop true when the host address and the request address are the same.

### How was it tested?
Manual test


reyraa (approved):
Looks good. Thank you

PR #3036, opened 2020-08-28
Fix translation on request modal & Show only QR code on other accounts - resolves #3009

### LINKED ISSUE #3009: Request LSK functionality is missing in the new layout.
Request LSK functionality is missing in the new layout. 

image

1.27.0

isalga:
image
### What was the problem?
Translation where not properly passed to the Request component. Resolves issues with #3009 

### How was it solved?
Import and use withTranslation

### How was it tested?
Manual test

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -371,6 +371,7 @@
   "Save changes": "Save changes",
   "Save it on an encrypted hard drive: USB key or a backup drive": "Save it on an encrypted hard drive: USB key or a backup drive",
   "Save your passphrase": "Save your passphrase",
+  "Scan address": "Scan address",
   "Search by name or address": "Search by name or address",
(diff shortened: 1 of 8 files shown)


UsamaHameed (approved):

UsamaHameed (changes requested):
We also need to show the sign in tooltip when the user is not logged in. @isalga @sridharmeganathan

reyraa (approved):
Looks good. Thanks

UsamaHameed (approved):

sridharmeganathan (changes requested):
It is not working as expected and also without sign-in the functionality is not working.

sridharmeganathan (approved):
Functionality is working as expected

reyraa (approved):
Looks good. Thanks

PR #3038, opened 2020-08-31
Reset voting mode status - Closes #3015

### LINKED ISSUE #3015: Confirm button in voting page has to be reset after voting confirmation
### Expected behavior
Confirm button in voting page has to be reset after voting confirmation

### Steps to reproduce
Step 1: Go to voting page and select the edit voting button
Step 2: Select or deselect delegates votes
Step 3: Confirm votes
Step 4: Review and Confirm selected votes
Step5: View the summary page, but the confirm button (edit mode) is not changed into edit votes

isalga:
### What was the problem?
This PR resolves #3015 

### How was it solved?
Adding an effect that sets voting mode to false when a vote has been submitted.

### How was it tested?
Manual test&Adding a unit test


src/components/screens/voting/index.test.js
```diff
@@ -46,16 +52,35 @@ describe('Delegates', () => {
...
+        history: {
+          ...mockStore.history,
+          location: {
+            search: '?modal=votingSummary&isSubmitted=true',
+          },
+        },
+      },
+    );
+
+    expect(wrapper.find('.start-voting-button')).to.have.lengthOf(1);
```
UsamaHameed, line 78: You can use `expect(wrapper).toContainMatchingElement(/matcher/)` instead

UsamaHameed (changes requested):

sridharmeganathan (changes requested):
Edit voting button is not working.

image

src/components/screens/voting/index.test.js
isalga: For doing so we should refactor the whole unit test file from using expect from chai to the default expect. I think that would be contra-productive to do it now. What do you think?

sridharmeganathan: I checked the functionality. It is working as expected. I will approve once @UsamaHameed approves the code review

UsamaHameed (approved):

sridharmeganathan (approved):
Working as expected

PR #3041, opened 2020-09-02
Fix broken e2e tests - Closes #3040

### LINKED ISSUE #3040: E2E tests fail due to the dynamic fee changes
### Expected behavior
All e2e tests should pass as before introducing the dynamic fee.

### Actual behavior
13 tests pass and rest fail. Looks like the db records are different than expected by tests.

### Steps to reproduce
Run e2e tests.

reyraa:
### What was the problem?
This PR resolves #3040

### How was it solved?
Updated e2e tests
Updated unit tests


UsamaHameed (approved):

Jenkinsfile
```diff
@@ -15,7 +15,7 @@ pipeline {
 		stage('Install npm dependencies') {
 			steps {
 				nvm(getNodejsVersion()) {
-					sh 'npm install --registry https://npm.lisk.io'
+					sh 'npm install --registry https://npm.lisk.io --no-optional'
```
UsamaHameed, line 18: Should we remove the `no-optional` flag?

UsamaHameed (approved):

Jenkinsfile
reyraa: I think we can keep it. it is intended exactly for the browser.

PR #3045, opened 2020-09-04
Add bookmark after transaction

### LINKED ISSUE #2977: Re-enable the bookmark button when the modal routing is fixed
### Actual behavior
I have temporary disabled the bookmark button on the transaction status until we fix the modal routing issue. Since atm we can not route to the add bookmark modal, having this button on the tx status view is pointless.

### Expected behavior
We should re-enable this button and the corresponding unit test after the routing is fixed.

isalga:
### What was the problem?
This PR resolves #2977 

### How was it solved?
Add bookmark button when transaction is done and recipient is neither bookmarked or own address

### How was it tested?
Unit test & manual test

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -67,7 +67,6 @@
   "Blocks forged": "Blocks forged",
   "Blocks overview": "Blocks overview",
   "Blog": "Blog",
-  "Bookmarked": "Bookmarked",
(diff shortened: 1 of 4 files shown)


reyraa (approved):
Looks good. Thank you

UsamaHameed (approved):

PR #3048, opened 2020-09-07
Add tabs routing support - Closes #2993

### LINKED ISSUE #2993: Enable routing on screen tabs
### Description
As a user I'd like to see the delegate profile tab when I search for a delegate or click on a delegate on Delegates page. This should be the default behavior for all tabs.

### Motivation
Enable easy access to different sets of information on each screen.

### Acceptance Criteria
Users must be able to:
- Click on an account on accounts screen and see the transaction history of the account.
- Click on an delegate on delegates screen and see the delegate profile of the account.
- Search for a username and choose a result and see the delegate profile of the account.
- Search for an address and choose a result and see the transaction history of the account.

isalga:
### What was the problem?
This PR resolves #2993 

### How was it solved?
Replace the way activeTab is set on tabsContainer for a solution that uses URL search parameters

### How was it tested?
Update unit test, manual test

### DIFF
--- a/src/components/toolbox/switcher/index.js
+++ b/src/components/toolbox/switcher/index.js
@@ -13,7 +13,7 @@ const TabLink = withRouter(({
   };
 
   return (
-    <li onClick={onClick} ref={linkEl} className={className}>{ children }</li>
+    <li onClick={onClick} ref={linkEl} className={className} data-value={to}>{ children }</li>
   );
 });
 
@@ -27,7 +27,6 @@ const Switcher = ({ options, active }) => (
             tab.className,
           ].filter(Boolean).join(' ')}
           key={tab.value}
-          data-value={tab.value}
           to={tab.value}
         >
           {tab.name}
--- a/src/components/toolbox/switcher/index.js
+++ b/src/components/toolbox/switcher/index.js
@@ -1,22 +1,37 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useRef } from 'react';
+import { withRouter } from 'react-router';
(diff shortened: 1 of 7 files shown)


src/components/toolbox/switcher/index.js
```diff
@@ -1,22 +1,36 @@
...
-        <li
+        <TabLink
           className={[
             tab.value === active && styles.active,
             tab.className,
           ].filter(Boolean).join(' ')}
           key={tab.value}
-          data-value={tab.value}
-          onClick={onClick}
+          to={tab.value}
```
UsamaHameed, line 30: It would be better if we use an `id` to identify the tab instead of the tab name (tab.value). 
With the above implementation, the routes will look like: `...?tab=Delegate%20profile`
Instead, it would be better to have: `...?tab=delegateProfile`

UsamaHameed, line 30: It would be better if we use an id to identify the tab instead of the tab name (tab.value).
With the above implementation, the routes will look like: ...?tab=Delegate%20profile
Instead, it would be better to have: ...?tab=delegateProfile

UsamaHameed (changes requested):

src/components/toolbox/switcher/index.js
isalga: It does not exist an id property. Only 'name' and 'value' are attributes of 'tab'

UsamaHameed: I think it would be valuable to add a new property to the `tab` object

reyraa (approved):

PR #3056, opened 2020-09-10
Enable users to add delegates to vote queue - Closes #2973

### LINKED ISSUE #2973: Enable user to add delegates to vote queue
### Description
As a user I should be able to navigate to a delegate profile page and add the delegate to my vote queue by dedicating a vote weight.

### Motivation
Enable user to contribute in safekeeping the network.

### Acceptance Criteria
- Users must be able to add delegate by clicking on a single button in delegate profile page.  [Refer the design](https://www.figma.com/proto/yy39tEdAaYoVjl6LaqgpyH/Prototype?node-id=787%3A246&viewport=-636%2C-268%2C0.13392853736877441&scaling=min-zoom)
- Users must be presented by information about vote weight.
- Users must be able to define a vote weight which is less than the available balance of their account. [Refer the design](https://www.figma.com/proto/yy39tEdAaYoVjl6LaqgpyH/Prototype?node-id=1329%3A0&viewport=-636%2C-268%2C0.13392853736877441&scaling=min-zoom)
- After adding the vote weight, the vote weight with delegate name should be added in the voting queue. [Refer the design](https://www.figma.com/proto/yy39tEdAaYoVjl6LaqgpyH/Prototype?node-id=469%3A621&viewport=-636%2C-268%2C0.13392853736877441&scaling=min-zoom)
- The Add to votes button must be disabled if the user has already voted/added to votes the maximum of 10 delegates.
- If a delegate is added to votes, the add to votes button should turn to remove from votes.

### Additional info

- Add vote weight to votes/[username] in Redux store.
- Reduce the max vote cap to 10.

reyraa:
### What was the problem?
This PR resolves #2973

### How was it solved?
- Updated voting reducer. Since we don't have to show delegates which we haven't voted for, I removed delegates array from voting. I've also made some adjustments to add two vote amounts
   - confirmed: The vote amount stored on the blockchain.
  - unconfirmed: The vote amount defined by the user but not submitted to the blockchain.
- Update voting actions.

### How was it tested?
- Updated unit tests.


isalga (changes requested):
I think it would be good to add a unit test to check that no more than 10 votes can be added to the state.

src/components/screens/wallet/votes/voteRow.js
```diff
@@ -27,7 +26,6 @@ const VoteRow = ({
         <div className={styles.accountInfo}>
           <span className={`${styles.title} vote-username`}>{data.username}</span>
           <span className="showOnLargeViewPort">{data.address}</span>
-          <span className="hideOnLargeViewPort">{data.address.replace(regex.lskAddressTrunk, '$1...$3')}</span>
```
isalga, line 30: Why is this line removed? This will cause that the address won't be visible on small screensizes.

src/components/screens/transactionDetails/transactionVotes.test.js
```diff
@@ -29,14 +29,14 @@ describe('Transaction Votes', () => {
...
-  it('Should render with added and deleted Votes', () => {
+  it.skip('Should render with added and deleted Votes', () => {
     wrapper = mount(<TransactionVotes {...props} />);
     expect(wrapper).toContainMatchingElements(2, '.votesContainer');
     expect(wrapper.find('.rank').first().text()).toEqual('#-');
     expect(wrapper.find('.username').first().text()).toEqual('Loading...');
   });
 
-  it('Should fetch and render delegate names and ranks', () => {
+  it.skip('Should fetch and render delegate names and ranks', () => {
```
isalga, line 39: I see in general lot of skipped unit test. Is it plan in the future to replace or update them?

reyraa: Yes, I had to skip it to limit the number of irrelevant chnages to the title of my PR.
We should add/update unit tests by #3074

src/components/screens/wallet/votes/voteRow.js
reyraa: This doesn't account for when there's no receipt. There were too many of such small issues caused by the new structure of vote tx. I'll fix these issues by #3074.

src/components/screens/wallet/overview/balanceInfo/index.js
```diff
@@ -41,6 +41,18 @@ const BalanceInfo = ({
         </div>
         <SignInTooltipWrapper position="bottom">
           <div className={styles.actionRow}>
+            {
+              activeToken === 'LSK' && (
```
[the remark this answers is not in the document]
reyraa: Done

isalga (approved):

PR #3080, opened 2020-09-25
Fix fee calculation issues - Closes #3051

### LINKED ISSUE #3051: Correct fee caculation
### Description
The current implementation of fee calculation is slightly different than that of Lisk Elements. Therefore some tx get rejected by the API. 
- The calculation of the min fee required is off by a few beddows.
- When changing between different transaction priority options, the calculated fee is inconsistent. 

### Acceptance Criteria
- Refactor the fee calculation algorithm so that it calculates the correct fee.
- The fee calculated for different priority types should be the same for the same transaction. 
- Use the real min fee for the minimum required for the transaction instead of the lowest transaction priority option.

UsamaHameed:
### What was the problem?
This PR resolves #3051

### How was it solved?
- [x] Refactor the fee calculation algorithm so that it calculates the correct fee.
- [x] The fee calculated for different priority types should be the same for the same transaction.
- [x] Use the real min fee for the minimum required for the transaction instead of the lowest transaction priority option.

### How was it tested?
Unit and e2e tests

### DIFF
--- a/src/utils/api/lsk/transactions.js
+++ b/src/utils/api/lsk/transactions.js
@@ -81,8 +81,8 @@ const txTypeClassMap = {
 
 // eslint-disable-next-line max-statements
 export const createTransactionInstance = (rawTx, type) => {
-  const DUMMY_FEE = '18446744073709551615';
-  const DUMMY_SIGNATURE = '204514eb1152355799ece36d17037e5feb4871472c60763bdafe67eb6a38bec632a8e2e62f84a32cf764342a4708a65fbad194e37feec03940f0ff84d3df2a05';
+  const FEE_BYTES_PLACEHOLDER = '18446744073709551615';
+  const SIGNATURE_BYTES_PLACEHOLDER = '204514eb1152355799ece36d17037e5feb4871472c60763bdafe67eb6a38bec632a8e2e62f84a32cf764342a4708a65fbad194e37feec03940f0ff84d3df2a05';
   const asset = {
     data: rawTx.data,
   };
@@ -101,8 +101,8 @@ export const createTransactionInstance = (rawTx, type) => {
     senderPublicKey: rawTx.senderPublicKey,
     nonce: rawTx.nonce,
     asset,
-    fee: DUMMY_FEE,
-    signatures: [DUMMY_SIGNATURE],
+    fee: FEE_BYTES_PLACEHOLDER,
+    signatures: [SIGNATURE_BYTES_PLACEHOLDER],
   });
 
   return tx;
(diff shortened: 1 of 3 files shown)


src/utils/api/lsk/transactions.js
```diff
@@ -74,21 +72,40 @@ export const getSingleTransaction = ({
...
+  transfer: Lisk.transactions.TransferTransaction,
+  registerDelegate: Lisk.transactions.DelegateTransaction,
+  vote: Lisk.transactions.VoteTransaction,
+};
+
+
+// eslint-disable-next-line max-statements
+export const createTransactionInstance = (rawTx, type) => {
+  const DUMMY_FEE = '18446744073709551615';
+  const DUMMY_SIGNATURE = '204514eb1152355799ece36d17037e5feb4871472c60763bdafe67eb6a38bec632a8e2e62f84a32cf764342a4708a65fbad194e37feec03940f0ff84d3df2a05';
```
isalga, line 85: Are these dummy values planned to be removed? and when?

src/utils/api/lsk/transactions.js
```diff
@@ -139,7 +156,7 @@ export const broadcast = (transaction, network) => new Promise(
  */
 export const getTransactionBaseFees = () => (
   new Promise(async (resolve) => {
-    const fee = 1e7;
+    const fee = 1e3;
```
isalga, line 159: Where did you get this value?

UsamaHameed: This is just a realistic mock and we will eventually get this from the service.

src/utils/api/lsk/transactions.js
UsamaHameed: These are necessary to calculate the transaction size. I don't know when we will/can remove them right now.

reyraa: @isalga Their names are somewhat confusing. Those are values of 8 and 64bytes. We need them to. calculate the minimum fee to use as a base, even before the user enters any value in Send form fields.

src/utils/api/lsk/transactions.js
```diff
@@ -74,21 +72,40 @@ export const getSingleTransaction = ({
...
+  const DUMMY_SIGNATURE = '204514eb1152355799ece36d17037e5feb4871472c60763bdafe67eb6a38bec632a8e2e62f84a32cf764342a4708a65fbad194e37feec03940f0ff84d3df2a05';
+  const asset = {
+    data: rawTx.data,
+  };
+
+  if (type === 'transfer') {
+    asset.recipientId = rawTx.recipient;
+    asset.amount = rawTx.amount;
+  } else if (type === 'registerDelegate') {
+    asset.username = rawTx.username || '';
```
reyraa, line 94: `rawTx.username || ''` could have been done in the React component.

src/utils/api/lsk/transactions.js
reyraa: @UsamaHameed  Maybe add a comment that this function will be refactored and refer to the Github ticket number.

reyraa (approved):
Good job @UsamaHameed .

PR #3115, opened 2020-10-16
Add Editor component for the multi signature modal - resolves #3091

### LINKED ISSUE #3091: Enable users to register a new multisignature group
### Description
Users must be able to convert an existing account with sufficient balance to a multisignature group. 

**Preparation**
- You need an account with minimum balance in the amount of transaction fee.

**Initiate the tx**
- Input number of required signatures between 2 and 64
- Add up to 64 members.
- Define mandatory/optional status for each member individually.
- Choose low, medium or high priority. Or Input a custom fee.

**Review and sign**
- Review registration details.
- Sign the transaction.

**Share**
- Download JSON file or copy the transaction to share with members.

### Acceptance Criteria
- Multisignature groups must have a button on the wallet address box to open the group registration modal.
- The modal has a form to input:
   - The number of required signatures between 2 and 64
   - Address, public key, bookmark title or delegate username of each member.
   - Define mandatory/optional status for each member,
- Users must be able to add up to 64 members
- Users must be able to choose tx processing speed and navigate to tx review step
 - After reviewed, they should be able to sign and share the tx with other members.

UsamaHameed:
### What was the problem?
This PR resolves #3091

### How was it solved?
Added relevant UI

### How was it tested?
Added unit test


src/components/screens/multiSignature/editor/editor.js
```diff
@@ -7,16 +7,36 @@ import TransactionPriority from '../../../shared/transactionPriority';
...
 const token = tokenMap.LSK.key;
-const txType = 'multiSignature';
+const txType = 'createMultiSig';
+const MAX_MULTI_SIG_MEMBERS = 64;
 
+const placeholderMember = {
+  identifier: undefined, isMandatory: false,
+};
+
+// eslint-disable-next-line max-statements
```
isalga, line 25: What do you think about breaking down this component like `/screens/lockedBalance/lockedBalance` in 2, `editor` `form`? It is already exceeding max-statements, that would help it to make it more readable.

src/components/screens/multiSignature/result/result.js
```diff
@@ -36,10 +38,11 @@ const Result = ({
   }, [transactionInfo]);
 
   return (
-    <section className="transaction-status">
+    <section className={`${styles.wrapper} transaction-status`}>
       <div className={styles.header}>
         <h1>{t('Register multisignature account')}</h1>
       </div>
+      <ProgressBar current={4} />
```
isalga, line 45: Should not be 3?

isalga (changes requested):

src/components/screens/multiSignature/editor/editor.js
UsamaHameed: It would not make much of a difference because all of the state will still be in the parent and we will only be able to more some jsx to a new component.

src/components/screens/multiSignature/result/result.js
UsamaHameed: Technically yes but in the design, when we are at the 3rd step, we want to style as if it were an older step (like step would be styled when we are at step 2). Thats why, I used 4 here.

isalga (approved):

PR #3130, opened 2020-10-23
Enable users to sign multisignature transactions - closes #3092

isalga:
### What was the problem?
This PR resolves #3092 

### How was it solved?
- Add modal
- Add sidebar menu
- Add UI components
- Validate JSON input
- Sign transaction using lisk-core
- Copy and download JSON transaction

### How was it tested?
- Add unit test

PR #3132, opened 2020-10-28
 Enable users to view multisignature group configuration - Closes #3094

### LINKED ISSUE #3094: Enable users to view multisignature group configuration
### Description
For a given multisignature group, searched or signed in with, users must be able to:
- View account members list.
- View requires signatures.
- Copy members address and/or public key.
Said information must be available through the wallet page.

[UI Design](https://www.figma.com/proto/kNu67yxyOZCJKU21fgFvBl/Multisignature?node-id=228%3A24904&viewport=145%2C516%2C0.05319149047136307&scaling=min-zoom)

### Acceptance Criteria
- Multisignature groups must have a button on the wallet address box to open the group configuration modal.
- The group configuration modal displays:
   - The list of members with avatar, address and potential delegate username or bookmark title
   - Mandatory/optional state of each member
   - Minimum required signatures.

reyraa:
### What was the problem?
This PR resolves #3094

### How was it solved?
- Changed the multisig button to open the account details modal for multisig accounts.
- Created multisig account details modal.

### How was it tested?
Unit tests


src/components/screens/multiSignature/summary/summary.js
```diff
@@ -27,7 +27,7 @@ const Member = ({ member, i, t }) => (
   </div>
 );
 
-const Members = ({ members, t }) => {
+export const Members = ({ members, t }) => {
```
UsamaHameed, line 30: Should we consider moving this to `shared` components? This is already being used at more than 1 places.

src/components/screens/multisigAccountDetails/index.js
```diff
@@ -0,0 +1,69 @@
...
+import styles from './styles.css';
+
+const MultisigAccountDetails = ({ t }) => {
+  const account = useSelector(state => getActiveTokenAccount(state));
+
+  // @todo We shouldn't be needing this if we lazyload the routes.
+  if (!account.keys || account.keys.numberOfSignatures === 0) return null;
+
+  const { numberOfSignatures, optionalKeys, mandatoryKeys } = account.keys;
+  const [members, setMembers] = useState([]);
```
UsamaHameed, line 23: Why do you need to store members in state? Can we not extract them directly from `account.keys...`?

UsamaHameed (changes requested):

src/components/screens/wallet/overview/accountInfo/index.js
```diff
@@ -19,9 +19,92 @@ const BookmarkIcon = ({ bookmark }) => (
...
+    >
+      <p>{isMultisig ? t('View multisignature account details') : t('Register multisignature')}</p>
+    </Tooltip>
+  </div>
+);
+
+const BookmarkButton = ({
+  delegate, address, bookmark, host, t,
+}) => {
+  if (host === address) return null;
```
UsamaHameed, line 42: Should this kind of validation not be done by the parent? If we do it like this, it is hiding some implementation detail that the component is not concerned with and should really exist in the parent. Also, in case when `host === address`, we will still see `BookmarkButton` in the react tree, albeit empty.

src/components/screens/multisigAccountDetails/index.js
```diff
@@ -0,0 +1,69 @@
...
+import BoxContent from '../../toolbox/box/content';
+import BoxInfoText from '../../toolbox/box/infoText';
+import Dialog from '../../toolbox/dialog/dialog';
+import Tooltip from '../../toolbox/tooltip/tooltip';
+import { Members } from '../multiSignature/summary/summary';
+import { getActiveTokenAccount, extractAddress } from '../../../utils/account';
+import styles from './styles.css';
+
+const MultisigAccountDetails = ({ t }) => {
+  const account = useSelector(state => getActiveTokenAccount(state));
```
UsamaHameed, line 17: I have started to follow the general structure of having an index file for every exported component and then using higher order components or hooks that I don't wanna be concerned with called in that index file so that the unit tests for that particular component don't have any details of how the hook/hocs are implemented. This makes the unit tests clearer and easier to write/debug. 

In general, we should start using the ES module system more i.e. within folders, have an index file that exposes modules out of the that folder and code inside that folder should only be accessed through that index file. This also helps with tree shaking and lazy loading.

src/components/screens/multiSignature/summary/summary.js
reyraa: I just didn't know if I'll cause a conflict with your other PRs.

src/components/screens/multisigAccountDetails/index.js
reyraa: The cyprographical calculation of address from publicKey is a heavy duty. If we perform this on up to 64 member, it'll slow down the re-render significantly.

src/components/screens/wallet/overview/accountInfo/index.js
reyraa: It was to reduce the number of statements in the parent so we don't have to disable the eslint rule. but you're right in general.

src/components/screens/multisigAccountDetails/index.js
UsamaHameed: In that case, you probably wanna use useMemo and calculate the members only when account.keys change.

src/components/screens/wallet/overview/accountInfo/index.js
reyraa: Done

src/components/screens/multisigAccountDetails/index.js
reyraa: Thought it's a good point for making the tests easier, but it also scatters the code in more files. I wasn't aware of this decision and I requested contradicting changes on #3130 from @isalga .

src/components/screens/multiSignature/summary/summary.js
reyraa: Done

src/components/screens/multisigAccountDetails/index.js
reyraa: Done

UsamaHameed (approved):

PR #3179, opened 2020-11-18
Improve voting search params checks - Closes #3160

### LINKED ISSUE #3160: Voting state is not initialised correctly through search params
### Expected behavior
The process of adding votes through the search params in the url does not work, consequently the launch protocol will not work either. We also need to modify the validations on the search params to be able to accept a single vote/unvote rather than an array.

### Actual behavior
The votes in the search params are not used to initialise the local voting state. 

### Steps to reproduce
Navigate to `?votes=delegate1` or `?votes=delegate1,delegate2`.

UsamaHameed:
### What was the problem?
This PR resolves #3160

### How was it solved?
Improve the normalisation function

### How was it tested?
Manually

### DIFF
--- a/src/actions/urlProcessor.js
+++ b/src/actions/urlProcessor.js
@@ -21,10 +21,20 @@ const getAccounts = async (usernames, network) =>
  * @returns {[String]} Array of strings or an empty array
  */
 const normalizeUsernames = (usernames) => {
-  if (!Array.isArray(usernames)) return [];
+  if (usernames === undefined) {
+    return [];
+  }
+
+  if (!Array.isArray(usernames)) {
+    return [usernames];
+  }
+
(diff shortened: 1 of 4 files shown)


src/actions/urlProcessor.js
```diff
@@ -13,6 +13,8 @@ import regex from '../utils/regex';
 const getAccounts = async (usernames, network) =>
   Promise.all(usernames.map(username => getAccount({ username, network })));
 
+const isUsernameValid = username => typeof username === 'string' && username.length > 3;
```
reyraa, line 16: There's a regex to check the username validity instead of `username.length > 3`

isalga (approved):

reyraa (approved):

PR #3208, opened 2020-11-30
Fix visual issues - Closes #3206

### LINKED ISSUE #3184: Reinstate date values, sorting and filtering options
### Description
Changed made by #3182 were temporary and we should revert them after changes defined by #2890 are implemented.
Refer to #3185 for details of the changes.

### Acceptance Criteria
- Display date from tx list
- Display date from tx details
- Display sorting by date
- Display filtering by date
- Reinstate related e2e tests

### LINKED ISSUE #3206: Visual issues discovered testing v2.0.0-alpha.0
### Description
The following issues are discovered testing v2.0.0-alpha.0

Delegates profile page has alignment issues
image

In register delegate page, nick name is not accepting uppercase letters
image

In register delegate page, tooltip is misaligned
image

In Transaction report page, no transaction information is available and also the graph 'unlock LSK' is striked off
image

Custom transaction fee in Send LSK is confusing. Values are different on focus and on focus out
image

Can we restrict the text as per the limit ? Right now 'Go to confirmation' button is not active which is good , but it will be nice if we restrict it
image
Can we have the transaction fee text in single line ?
image
Screenshot 2020-11-30 at 11 30 15
Screenshot 2020-11-30 at 11 30 11
Screenshot 2020-11-30 at 11 30 06
Screenshot 2020-11-30 at 11 30 02
Screenshot 2020-11-30 at 11 29 56

UsamaHameed:
### What was the problem?
This PR resolves #3206

### How was it solved?

TODO:
- [x] Empty voting state
- [x] Delegates profile page has alignment issues
- [x] Update service avatars
- [x] Custom transaction fee in Send LSK is confusing. Values are different on focus and on focus out
- [x] Incorrect recipient avatars
- [x] Copy icon
- [x] Icon colour
- [x] In register delegate page, tooltip is misaligned

Won't fix:

- ~~Can we restrict the text as per the limit ? Right now 'Go to confirmation' button is not active which is good , but it will be nice if we restrict it~~
As already mentioned, we have two feedbacks to show the user that the message length is more than allowed. It would be hard to implement the check that has been asked for and would not bring much value. 

- ~~Can we have the transaction fee text in single line ?~~
We will get the table column widths back to normal once we resolve #3184.

- ~~In Transaction report page, no transaction information is available and also the graph 'unlock LSK' is striked off~~
We need to connect to an appropriate lisk service node for the transactions to be visible. 

- ~~In register delegate page, nick name is not accepting uppercase letters~~

### How was it tested?
Manually


reyraa:
So far so good. 😊

src/components/screens/votingQueue/editor/editor.js
```diff
@@ -158,60 +138,75 @@ const Editor = ({
...
-              header={header(t)}
-              row={VoteRow}
-              iterationKey="address"
-              canLoadMore={false}
-              emptyState={{ message: t('No votes in queue.') }}
+        <div className={styles.headerContainer}>
+          <header>
+            {t('Voting Queue')}
+          </header>
+          {!showEmptyState ? (
```
isalga, line 151: I prefer ```!showEmptyState &&``` . Then it won't need ``` : null}``` which will make the code cleaner

src/components/screens/votingQueue/styles.css
```diff
@@ -29,21 +29,11 @@
...
-    @mixin headingNormal;
-
-    font-size: var(--font-size-h5);
-    width: 45%;
-    color: var(--color-maastricht-blue);
-  }
 
   & .voteStats {
-    width: 18%;
+    padding: 0 16px;
```
isalga, line 36: Suggested change: ```padding: 0 var(--horizontal-padding-m);```

src/components/screens/votingQueue/editor/editor.js
UsamaHameed: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

reyraa: I knew about this problem but showEmptyState is a boolean and should be fine. I applied the requested change, but if @UsamaHameed disagrees I can revert it

reyraa (approved):

PR #3220, opened 2020-12-10
Refactor delegate api module

isalga:
### What was the problem?
Delegate module was not following the same code style as the other modules and was duplicating API validation.

### How was it solved?
Refactor code and remove unncessary validation

### How was it tested?
Unit test


reyraa (approved):
Looks good. Thank you

Issue #3229, opened 2020-12-16
Dashboard: Lisk balance displays bitcoin icon


isalga:
### Expected behavior
Should display lisk icon

### Actual behavior
Displays bitcon icon

### Steps to reproduce
- Log in
- Go to dashboard

### Which version(s) does this affect? (Environment, OS, etc...)
`2890-support-sdk-4`

PR #3244, opened 2020-12-16
Fix dashboard after api module refactor - Closes #3228 & #3229

### LINKED ISSUE #3228: Dashboard: news feeds is not displayed
### Expected behavior
Should display news

### Actual behavior
News feeds appears empty

### Steps to reproduce
- Log in
- Go to dashboard

`2890-support-sdk-4`

isalga:
### What was the problem?
This PR resolves #3228 and #3229 

### How was it solved?
- Implement getNews method
- Fix icon condition

### How was it tested?
Add unit test

### DIFF
--- a/src/components/screens/dashboard/newsFeed/newsFeed.js
+++ b/src/components/screens/dashboard/newsFeed/newsFeed.js
@@ -15,18 +15,15 @@ const NewsFeed = (props) => {
     newsFeed,
     t,
   } = props;
-  const filteredNewsFeed = newsFeed.data || [];
+  const filteredNewsFeed = newsFeed.data.data || [];
   const serviceUrl = useSelector(state => state.network.serviceUrl);
   const sources = ['twitter_lisk', 'drupal_lisk_general'];
   const [source, setSource] = useState(sources.join(','));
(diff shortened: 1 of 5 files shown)


src/components/screens/dashboard/newsFeed/newsFeed.js
```diff
@@ -15,18 +15,15 @@ const NewsFeed = (props) => {
     newsFeed,
     t,
   } = props;
-  const filteredNewsFeed = newsFeed.data || [];
+  const filteredNewsFeed = newsFeed.data.data || [];
```
UsamaHameed, line 18: Can we use the `withData` `transformResponse` function here to only get data instead of `data.data`?

UsamaHameed (changes requested):

src/components/screens/dashboard/newsFeed/newsFeed.js
isalga: Some lines below it needs error `newsFeed.error` thus transformResponse would get rid of it.

UsamaHameed (approved):

PR #3271, opened 2021-01-07
Fix delegate monitor tables - Closes #3262

### LINKED ISSUE #3262: Fix breaking issues of delegates monitor caused by the API refactoring
### Expected behavior
- The standby delegates list should show all existing delegates with rank over 101.
- The latest votes tab should populate all designated data rows.

### Actual behavior
- The standby delegates list is empty.
- The latest votes tab:
    - Drop current balance
    - Display the list of vote username and amounts for each row.

### Steps to reproduce
- Connect to betanet
- Open delegate monitor page

2.0.0

isalga:
### What was the problem?
This PR resolves #3262 

### How was it solved?
- Implement pagination for standby delegates and latest votes
- Use offset to retrieve standby delegates
- Use correct vote type code
- Retrieve voted delegate names
- Update vote row from latest votes table

### How was it tested?
- Update unit test


reyraa:
The standby delegates rank is wrong. it shouldn't start from 1, but 102.

reyraa (approved):

PR #3268, opened 2021-01-06
Refactor network handling - Closes #3267

UsamaHameed:
### What was the problem?
This PR resolves #3267

### How was it solved?
Refactored the code so that it is much cleaner and follows our coding standards

### How was it tested?
Existing unit and e2e should be enough

PR #3274, opened 2021-01-07
Fix errors thrown because of missing server URL - Closes #3272

### LINKED ISSUE #3272: Error thrown because of missing server URL
### Expected behavior
The subscription to listen new blocks should handle correctly.

### Actual behavior
The following errors are thrown:
Screenshot 2021-01-07 at 12 45 05

reyraa:
### What was the problem?
This PR resolves #3272

### How was it solved?
Node url was `undefined` because we dispatch `networkSet` once for each token, and `BTC` resolves faster, making `node` address `undefined` at the time of subscription. I moved the subscription into a `LSK` check condition.

### DIFF
--- a/src/store/middlewares/block.js
+++ b/src/store/middlewares/block.js
@@ -69,8 +69,10 @@ const blockMiddleware = store => (
     next(action);
     switch (action.type) {
       case actionTypes.networkSet:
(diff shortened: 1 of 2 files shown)


isalga (approved):

reyraa:
#3268 resolves this issues.  I closed this.

PR #3275, opened 2021-01-07
Update unit tests after updating to SDK v4 -Closes #3247

### LINKED ISSUE #3247: Update unit tests after updating to SDK v4
### Expected behavior
The unit tests should pass.

### Actual behavior
Some unit tests fail because of the changes introduced as a result of the update to SDK v4.

### Steps to reproduce
Run the unit tests. 

SDK v4, lisk desktop v2.0.0-beta.0

reyraa:
### What was the problem?
This PR resolves #3247

### How was it solved?
- Updated broken unit tests
- Improved test coverage where possible


src/components/screens/login/networkSelector/networkSelector.js
```diff
@@ -76,7 +76,7 @@ const NetworkSelector = ({
 
   // eslint-disable-next-line max-statements
   const validateCorrectNode = async (networkName) => {
-    const networkToSet = getNetwork(networkName);
+    const networkToSet = networks[networkName];
```
UsamaHameed, line 79: This was using the getNetwork function from the component itself, not the one you removed.

reyraa: oh, Thanks.

src/utils/searchParams.js
```diff
@@ -30,7 +30,7 @@ export const selectSearchParamValue = (search, paramToSelect) =>
  * @param {object} params the parsed searchParams object
  */
 // eslint-disable-next-line import/prefer-default-export
-export const strigifySearchParams = (params) => {
+export const stringifySearchParams = (params) => {
```
UsamaHameed, line 33: This is a named export. You probably would want to change the names at all the imports as well.

src/actions/account.js
```diff
@@ -156,9 +156,12 @@ export const login = ({ passphrase, publicKey, hwInfo }) => async (dispatch, get
     toast.error(getConnectionErrorMessage(error));
     dispatch(accountLoggedOut());
   } else {
+    const loginType = hwInfo
+      ? ['trezor', 'ledger'].find(item => hwInfo.deviceModel.toLowerCase().indexOf(item) > -1)
```
UsamaHameed, line 160: `indexOf()` would only match the first index in both the strings so in case we have more complicated strings, this would break but for now it should be fine.

UsamaHameed (changes requested):

src/utils/searchParams.js
reyraa: It is updated.

src/actions/account.js
reyraa: All devices that we support have the producer name in their `deviceModel`s, I'm just checking the producer.

UsamaHameed (approved):

PR #3300, opened 2021-01-20
Fix unvote modal display and transaction broadcast - Closes #3282

### LINKED ISSUE #3282: Unlock token screen display wrong times
### Expected behavior
The unlock times must be displayed correctly and the screen should enable the confirm button once there are tokens available to unlock.

### Actual behavior
Due to changes to the account API endpoint response, this screen is broken. It always displays "will be available to unlock in a few seconds".

UsamaHameed:
### What was the problem?
This PR resolves #3282

### How was it solved?
Fix the bugs related to the display and broadcast

### How was it tested?
Manually


src/utils/account.js
```diff
@@ -54,32 +54,56 @@ export const getActiveTokenAccount = state => ({
...
+*/
+export const calculateBalanceLockedInUnvotes = (unlocking = []) =>
+  unlocking.reduce((acc, vote) => acc + parseInt(vote.amount, 10), 0);
 
 // TODO handle delegate punishment when Lisk Service is ready
 export const getDelayedAvailability = isSelfVote => (isSelfVote
   ? unlockTxDelayAvailability.selfUnvote : unlockTxDelayAvailability.unvote);
 
-export const isBlockHeightReached = ({ unvoteHeight, delegateAddress }, currentBlock, address) => {
+export const isBlockHeightReached = ({ height, delegateAddress }, currentBlock, address) => {
```
isalga, line 73: If the `vote.height.end` is the block when the locked balance is available to be unlocked then we should just compare ```currentBlock.height <= vote.height.end```

src/components/screens/lockedBalance/lockedBalance/form.js
```diff
@@ -53,12 +53,16 @@ const Form = ({
       nonce: account.nonce,
       fee: `${toRawLsk(parseFloat(selectedFee))}`,
       passphrase: account.passphrase,
-      unlockingObjects: getAvailableUnlockingTransactions(account, currentBlock),
+      unlockingObjects: getUnlockableUnlockingObjects(account, currentBlock),
```
isalga, line 56: I think `getUnlockableUnlockingObjects` is a bit confusing name. In this context `getUnlockingObjects` explains itself better

src/components/screens/lockedBalance/lockedBalance/unlockingList.js
```diff
@@ -5,10 +5,10 @@ import { getDelayedAvailability, isBlockHeightReached } from '../../../../utils/
 import LiskAmount from '../../../shared/liskAmount';
 import { tokenMap } from '../../../../constants/tokens';
 
-const getPendingTime = ({ unvoteHeight, delegateAddress }, currentBlockHeight, { address }) => {
+const getPendingTime = ({ height, delegateAddress }, currentBlockHeight, { address }) => {
   const isSelfVote = address === delegateAddress;
   const delayedAvailability = getDelayedAvailability(isSelfVote);
-  const awaitingBlocks = delayedAvailability - (currentBlockHeight - unvoteHeight);
+  const awaitingBlocks = delayedAvailability - (currentBlockHeight - height.start);
```
isalga, line 11: If `height.end` Is when the balance is able to be unlocked then we don't need `delayed availability`. We should just substract `currentBlockHeight` to `height.end` : ```currentBlockHeight - height.end```

isalga: Thus we won't need `getDelayedAvailability` at all

reyraa: Good point. that part is already calculated by Lisk Service.

src/components/screens/lockedBalance/lockedBalance/form.js
UsamaHameed: the function filters the unlocking objects and only returns the `unlockingObjects` that are `unlockable` so `getUnlockable-UnlockingObjects` sounds like a very fitting name to me 😃

reyraa (dismissed):
Looks good.

src/components/screens/lockedBalance/lockedBalance/unlockingList.js
```diff
@@ -1,47 +1,43 @@
...
+import { isBlockHeightReached } from '../../../../utils/account';
 import LiskAmount from '../../../shared/liskAmount';
 import { tokenMap } from '../../../../constants/tokens';
 
-const getPendingTime = ({ unvoteHeight, delegateAddress }, currentBlockHeight, { address }) => {
-  const isSelfVote = address === delegateAddress;
-  const delayedAvailability = getDelayedAvailability(isSelfVote);
-  const awaitingBlocks = delayedAvailability - (currentBlockHeight - unvoteHeight);
+const getPendingTime = (unvoteHeight, unlockHeight) => {
+  const awaitingBlocks = unvoteHeight - unlockHeight;
```
isalga, line 9: Should not be ```unlockHeight - unvoteHeight``` ?

isalga (changes requested):

isalga (approved):

PR #3337, opened 2021-02-03
Update account info after sending tokens - Closes #3322

### LINKED ISSUE #3322: Wallet balance does not update its value after sending tokens
### Expected behavior
If an account has X LSK and sends some tokens, those amount of tokens should be subtracted after the transaction is confirmed

### Actual behavior
After transaction is confirmed account balance remains the same as previous to the transaction

### Steps to reproduce
- Log in with one account
- Send LSK
- Observe that account balance remains the same

macOs, betanet

reyraa:
### What was the problem?
This PR resolves #3322

### How was it solved?
The API call to retrieve the account and transactions were called/fired too soon. It takes a few ms for Lisk Core to update the tx indexes and I added a delay of 100ms before fetching the information (after receiving new block info).

I had to refactor a few actions to make it work, therefore I had to adapt the account middleware and reducer. 

I also disabled the account initialization banner, since the new address system doesn't have this issue.

### How was it tested?
Unit tests

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -5,30 +5,6 @@ import { getConnectionErrorMessage } from '../utils/getNetwork';
 import loginTypes from '../constants/loginTypes';
 import { networkStatusUpdated } from './network';
 import actionTypes from '../constants/actions';
-import { tokenMap } from '../constants/tokens';
-
-/**
- * Trigger this action to remove passphrase from account object
- *
- * @param {Object} data - account data
- * @returns {Object} - Action object
- */
-export const removePassphrase = data => ({
-  data,
-  type: actionTypes.removePassphrase,
-});
-
-/**
- * Trigger this action to update the account object
- * while already logged in
- *
- * @param {Object} data - account data
- * @returns {Object} - Action object
- */
-export const accountUpdated = data => ({
-  data,
-  type: actionTypes.accountUpdated,
-});
 
 /**
  * Trigger this action to log out of the account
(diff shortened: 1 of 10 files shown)


src/actions/account.js
```diff
@@ -145,11 +99,13 @@ export const login = ({ passphrase, publicKey, hwInfo }) => async (dispatch, get
   const { network, settings } = getState();
   dispatch(accountLoading());
 
-  const activeTokens = Object.keys(settings.token.list)
-    .filter(key => settings.token.list[key]);
-  const [error, info] = await to(getAccounts(activeTokens, {
-    network, publicKey, passphrase,
-  }));
+  const params = Object.keys(settings.token.list)
+    .filter(key => settings.token.list[key])
```
UsamaHameed, line 103: Are you filtering something here? Should this be a `map` may be?

src/store/middlewares/account.js
```diff
@@ -21,11 +20,19 @@ import transactionTypes from '../../constants/transactionTypes';
...
+ * After a new block is created and broadcasted
+ * it takes a few ms for Lisk Service
+ * to update transactions index, so we need to wait
+ * before retrieving the the transaction by blockId
+ *
+ * @returns {Promise} resolves with True after 100ms
+ */
+const delay = () => new Promise((resolve) => {
+  setTimeout(() => {
+    resolve(true);
```
UsamaHameed, line 33: if not needed, please remove `true`

UsamaHameed (changes requested):

src/actions/account.js
reyraa: yes, the value of `settings.token.list[key]` is a boolean.

isalga (approved):

UsamaHameed (approved):

PR #3375, opened 2021-02-11
Fix delegate status column - Closes #3356

### LINKED ISSUE #3356: The delegate status tooltip displays raw numbers
### Expected behavior
- Forging and not forging delegates must have different icons.
- The number inside the tooltip must display forging time.
- icons in the design seem smaller.
-  Forging times are updated after new block is created.

### Actual behavior
Screenshot 2021-02-09 at 09 43 04
Screenshot 2021-02-09 at 09 42 55

### Steps to reproduce
Connect to  betanet and navigate to delegates monitor.

2.0.0-beta.2

isalga:
### What was the problem?
This PR resolves #3356 

### How was it solved?

- [x] Forging and not forging delegates must have different icons.
- [x] The number inside the tooltip must display forging time.
- [x] icons in the design seem smaller.
- [x] forgingTimes are updated after new block is created.

### How was it tested?
Manually


src/actions/blocks.js
```diff
@@ -84,7 +84,7 @@ export const forgingTimesRetrieved = nextForgers => async (dispatch, getState) =
   latestBlocks.forEach((item, index) => {
     if (index > forgedInRoundNum && !forgingTimes[item.generatorPublicKey]) {
       // if I can't find it in the previous round
-      if (index >= forgedInRoundNum + 101 && latestBlocks.indexOf(item) === index) {
+      if (index >= forgedInRoundNum + 103 && latestBlocks.indexOf(item) === index) {
```
reyraa, line 87: Please read 103 from constants.

src/components/screens/monitor/delegates/delegatesTable/delegateRow.js
```diff
@@ -17,14 +17,14 @@ import { addedToWatchList, removedFromWatchList } from '../../../../../actions/w
 const roundStatus = {
   forging: 'Forging',
   awaitingSlot: 'Awaiting slot',
-  notForging: 'Not forging',
+  notForging: 'Missed block',
```
reyraa, line 20: Please remove `notForging` entirely.

reyraa (changes requested):

src/actions/blocks.js
UsamaHameed, line 87: I am looking at this statement and I don't understand why is it there. `latestBlocks.indexOf(item) === index` This should always be true or am I missing something?

UsamaHameed, line 87: I am looking at `latestBlocks.indexOf(item) === index` and I don't understand why is it there. This should always be true or am I missing something?

UsamaHameed (approved):

src/actions/blocks.js
reyraa: Latest blocks are 206, 2 rounds. this checks if the item exists in the lest round or the one before. We can improve though.

UsamaHameed: Okay but the `index` is already the _index_ at which the `item` is situated at in the `latestBlocks` array. i.e. `latestBlocks[index]` returns `item`. So, `latestBlocks.indexOf(item) === index` would always be true.

reyraa: Assume we have `arr = ['a', 'b', 'c, 'a']`, then `arr.forEach(item => arr.indexOf(item))` will return `[0, 1, 2, 0]`.

UsamaHameed (dismissed):

UsamaHameed (dismissed):

reyraa (approved):
Looks great. Thank you for applying the requests.

PR #3394, opened 2021-02-19
Fix race condition between account updated action and initialization transaction

### LINKED ISSUE #3393: Fix race condition between account initialization transaction and account updated api call
### Expected behavior
The user should not see the intialization transaction screen after having submitted the initialization transaction. 

### Actual behavior
The user sees the initialization transaction after having broadcasted the transaction.

UsamaHameed:
### What was the problem?
This PR resolves #3393

### How was it solved?
By executing the account init logic after the account updated action

### How was it tested?
Manually and unit tests


src/store/middlewares/account.js
```diff
@@ -239,6 +247,11 @@ const accountMiddleware = store => next => (action) => {
     case actionTypes.updateTransactions:
       votePlaced(store, action);
       break;
+    case actionTypes.accountUpdated:
```
isalga, line 250: If this case is empty then it should be removed to it goes to `default: break;`

isalga (changes requested):

src/store/middlewares/account.js
UsamaHameed: This case is not empty. I have combined two cases together because the function to use in both the cases is the same.

isalga (approved):

PR #3408, opened 2021-02-26
Remove CHF from settings - Closes #3407

### LINKED ISSUE #3407: Remove CHF conversion
### Expected behavior

The user should be able to see the their balance in CHF if they choose to do so.

### Actual behavior

The CHF currency conversion is broken in lisk-cloud and therefore the conversion is broken in lisk-desktop as well.

### Steps to reproduce
Choose CHF as the currency in settings.

UsamaHameed:
### What was the problem?
This PR resolves #3407

### How was it solved?
Removed CHF from constants

### How was it tested?
Manually

### DIFF
--- a/src/constants/settings.js
+++ b/src/constants/settings.js
@@ -1,5 +1,5 @@
 const settings = {
-  currencies: ['USD', 'EUR', 'CHF'],
+  currencies: ['USD', 'EUR'],
   keys: {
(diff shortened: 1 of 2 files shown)


reyraa (approved):
Too imperative for my taste. But since it's temporary, I guess we can live with that.

PR #3433, opened 2021-03-17
Support new fee api from lisk service - Closes #3296

### LINKED ISSUE #3296: Adapt to the new fees API implementation
### Description
The new version of Lisk Service renames `fee_estimates` endpoint to `fees`. This endpoint doesn't accept any parameters and the response includes:
- feeEstimatePerByte
   - low
   - medium
   - high
- baseFeeById
  - includes name fees for each moduleAsset id
-  baseFeeByName
   - includes name fees for each moduleAsset name
- minFeePerByte

We should retrieve, and use this information for name fees too. This replaces name fees hard coded in `transactionTypes.js` file. We won't store this information on the Redux store. Since these values are prone to change, we should retrieve them at the time of rendering the corresponding component.

### Action points
- Update API method
- Create the corresponding action
- Update transaction reducer to store fees

### Motivation
Support Lisk SDK v5.

### Acceptance Criteria
 - Fee values must be retrieved from Lisk Service `fees` endpoint, and used to make new transactions.
 - Adequate unit test coverage if required.

UsamaHameed:
### What was the problem?
This PR resolves #3296

### How was it solved?
Adapt the fee api

### How was it tested?
Unit and e2e should work when the epic is complete


src/components/shared/transactionPriority/reducer.js
```diff
@@ -0,0 +1,66 @@
+/* eslint-disable import/prefer-default-export */
+import { tokenMap, minAccountBalance } from '@constants';
```
isalga, line 2: Can you add unit test for this file?

isalga (changes requested):

src/utils/regex.js
```diff
@@ -1,5 +1,5 @@
 export default {
-  address: /^[1-9]\d{0,19}L$/,
+  address: /^lsk[a-z0-9]{0,38}$/,
```
reyraa, line 2: I assume this is a temporary change and will be replaced by Lisk Elements.

src/components/screens/votingQueue/editor/editor.js
```diff
@@ -2,13 +2,12 @@ import React, { useMemo, useState } from 'react';
 
 import { tokenMap } from '@constants';
```
reyraa, line 3: I like these aliases.

src/constants/transactions.js
```diff
@@ -2,8 +2,15 @@ const maxMessageLength = 64;
 const minAccountBalance = 5000000;
 const minFeePerByte = 1000;
 
+const DUMMY_RECIPIENT_ADDRESS = 'lskdxc4ta5j43jp9ro3f8zqbxta9fn6jwzjucw7yt';
+const DEFAULT_NUMBER_OF_SIGNATURES = 1;
+const DEFAULT_SIGNATURE_BYTE_SIZE = 64;
+
 export {
   maxMessageLength,
```
reyraa, line 10: This a mix of camelCase and UPPERCASE constants. We didn't have uppercase constants in the past, they spread gradually. Given you prefer them, would you be so kind and change the rest too. If not in this PR, then please just create a ticket so we do it later.

src/components/screens/send/form/formLsk.js
```diff
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { maxMessageLength, MODULE_ASSETS } from '@constants';
 import { toRawLsk } from '@utils/lsk';
+import TransactionPriority, { useTransactionPriority, useTransactionFeeCalculation } from '@shared/transactionPriority';
```
reyraa, line 4: Good choice moving this to shared.

src/constants/transactions.js
UsamaHameed: done: https://github.com/LiskHQ/lisk-desktop/issues/3437

src/utils/regex.js
UsamaHameed: Yes and we need to update it all across the repo. This is only temporary.

isalga (approved):

reyraa (approved):

PR #3434, opened 2021-03-17
Adapt accounts api - Closes 3292

### LINKED ISSUE #3292: Adapt to the new accounts API implementation
### Description
The new version of Lisk Service API changes the accounts API parameters and response, affecting Lisk Desktop as consumer. The changes affect at least the following components:
- `wallet/AccountInfo`
- `wallet/BalanceInfo`
- `send/Form`
- `RegisterDelegate`
- `dashboard/WalletDetails`

and functions used in transaction creation process. The said functions access account values such as address, balance and public key and we need to make sure they'll continue working as expected.

We should store the information in the Redux store with the new data structure, hence we should make sure all checks with the account address, public key, balance, etc work as expected.

### Motivation
Support Lisk SDK v5

### Acceptance Criteria
- Lisk Desktop should display account information correctly.
- Lisk Desktop should be able to create all types of transactions.
- Unit tests must be updated.
    - Account visual components.
    - Transaction creation visual components.
    - Transaction creation functions.

isalga:
### What was the problem?
This PR resolves #3292 

### How was it solved?
- Update data structure

### How was it tested?
Updated unit tests


src/utils/regex.js
```diff
@@ -1,5 +1,5 @@
 export default {
-  address: /^[1-9]\d{0,19}L$/,
+  address: /^lsk[a-z0-9]{0,38}$/,
```
reyraa, line 2: Is this temporary? Cause we should test the address using Lisk Elements. A simple regex check won't suffice.

reyraa: ok then we'll have a dedicated PR for it.

reyraa (approved):

reyraa (approved):

UsamaHameed (approved):

PR #3439, opened 2021-03-22
Use consistent selectors across the project - Closes #3438

### LINKED ISSUE #3438: Use consistent selectors across the project
### Expected behavior
The code to select some data from the redux store is the same all across the repo but is being repeated making it harder to maintain and refactor. We should make use of selectors to access data from the redux store.

UsamaHameed:
### What was the problem?
This PR resolves #3438

### How was it solved?
Create selectors and custom hooks for frequently referenced information on the Redux store.

### How was it tested?
NA


src/components/screens/wallet/overview/index.js
```diff
@@ -3,32 +3,23 @@ import { useSelector } from 'react-redux';
...
-};
-
 const Overview = ({
   t, activeToken, transactions, hwInfo,
   discreetMode, isWalletRoute, account,
 }) => {
-  const address = getProp(account, 'address', '');
-  const publicKey = getProp(account, 'publicKey', '');
-  const balance = getProp(account, 'balance', 0);
+  const { address, publicKey, balance } = account.info.LSK.summary;
```
reyraa, line 17: Do we need a null check here?

src/components/shared/transactionPriority/reducer.js
```diff
@@ -0,0 +1,66 @@
...
+import { tokenMap, minAccountBalance } from '@constants';
+import { toRawLsk } from '@utils/lsk';
+
+const calculateAvailableBalance = (balance, token) => {
+  if (token !== tokenMap.LSK.key) {
+    return balance;
+  }
+  if (balance <= minAccountBalance) {
+    return balance;
+  }
```
reyraa, line 11: Can a user transfer this? It just seems somewhat inconstant with the next line.

reyraa: `Math.max(balance - minAccountBalance, 0)`

reyraa (approved):

src/components/screens/wallet/overview/index.js
isalga: Yes, and it should be `account.summary?`

isalga (changes requested):

isalga (approved):

Issue #3467, opened 2021-04-06
Convert bigint type to string before passing around the code


UsamaHameed:
### Expected behavior
We should be able to pass around bigint values normally between actions, middlewares, props etc.

### Actual behavior
Parsing bigints in JSON is not natively supported.
`Uncaught (in promise) TypeError: Do not know how to serialize a BigInt`

PR #3468, opened 2021-04-07
Fix transaction broadcast - Closes #3459, #3467

### LINKED ISSUE #3459: Adapt transaction broadcast
### Expected behavior
The client should be able to broadcast transactions to lisk service.

### Actual behavior
The client broadcasts the transactions to lisk core.

UsamaHameed:
### What was the problem?
This PR resolves #3459, #3467

### How was it solved?
- Fixes issues with transaction broadcast
- Added transformer function to convert the broadcasted transaction back to lisk-service format to ensure compat with UI
- Polyfilled native bigint with `toJSON` function for `JSON.stringify` to be able to serialize it.

Additionally: 
- Fixed a few issues where wallet/votes and wallet/delegateProfile views were crashing
- Fixed an issue with the action `votesRetrieved`

### DIFF
--- a/src/utils/account.js
+++ b/src/utils/account.js
@@ -16,6 +16,23 @@ export const extractPublicKey = (passphrase) => {
   return undefined;
 };
 
+export const extractAddressFromPublicKey = (data) => {
+  if (regex.publicKey.test(data)) {
+    return cryptography.getBase32AddressFromPublicKey(data).toString('hex');
+  }
+  if (Buffer.isBuffer(data)) {
+    return cryptography.getBase32AddressFromPublicKey(data);
+  }
+  return undefined;
+};
+
+export const extractAddressFromPassphrase = (data) => {
+  if (LiskPassphrase.Mnemonic.validateMnemonic(data)) {
+    return cryptography.getBase32AddressFromPassphrase(data).toString('hex');
+  }
+  return undefined;
+};
+
 /**
  * Extracts Lisk address from given passphrase or publicKey
  *
@@ -23,14 +40,11 @@ export const extractPublicKey = (passphrase) => {
  * @returns {String?} - Extracted address for a given valid passphrase or publicKey
  */
 export const extractAddress = (data) => {
-  if (LiskPassphrase.Mnemonic.validateMnemonic(data)) {
-    return cryptography.getBase32AddressFromPassphrase(data).toString('hex');
-  }
-  if (regex.publicKey.test(data)) {
-    return cryptography.getBase32AddressFromPublicKey(data).toString('hex');
-  }
   if (regex.address.test(data)) {
     return cryptography.getAddressFromBase32Address(data);
+  }
+  if (Buffer.isBuffer(data)) {
+    return cryptography.getBase32AddressFromAddress(data);
   }
   return undefined;
 };
--- a/src/utils/account.js
+++ b/src/utils/account.js
@@ -40,7 +40,7 @@ export const extractAddressFromPassphrase = (data) => {
  * @returns {String?} - Extracted address for a given valid passphrase or publicKey
  */
 export const extractAddress = (data) => {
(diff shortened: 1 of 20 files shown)


src/utils/account.js
```diff
@@ -16,22 +16,36 @@ export const extractPublicKey = (passphrase) => {
   return undefined;
```
reyraa, line 16: These retuened `undefined` are essentially silent failures. I think we should add a logger here so we don't overlook the broken features.

reyraa: Same goes for the other two convertors.

src/utils/api/account/lsk.js
```diff
@@ -69,6 +72,7 @@ export const getAccount = async ({
 }) => {
   const normParams = getAccountParams(params);
 
+  console.log({ params, normParams });
```
reyraa, line 75: Remove?

src/utils/api/transaction/lsk.js
```diff
@@ -17,6 +17,7 @@ import ws from '../ws';
 import { getDelegates } from '../delegate';
 import regex from '../../regex';
 import { validateAddress } from '../../validators';
+import { extractAddressFromPublicKey } from '../../account';
```
reyraa, line 20: Should we use aliases here?

src/utils/api/transaction/lsk.js
```diff
@@ -219,51 +220,106 @@ const splitModuleAndAssetIds = (moduleAssetId) => {
...
-    senderPublicKey: Buffer.from(senderPublicKey, 'hex'),
-    nonce: BigInt(nonce),
-    fee: BigInt(fee),
-    signatures: [],
+export const transformTransaction = (transaction) => {
+  const moduleAssetId = [transaction.moduleID, transaction.assetID].join(':');
+  const senderAddress = extractAddressFromPublicKey(transaction.senderPublicKey);
+  const senderPublicKey = transaction.senderPublicKey.toString('hex');
+
+  console.log(transaction);
```
reyraa, line 228: Don't forget to remove this.

src/utils/api/transaction/lsk.js
```diff
@@ -219,51 +220,106 @@ const splitModuleAndAssetIds = (moduleAssetId) => {
...
+  const transformedTransaction = {
+    id: transaction.id.toString('hex'),
+    moduleAssetId,
+    fee: String(transaction.fee),
+    nonce: String(transaction.nonce),
+    sender: { publicKey: senderPublicKey, address: senderAddress },
+    signatures: transaction.signatures,
   };
 
   if (moduleAssetId === MODULE_ASSETS_NAME_ID_MAP.transfer) {
```
reyraa, line 238: A switch-case seems more readable here.

src/utils/api/transaction/lsk.js
```diff
@@ -219,51 +220,106 @@ const splitModuleAndAssetIds = (moduleAssetId) => {
...
+    const transaction = {
+      moduleID,
+      assetID,
+      senderPublicKey: Buffer.from(senderPublicKey, 'hex'),
+      nonce: BigInt(nonce),
+      fee: BigInt(fee),
+      signatures: [],
+    };
+
+    if (moduleAssetId === MODULE_ASSETS_NAME_ID_MAP.transfer) {
```
reyraa, line 290: A switch-case seems more readable here.

reyraa (changes requested):

src/utils/api/transaction/lsk.js
UsamaHameed: Already done in the next PR.

src/utils/api/transaction/lsk.js
UsamaHameed: Taken care of in the next PR.

reyraa (approved):

PR #3483, opened 2021-04-16
Determine non migrated accounts and redirect them to reclaim screen - Closes #3473

### LINKED ISSUE #3473: Determine non-migrated accounts and redirect them to the balance reclaim view
### Description
Once the user signed in, Lisk Desktop should determine if the account is not migrated yet, and prevent access to the wallet feature until the migration is accomplished.

### Motivation
Ensure users migrate their account to the new address system on a timely manner.

### Acceptance Criteria
After signed in:
- If the account did not exist in the old blockchain, it should navigate to the wallet page
- If the account is already migrated, if should navigate to the wallet page.
- If the account is not migrated, even if its balance is currently zero, it should navigate to the balance reclaim screen.

UsamaHameed:
### What was the problem?
This PR resolves #3473

### How was it solved?
- Redirect non-migrated account to the `/recalim` route
- Update unit tests

### How was it tested?
Unit test + Manually

### DIFF
--- a/src/store/middlewares/account.js
+++ b/src/store/middlewares/account.js
@@ -122,32 +122,12 @@ const autoLogInIfNecessary = async ({ dispatch, getState }) => {
   }
 };
 
-// eslint-disable-next-line max-statements
 const checkAccountInitializationState = (action) => {
-  const search = window.location.href.split('?')[1];
-
-  const initialization = selectSearchParamValue(search, 'initialization');
-  let serverPublicKey = '';
-  let balance = 0;
-  // let redirectUrl = '';
-
   if (action.type === actionTypes.accountLoggedIn) {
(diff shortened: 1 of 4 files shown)


isalga (approved):

src/store/middlewares/account.js
```diff
@@ -122,32 +121,12 @@ const autoLogInIfNecessary = async ({ dispatch, getState }) => {
   }
 };
 
-// eslint-disable-next-line max-statements
 const checkAccountInitializationState = (action) => {
```
reyraa, line 124: Please rename the function to a more relevant one.

reyraa (changes requested):

reyraa (approved):

PR #3491, opened 2021-04-19
Display old address in the wallet screen - Closes #3474 

### LINKED ISSUE #3474: Display old address in the wallet screen
### Description
Once Lisk blockchain migrates to Core v3, all transactions and account information will be displayed using the new address system. Users might need to access their old address to review their account history on a Explorer. The old address must be available through the address summary card in the wallet page. 

### Motivation
Enable users to access their account history.

### Acceptance Criteria
Users must be able to
  - Read and copy their new address.
  - Read and copy their old address.

reyraa:
### What was the problem?
This PR resolves #3474

### How was it solved?
- Refactored the AccountInfo component
- Updated tooltip colors, although they don't look appealing in many other places now. I've already notified Anna, and she'll give us a guideline with which I'll create another PR to fix the bookmarks.

### How was it tested?
Visually

### DIFF
--- a/src/components/screens/wallet/overview/accountInfo/identity.js
+++ b/src/components/screens/wallet/overview/accountInfo/identity.js
@@ -31,7 +31,6 @@ const WithTitle = ({
     </span>
     <Tooltip
       className={`${styles.tooltip} ${styles.secondary} ${styles.truncated}`}
-      title={t('Click to see full and old addresses')}
       content={(
         <span
           className={`${styles.text} ${styles.noSelect} account-secondary-truncated`}
@@ -40,10 +39,11 @@ const WithTitle = ({
           {truncateAddress(newAddress)}
         </span>
       )}
-    />
+    >
+      <span>{t('Click to see full and old addresses')}</span>
+    </Tooltip>
     <Tooltip
       className={`${styles.tooltip} ${styles.secondaryTooltip} ${styles.full}`}
-      title={t('Click to hide full and old addresses')}
       content={(
         <span
           className={`${styles.text} ${styles.noSelect} account-secondary-full`}
@@ -52,7 +52,9 @@ const WithTitle = ({
           {newAddress}
         </span>
       )}
-    />
+    >
+      <span>{t('Click to hide full and old addresses')}</span>
+    </Tooltip>
     <Legacy legacyAddress={legacyAddress} />
   </div>
 );
@@ -66,7 +68,6 @@ const WithOutTitle = ({
   <div className={styles.identity}>
     <Tooltip
       className={`${styles.tooltip} ${styles.primary} ${styles.truncated}`}
-      title={t('Click to see full and old addresses')}
       content={(
         <span
           className={`${styles.text} ${styles.primary} ${styles.noSelect} account-primary-truncated`}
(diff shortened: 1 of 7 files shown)


src/components/screens/wallet/overview/accountInfo/identity.js
```diff
@@ -1,54 +1,116 @@
...
+    <CopyToClipboard
+      text={legacyAddress}
+      className={styles.text}
+    >
+      {legacyAddress}
+    </CopyToClipboard>
+  </div>
+);
+
+const WithTitle = ({
```
UsamaHameed, line 20: Can't we abstract WithTitle and WithoutTitle into one component?

src/components/screens/wallet/overview/accountInfo/identity.js
```diff
@@ -1,54 +1,116 @@
...
-              onClick={onClick}
-            >
-              {isTruncated ? truncatedAddress : address}
-            </span>
-          )
-      }
-    </div>
-  );
+const Identity = (props) => {
+  if (!props.newAddress) {
```
UsamaHameed, line 107: This check should be done where the component is used, not inside the component.

src/components/screens/wallet/overview/accountInfo/index.js
```diff
@@ -1,144 +1,57 @@
...
+            address={address}
+            size={30}
+          />
+          <Identity
+            newAddress={address}
+            legacyAddress={account.summary.legacyAddress}
+            username={username}
+            bookmark={bookmark}
+            showLegacy={showLegacy}
+            onClickFn={onClick}
```
UsamaHameed, line 35: We use `onClick` everywhere :)

UsamaHameed (changes requested):

src/components/screens/wallet/overview/accountInfo/identity.js
reyraa: It was together first, it had too many conditions, I preferred a little bit of duplication over a soup of ternaries.

src/components/screens/wallet/overview/accountInfo/identity.js
reyraa: Merged the components.

isalga (approved):

src/components/screens/wallet/overview/accountInfo/identity.js
```diff
@@ -11,44 +25,74 @@ import styles from './accountInfo.css';
...
   username = '',
+  t = str => str,
+  setShowLegacy,
 }) => {
-  if (!address) return null;
-  const truncatedAddress = truncateAddress(address);
-  const [isTruncated, setTruncationState] = useState(true);
-  const onClick = () => setTruncationState(!isTruncated);
+  if (!newAddress) {
+    return null;
```
UsamaHameed, line 36: Its generally considered an anti pattern to have such a check inside the component. If we do this, the component will still be in the dom tree even if it returns `null`. I know we do this at other places as well but we have to slowly improve. Please let me know if you have a different POV. 👍

UsamaHameed (changes requested):

src/components/screens/wallet/overview/accountInfo/identity.js
reyraa: I totally agree with you. In fact, we should hand this in the HOC level and never get here if the account doesn't exist. or fill it with a placeholder to gracefully handle loading state. but at this time, there's a chain of loading-not-handling components which cause this component to crash before the account is loaded. 
Given the due dates, I rather handle these issues with [this](https://github.com/LiskHQ/lisk-desktop/issues/3485), whenever we decide it's time.

UsamaHameed: What I mean is, can't we do this in the render method of the parent? `newAddress && <Identity />`

reyraa: Sure. applying now

UsamaHameed (approved):

Issue #3493, opened 2021-04-20
Fix transaction transformers for unlock transaction


UsamaHameed:
### Expected behavior
The user should be able to send unlock transactions.

### Actual behavior
The users are unable to send unlock transactions.

Issue #3476, opened 2021-04-12
Locked vote amount is not visible


UsamaHameed:
### Expected behavior
The user should be able to see the locked vote amount in the wallet page. 

### Actual behavior
There is no amount displayed.

### Steps to reproduce
Vote for a delegate and then check your wallet page.

PR #3494, opened 2021-04-20
Fix unlock transaction transformers - Closes #3492, #3493, #3476

### LINKED ISSUE #3492: Fix transactions for uninitialized accounts
### Expected behavior
Users should be able to send transactions using the app.

### Actual behavior
Users with uninitialized accounts are unable to send any transactions.

UsamaHameed:
### What was the problem?
This PR resolves #3492, #3493, #3476

### How was it solved?
- Properly convert transactions for unlock 
- Properly extract public key for accounts that are uninitialized

### How was it tested?
Unit + Manually

### DIFF
--- a/src/utils/api/account/lsk.js
+++ b/src/utils/api/account/lsk.js
@@ -83,7 +83,7 @@ export const getAccount = async ({
     if (response.data[0]) {
       const account = { ...response.data[0] };
       const isAccountUninitialized = !account.summary.publicKey;
-      if (isAccountUninitialized) {
+      if (isAccountUninitialized && (params.publicKey || params.passphrase)) {
         const publicKey = params.publicKey ?? extractPublicKey(params.passphrase);
         account.summary.publicKey = publicKey;
       }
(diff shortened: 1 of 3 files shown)


isalga (dismissed):

src/utils/api/account/lsk.js
```diff
@@ -79,15 +80,21 @@ export const getAccount = async ({
       params: normParams,
     });
 
-    return response.data[0];
+    if (response.data[0]) {
+      const account = { ...response.data[0] };
+      const isAccountUninitialized = !account.summary?.publicKey;
```
reyraa, line 85: Since you assumed `response.data[0]` is an object, then the account existed and `account.summary` is an object too. You don't need the null check.

src/utils/api/account/lsk.js
```diff
@@ -79,15 +80,21 @@ export const getAccount = async ({
...
+    }
   } catch (e) {
     // eslint-disable-next-line no-console
     console.log('Lisk account not found.');
-    let publicKey = params.publicKey;
-    if (!publicKey && params.passphrase) {
-      publicKey = extractPublicKey(params.passphrase);
-    }
 
+    const publicKey = params.publicKey ?? extractPublicKey(params.passphrase);
```
reyraa, line 97: If the method is called with address and it returned 404, this assignment will throw. The previous one was better, although it didn't look as dope.

src/utils/api/account/lsk.js
```diff
@@ -79,15 +80,21 @@ export const getAccount = async ({
       params: normParams,
     });
 
-    return response.data[0];
+    if (response.data[0]) {
+      const account = { ...response.data[0] };
+      const isAccountUninitialized = !account.summary?.publicKey;
+      if (isAccountUninitialized) {
+        const publicKey = params.publicKey ?? extractPublicKey(params.passphrase);
```
reyraa, line 87: If the method is called with address and it's not initialized, this will throw.

reyraa (changes requested):

src/utils/api/account/lsk.js
UsamaHameed: Can you please explain 1. why do you think the assignment will throw 2. why the previous one was better (except readability)?

src/utils/api/account/lsk.js
UsamaHameed: This is fine. We can not convert the address to public key or passphrase ourselves. The consumer of the method needs to be aware of this. Thats why we throw an error at the end of the function.

reyraa (approved):

isalga (approved):

PR #3497, opened 2021-04-21
Validate addresses using functions from Lisk Elements - Closes #3490

### LINKED ISSUE #3490: Validate addresses using functions from Lisk Elements
### Expected behavior
Addresses should be validated using functions provided by Lisk Elements instead of regex.

### Actual behavior
Addresses are validated using regex but now that Lisk Elements provides a validation function, we can use that.

reyraa:
### What was the problem?
This PR resolves #3490

### How was it solved?
Used the address validator from the `@utils/validators` which in turn uses the validator from Lisk Elements.

### How was it tested?
Unit tests

### DIFF
--- a/src/components/shared/transactionTypeFigure/index.js
+++ b/src/components/shared/transactionTypeFigure/index.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { MODULE_ASSETS_NAME_ID_MAP, MODULE_ASSETS_MAP } from '@constants';
+import { MODULE_ASSETS_NAME_ID_MAP, MODULE_ASSETS_MAP, tokenMap } from '@constants';
 import { validateAddress } from '@utils/validators';
 import AccountVisual from '@toolbox/accountVisual';
 import Icon from '@toolbox/icon';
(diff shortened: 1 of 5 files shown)


src/components/shared/transactionTypeFigure/index.js
```diff
@@ -8,7 +9,7 @@ const TransactionTypeFigure = ({
   moduleAssetId, address, avatarSize = 40, className = '', icon,
 }) => {
   const renderAvatar = () => {
-    if (regex.address.test(address)) {
+    if (validateAddress(tokenMap.LSK.key, address) === 0) {
```
UsamaHameed, line 12: Wouldn't it be better if we return a boolean from `validateAddress()`?

UsamaHameed (changes requested):

src/components/shared/transactionTypeFigure/index.js
reyraa: There are three states:
-1: address blank. This is useful for not displaying an error, but also not enabling the submit/continue button
0: address valid
1: address invalid

UsamaHameed: If we want to use codes, I think you can use the following:
1: valid (no need to check against 1, its a truthy value)
0: invalid (no need to check against 0, its a falsy value)
-1: (invalid and) empty string (-1 is truthy but you would need to specifically check against it) 

simplifies our assertions in the components

UsamaHameed (approved):

isalga (approved):

PR #3498, opened 2021-04-21
Handle 404 and other errors for the accounts api better - Closes #3487

### LINKED ISSUE #3487: Handle `/accounts` 404 better
### Expected behavior
If we get a 404 error from the `/accounts/address=...` endpoint, that means the account does not exist on the blockchain yet but, if the passphrase is valid, we show the user an empty account. 

### Actual behavior
We show the empty account on every type of error that may happen during the network call.

UsamaHameed:
### What was the problem?
This PR resolves #3487

### How was it solved?
- Add http status to the error thrown
- Return empty account only when we get 404
- Add a constants file for http error codes

### How was it tested?
Manually

### DIFF
--- a/src/utils/api/account/lsk.js
+++ b/src/utils/api/account/lsk.js
@@ -1,4 +1,4 @@
-import { tokenMap, regex } from '@constants';
+import { HTTP_CODES, tokenMap, regex } from '@constants';
 import http from '../http';
 import ws from '../ws';
 import { isEmpty } from '../../helpers';
@@ -66,7 +66,7 @@ const getAccountParams = (params) => {
  *
  * @returns {Promise}
  */
-// eslint-disable-next-line max-statements
+// eslint-disable-next-line complexity, max-statements
 export const getAccount = async ({
   network, params, baseUrl,
 }) => {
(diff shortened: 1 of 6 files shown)


src/utils/api/account/lsk.js
```diff
@@ -79,26 +80,37 @@ export const getAccount = async ({
...
+            publicKey,
+            balance: 0,
+            address: normParams.address,
+            token: tokenMap.LSK.key,
+          },
+        };
+        return account;
+      }
+    } else {
+      throw Error();
```
reyraa, line 110: This error statement needs a message

src/utils/api/account/lsk.js
```diff
@@ -79,26 +80,37 @@ export const getAccount = async ({
...
+    }
   }
+  throw Error('Error retrieving account');
```
reyraa, line 113: In what condition the code reaches here?

reyraa (changes requested):

src/utils/api/account/lsk.js
UsamaHameed: The text would be the same as the outer error which actually catches this error (and has the required text) so I did not add any text.

src/utils/api/account/lsk.js
UsamaHameed: Ideally, never. But to keep the code branches clean, I added this error. There is also the possibility of a network failure and then the code execution will end up here.

src/utils/api/account/lsk.js
reyraa: Yes. I meant it seems like you've done a good job covering all the scenarios. Even at the event of network failure as you said in the next comment, it would end up here and this error will suffice. but not a big deal, I'll approve it.

reyraa (approved):

isalga (approved):

PR #3501, opened 2021-04-22
Change login action to use public key - Closes #3500

### LINKED ISSUE #3500: Change login action to use public key
### Description
The login action should make the call to `/accounts` with a public key. 

### Motivation
We get more search results when querying with public key. 

### Acceptance Criteria

Change the order of preference of using the params passed to accounts api methods.

UsamaHameed:
### What was the problem?
This PR resolves #3500

### How was it solved?
- make api calls to `/accounts` using public key as query
- change order of pereference of api params in `getAccountParams`

### How was it tested?
Unit + Manual


reyraa (approved):
Looks good @UsamaHameed

PR #3515, opened 2021-04-30
Fix issues identified in v2.0.0-beta.2 - Closes #3442

### LINKED ISSUE #3342: Transactions overview graphs are broken
### Expected behavior
The following graphs must display data correctly
- Distribution of transaction types
- Amount per transaction (LSK)
- Transactions number / volume (LSK)

### Actual behavior
The graphs render empty.
Screenshot 2021-02-03 at 13 06 17

### Steps to reproduce
- Connect to Core v3
- Navigate to transaction monitor page.

2.0.0-beta.1

### LINKED ISSUE #3442: Fix issues identified in v2.0.0-beta.2
### Expected behavior

1. ' Register as delegate' button should be disabled / invisible if the user is already registered as delegate
2. I should able to filter transactions based on the received messages from the transaction list
3. Don't allow users to duplicate vote transactions.  
4. Allow user to vote max of 10 votes and 10 unvotes in single vote transactions

### Actual behavior

1. Even though i'm a delegate, still i'm getting an option to 'Register as delegate' in delegate wallet page

image

2. In wallet page, message filter is not working as expected in transaction list

image

3. Voting queue is not getting cleared immediately after submitting my vote. Sometime i able to duplicate the vote transaction or sometimes im unable to do further voting. Try with more than 5 vote transactions to reproduce
4. 
image

4.  As per the requirement,  it is possible to vote up-to 10 delegate and unvote 10 delegate in single vote transaction but this is not working right now 
image

image

Version 2.0.0-beta.2

isalga:
### What was the problem?
This PR resolves #3342

### How was it solved?

- [x] Disable register delegate when user is delegate
- [x] Remove message filter
- [x] Disable voting modal when voting tx is pending
- [x] Remove amount converter in voting queue
- [x] Fix voting validation

### How was it tested?
Manually & Adding unit test


src/components/screens/votingQueue/editor/editor.js
```diff
@@ -44,17 +68,20 @@ const normalizeVotesForTx = votes =>
...
+  const votesStats = getVoteStats(votes);
 
   if (areVotesInValid) {
     messages.push(t('Please enter vote amounts for the delegates you wish to vote for'));
   }
 
-  if (Object.keys(votes).length > 10) {
+  if (Object.keys(votesStats.added).length > 10
+    || Object.keys(votesStats.edited).length > 10
+    || Object.keys(votesStats.removed).length > 10) {
```
reyraa, line 84: A user can't have more than 10 votes at a time. if you are editing or removing votes, it's an action on the already voted delegates which can't be more than 10. you need to make sure after submission, there's only 10 votes left. Something like:
```
Object.keys(votesStats.added).length + Object.keys(votesStats.edited).length <= 10
```

reyraa (changes requested):

reyraa (approved):

PR #3540, opened 2021-05-08
Fix the broken network selector in Sigin in page when the latest selected network is down - Closes #3538 

### LINKED ISSUE #3538: If official nodes are down, users can't connect to custom node
Lisk Dekstop v1.28.0

If all official nodes of Lisk are down, no page is loading when clicking on "Sign in". We just had an incedent where all nodes were down and users were unable to set custom nodes to connect to due to this poor design. Users should be able to set custom nodes in any case. Also I wasn't able to get to the settings page. Nothing happens if I click "settings".

Here's the issue:
explorer

reyraa:
### What was the problem?
This PR resolves #3538

### How was it solved?
The current implementation of the login page depends on the latest network respond.  I refactored it, so if the endpoint is down, you still can see the Sign in page, and select a new network.

### How was it tested?
Unit tests


JesusTheHun:
I would also add a link to "Find community nodes", pointing to a Lisk.io deep page, or a github README. So we don't see a flood of people spamming social channels asking for what node/URL to use

reyraa:
@JesusTheHun There's not enough space for it with the current design. but I'll consider this for 2.0.0. Thank you so the suggestion.

karmacoma (approved):

fchavant (approved):
`Jenkinsfile` changes LGTM

PR #3561, opened 2021-05-26
Create multisignature transactions - Closes #3528

### LINKED ISSUE #3528: Refactor the unlock transaction screen for multisignature accounts
### Description

We need to update the user journey for the unvote transaction. We need to enable the users to create a transaction and then copy/download the transaction and share it among their multisignature group members. Therefore, we need to add an additional step after the first one to display the TransactionSummary modal (which will already have implementation to display multisignature transactions).
[UI Design](https://www.figma.com/file/kNu67yxyOZCJKU21fgFvBl/Multisignature?node-id=1020%3A297)

### Acceptance Criteria
Enable multisignature accounts to create and send unvote transactions

isalga:
### What was the problem?
This PR resolves #3528 

### How was it solved?
- Receive transaction in summary screen
- Implement new multi-signature tx summary views

### How was it tested?
Manually, skipped test will be reimplemented when tx creation function is refactored.


src/components/screens/lockedBalance/summary/index.js
```diff
@@ -0,0 +1,59 @@
...
+    onClick: onSubmit,
+  };
+  const onCancelAction = {
+    label: t('Cancel'),
+    onClick: () => { prevStep(); },
+  };
+
+  return (
+    <TransactionSummary
+      title={t('Summary of delegate registration')}
```
reyraa, line 38: Summary of unlock LSK?

reyraa (changes requested):

reyraa (approved):

src/utils/api/transaction/lsk.js
```diff
@@ -332,26 +309,23 @@ export const createMultiSignatureTransaction = ({
   } = transactionObject;
 
   const schema = moduleAssetSchemas[moduleAssetId];
-  if (moduleAssetId === MODULE_ASSETS_NAME_ID_MAP.registerMultisignatureGroup) {
```
UsamaHameed, line 335: The mandatory an optional keys need to be a part of the asset in case of the `registerMultisignatureGroup` transaction

src/utils/api/transaction/lsk.js
```diff
@@ -332,26 +309,23 @@ export const createMultiSignatureTransaction = ({
...
-      schema,
-      transaction,
-      Buffer.from(networkIdentifier, 'hex'),
-      passphrase,
-      keys,
-      true,
-    );
+    let signedTransaction;
+
+    if (keys.numberOfSignatures > 0) {
```
UsamaHameed, line 318: Would the numberOfSignatures property be defined in accounts api response in case of a non multisignature account?

UsamaHameed (changes requested):

UsamaHameed (changes requested):
It would be great if we could test all of the transactions for non-multisignature accounts as well. Due to the lack of e2e testing, we might have introduced some regressions that we are unaware of.

src/utils/api/transaction/lsk.js
isalga: Well spot. No, it will be defined as 0. You can check it here in this call http://service-v5.liskdev.net/api/v2/accounts?publicKey=1849c3a63b7c336fec832fbf394457e86b9c3b2ceef6e1029fd4c0c35b16ed88

reyraa (approved):

UsamaHameed (approved):

PR #3569, opened 2021-05-26
The transaction download prompter opens every 10 second - Closes #3563

### LINKED ISSUE #3563: The transaction download prompter opens every 10 second
### Expected behavior
The prompt view should open just once, and automatically close once the user clicks on save.

### Actual behavior
The prompter opens every 10 seconds. This happens because with every new block the component re-renders. 

### Steps to reproduce
- Create a new transaction using a multisignature account.
- Click on download button.
- Save the file but don't close the transaction summary screen.
- The download prompt with open again.

reyraa:
### What was the problem?
This PR resolves #3563

### How was it solved?
- Removed unnecessary props, one of which was the confirmations which caused the 10 seconds re-render.

### How was it tested?
Unit tests + manually


UsamaHameed:
Not needed anymore. We aren't able to reproduce the issue.

PR #3573, opened 2021-05-30
Fix issues identified in v2.0.0-beta.3

### LINKED ISSUE #3572: Fix issues identified in v2.0.0-beta.3
### Actual behavior

1. Unable to down the paper wallet while creating new account.
image

2. In wallet page, if there are no transaction then we should display a text informing the user that there is no transaction.
image

3. Unable to edit the bookmarked(created through 'Add new' button) delegates name
image

4. Unable to search through block height
image

5. Last forged block details are incorrect in delegate profile page
image

6. Incoming and outgoing transaction representation is incorrect for down vote  and unlock transactions
image

7. Productivity and rank details are missing in the voted delegates page
image

8. Unlock transaction amount is incorrect and also the transaction representations are incorrect
image

9. Alert message is getting displayed even if i have enough balance. 
Steps to reduce --> Enter 1LSK, select  confirmation button, select the edit button from the summary page
image

10. In transaction monitoring page, the filtered results are getting rest when user select the load more button.
image

11.  Remove watch list functionality is not working (inside and watched tab)
image

12. Forging time is incorrect in inside and watched tab. The forging time is showing only the future time.

image

13. Alignment needs to be fixed
image
14. Tool tip in transaction monitoring page shows undefined 
image
15. Create new account, transfer LSK using faucet and now try to transfer tokens to other account. The transaction fee is always zero.
16. Can we change the icon for the delegate registration ? Current icon might confuse the user.
image
17. Unable to vote for 10 delegates in single vote transaction. 
Steps to reproduce --> Add 11 delegates into voting queue and then delete 1 delegate from the queue
image

18. In voting summary page, we should display the delegates name. Right now we are displaying delegates accounts number.  
image

reyraa:
### What was the problem?
This PR resolves #3572

### How was it solved?
- [x] Fix broken paper wallet: Updated the JSDPF lib usage.
- [x] Make delegate bookmarks uneditable: The boolean check was not correct.
- [x] Fix the direction icon in recent transaction module: Host address was undefined.
- [x] Fix unlocking amount calculation in recent transactions: The values were treated as strings.
- [x] Fix tx type filter: There was an issue with filtering the blank values.
- [x] Added vote list addresses in voting queue do not translate to usernames

### How was it tested?
Unit tests


src/components/screens/bookmarks/addBookmark/addBookmark.js
```diff
@@ -99,7 +99,7 @@ const AddBookmark = ({
       {
         value: usernameValue,
         feedback: usernameFeedback,
-        readonly: isDelegate === true || usernameFeedback !== '',
+        readonly: isDelegate === 'true',
```
UsamaHameed, line 102: should it be the string `'true'` or a boolean `true`?

reyraa: It comes from the query parameters. It's a string.

UsamaHameed (approved):

PR #3581, opened 2021-06-03
Fix remaining issues for Lisk Desktop v2.0.0-beta.4

### LINKED ISSUE #3580: Fix remaining issues for Lisk Desktop v2.0.0-beta.4
### Description
The following issues are discovered in the multisignature tx development phase:

- [x] The members in Multisig group registration summary screen only displays the public key. We should derive and display the address and avatar.
- [x]  When you're the last member to sign, the Multisig tx status screen doesn't have a Send button and there's no back button to get back to the summary screen that does.
- [x]  The back to wallet button doesn't actually take to the wallet page
- [x] Trying to sign a transaction with an account that does not belong to a group it visually displays that has been signed succesfully. Output transaction remains not signed, its an UI issue
- [x] The tx id should be removed from Send summary.
- [x] The divider lines in tx summary and tx details view need a review (CSS issue) (Ali)
- [x] Delegate registration with normal account fails because the fee is less than minimum required. (I've registered 5 delegates so far and it works. Unable to reproduce)
- [x] Download tx used module asset id, it would be better to use tx id.
- [ ] Users should not be able to sign a fully signed transaction. (Usama)
- [x] Missing delegate registration icon in the transaction sign modal for delegate registration transactions. (Usama)
- [x] Voting queue crashes when I try to vote with an account that has never voted before. `account.dpos.sentVotes` is undefined. (Usama)
- [x] Transaction download fails because the id is malformed (Usama)
- [x] Dark mode colour adjustments needed all across the multisignature UI (Ali)
- [x] Remove register multisignature account button on wallet pages other than the logged in user's wallet page (Usama)
- [x] The confirm button in the first step of "Register multisignature account" modal should be inactive if the form is invalid (Ali).
- [x] As the last member I don't see the send button when registering a multisignature group (Ali)
- [x] The member type (optional/mandatory) in "Register multisignature group transaction summary" is wrong (Ali)
- [x] The copy vote tx button sometimes copies the last send tx (Ali)

isalga:
### What was the problem?
This PR resolves #3580 

### How was it solved?

- [x] The members in Multisig group registration summary screen only displays the public key. We should derive and display the address and avatar.
- [x] When you're the last member to sign, the Multisig tx status screen doesn't have a Send button and there's no back button to get back to the summary screen that does.
- [x] The back to wallet button doesn't actually take to the wallet page
- [x] Trying to sign a transaction with an account that does not belong to a group it visually displays that has been signed succesfully. Output transaction remains not signed, its an UI issue
- [x] The tx id should be removed from Send summary.
- [ ] The divider lines in tx summary and tx details view need a review (CSS issue)
- [x] Delegate registration with normal account fails because the fee is less than minimum required. (This seems to have been a run time error. I managed to register delegates several time after that without any issues) 
- [x] Download tx used module asset id, it would be better to use tx id.


src/components/shared/transactionSummary/index.js
```diff
@@ -21,8 +21,9 @@ const Footer = ({
 }) => {
   const [copied, setCopied] = useState(false);
 
-  const onDownload = (transaction) => {
-    downloadJSON(transaction, `tx-${transaction.moduleID}-${transaction.assetID}`);
+  const onDownload = (transaction = {}) => {
+    const jsonTransaction = JSON.parse(transactionToJSON(transaction));
+    downloadJSON(transaction, `tx-${jsonTransaction.id}`);
```
UsamaHameed, line 26:
```suggestion
    downloadJSON(transaction, `tx-${transaction.id}`);
```

UsamaHameed (changes requested):

src/components/screens/signMultiSignTransaction/reviewSign/reviewSign.js
```diff
@@ -56,37 +54,6 @@ const flattenTransaction = ({ moduleAssetId, asset, ...rest }) => {
   return transaction;
 };
 
-const getNumbersOfSignaturesRequired = ({ keys, isGroupRegistration }) => {
```
UsamaHameed, line 59: Why were these removed? I added this as a fix in https://github.com/LiskHQ/lisk-desktop/pull/3579

src/components/screens/signMultiSignTransaction/reviewSign/reviewSign.js
```diff
@@ -95,44 +62,16 @@ const ReviewSign = ({
...
   history,
   error,
-  dispatch,
   senderAccount,
 }) => {
-  // eslint-disable-next-line max-statements
-  const showSendButton = () => {
+  const isMember = useMemo(() => {
     if (senderAccount.data?.keys) {
-      const isGroupRegistration = transaction.moduleAssetId
```
UsamaHameed, line 104: Why were these removed? I added this as a fix in #3579

UsamaHameed (changes requested):

src/components/screens/signMultiSignTransaction/reviewSign/reviewSign.js
isalga: As I observe they were moved to helpers. @reyraa is it right?

UsamaHameed (approved):

PR #3585, opened 2021-06-07
Fix beta.4 issues

### LINKED ISSUE #3580: Fix remaining issues for Lisk Desktop v2.0.0-beta.4

UsamaHameed:
### What was the problem?
This PR resolves #3580

### How was it solved?

- Missing delegate registration icon in the transaction sign modal for delegate registration transactions.
- Voting queue crashes when I try to vote with an account that has never voted before. account.dpos.sentVotes is undefined. (Usama)
- Transaction download fails because the id is malformed
- Remove register multisignature account button on wallet pages other than the logged in user's wallet page

### DIFF
--- a/src/components/screens/signMultiSignTransaction/share/share.js
+++ b/src/components/screens/signMultiSignTransaction/share/share.js
@@ -58,7 +58,8 @@ const Share = ({
   const [copied, setCopied] = useState(false);
 
   const onDownload = () => {
-    downloadJSON(transaction, `tx-${transaction.id}`);
+    const id = transaction.id.toString('hex');
+    downloadJSON(transaction, `tx-${id}`);
   };
 
   const onCopy = () => {
--- a/src/components/screens/transactionDetails/layoutSchema.js
+++ b/src/components/screens/transactionDetails/layoutSchema.js
@@ -37,7 +37,7 @@ const LayoutSchema = {
     className: styles.registerDelegate,
   },
   [`${registerDelegate}-preview`]: {
-    components: [...previewBaseComponents],
+    components: [...previewBaseComponents, Illustration],
     className: styles.registerDelegatePreview,
   },
   [registerMultisignatureGroup]: {
--- a/src/components/shared/transactionInfo/voteDelegate.js
+++ b/src/components/shared/transactionInfo/voteDelegate.js
@@ -33,6 +33,7 @@ const VoteDelegate = ({
   const addedLength = Object.keys(added).length;
   const editedLength = Object.keys(edited).length;
   const removedLength = Object.keys(removed).length;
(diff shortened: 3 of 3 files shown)


src/components/screens/wallet/overview/accountInfo/actionBar.js
```diff
@@ -117,20 +117,22 @@ const ActionBar = ({
...
-        content={(
-          <DialogLink component={isMultisignature ? 'multisigAccountDetails' : 'multiSignature'}>
-            <Icon name="multiSignatureOutline" className={styles.multisigIcon} />
-          </DialogLink>
-        )}
-      >
-        <p>{isMultisignature ? t('View multisignature account details') : t('Register multisignature')}</p>
-      </Tooltip>
-    </div>
+    {address === host && (
```
reyraa, line 120: Should we display members of a multisig group that is not the one we've signed in with? If yes, then we should change this condition to always show `multisigAccountDetails` and only show `multiSignature` for the host.

reyraa (approved):

PR #3587, opened 2021-06-09
Fix dividers of the transaction details screen

reyraa:
### What was the problem?
This PR resolves some issue described in #3580

### How was it solved?
- [x] The divider lines in tx summary and tx details view need a review (CSS issue)
- [x] The copy vote tx button sometimes copies the last send tx : We were using the first member of `ransactions.transactionsCreated`. In multisig groups, the transactions create are never directly broadcasted, therefore this array doesn't get cleaned. next time that you initiated a new transaction, the `TransactionSummary` component should get the last (most recent) member of this array. We were using the first (oldest) member.

### How was it tested?
Visually

### DIFF
--- a/src/components/screens/transactionDetails/layoutSchema.js
+++ b/src/components/screens/transactionDetails/layoutSchema.js
@@ -20,7 +20,7 @@ const restComponents = [TransactionId, Fee, SignedAndRemainingMembersList];
 const LayoutSchema = {
   [transfer]: {
     components: [...baseComponents, Recipient, Amount, Message, ...timeComponents],
-    className: '',
+    className: styles.transferLayout,
   },
   [`${transfer}-preview`]: {
     components: [...previewBaseComponents, Recipient, Amount, Message, ...restComponents],
--- a/src/components/screens/transactionDetails/layoutSchema.js
+++ b/src/components/screens/transactionDetails/layoutSchema.js
@@ -12,48 +12,50 @@ const {
   // reclaimLSK,
 } = MODULE_ASSETS_NAME_ID_MAP;
 
-const baseComponents = [Sender, Confirmations, TransactionId, Fee, Date, BlockId, BlockHeight];
-const previewBaseComponents = [Sender, TransactionId, Fee, SignedAndRemainingMembersList];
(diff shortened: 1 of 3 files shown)


isalga (dismissed):

src/components/screens/transactionDetails/layoutSchema.js
```diff
@@ -12,48 +12,50 @@ const {
...
+const baseComponents = [Illustration, Sender];
+const timeComponents = [TransactionId, Date, BlockId, BlockHeight, Fee, Confirmations];
+const previewBaseComponents = [Illustration, Sender];
+const restComponents = [TransactionId, Fee, SignedAndRemainingMembersList];
 
 const LayoutSchema = {
   [transfer]: {
-    components: [...baseComponents, Recipient, Illustration, Amount, Message],
-    className: '',
+    components: [...baseComponents, Recipient, Amount, Message, ...timeComponents],
```
UsamaHameed, line 22: The order in which the layout is shown does not depend on the order of components here. If this is just for cleanup, then its fine

UsamaHameed (changes requested):

src/components/screens/transactionDetails/transactionDetails.css
```diff
@@ -107,21 +123,38 @@
...
 
+  & .unlockToken {
+    grid-template-areas:
+      "illustration illustration"
+      "sender sender"
+      "amount amount"
+      "transactionId date"
+      "blockId blockHeight"
+      "fee confirmations"
+      "signedAndRemainingMembersList signedAndRemainingMembersList";
```
UsamaHameed, line 134: Is the signedAndRemainingMembersList needed for the unlockToken transaction?

reyraa: no this was a mistake. I removed it.

src/components/screens/transactionDetails/layoutSchema.js
reyraa: Mostly clean up. It helped to have a more unified detail view

UsamaHameed (approved):

PR #3609, opened 2021-06-11
Fix delegate table alignments issues - Closes #3428

### LINKED ISSUE #3428: Fix issues identified in v2.0.0-alpha.5
### Expected behavior

1. The `Latest votes` tab should have the same column alignment/padding as other tabs
2. The contents of the `Watched` tab should contain "-" when no data is available
3. At the smallest width, the (`Inside round`, `Outside round`, `Sanctioned`...) tabs should not overlap 
4. The "hitbox" for adding/removing to/from the `Watched` tab should not be prone to user error

### Actual behavior

1.  The columns of the `Latest votes` tab have different column alignment/padding

Screenshot 01 - Latest votes

2.  The contents of the `Watched` tab are left empty (this is inconsistent with other parts of the application)

Screenshot 02 - Tabs - Watched contents

3. The margin between tab and listing overlaps

Screenshot 03 - Tabs - Smallest width

4. The "hitbox" for adding/removing to/from the watched tab is too small

Screenshot 04 - Tabs - Watched hitbox
 
### Steps to reproduce

1. Open Lisk Desktop
2. Navigate to the `Delegates` page
3. Inspect the issues as reported

- Version 2.0.0-alpha.5
- OS: macOS 11.2.2

isalga:
### What was the problem?
This PR resolves #3428

### How was it solved?
Adjust CSS

### How was it tested?
Visually


reyraa (changes requested):
This fixes issues in the delegate page. But the other pages remain unaligned. 
Please fix this more generically so all the pages are fixed.

reyraa (approved):

PR #3612, opened 2021-06-14
Fix multisignature minium fee is calculation - Closes #3584

### LINKED ISSUE #3584: Multisignature minium fee is calculated slightly high
### Expected behavior
The minimum fee should take the following signatures into account:
- 64 byte signatures in the number of signatures required.
- Empty string signature placeholders for optional members who didn't sign.

### Actual behavior
Lisk Elements 5.1.0 has a known bug. It doesn't take the signature placeholders into account. therefore the calculated minFee is lower than expected by Lisk Core.
We are calculating the minFee for full signatures for every single member. This is a temporary fix and we need to revert it once Lisk Elements is fixed.

### Dependency 
https://github.com/LiskHQ/lisk-sdk/issues/6470

reyraa:
### What was the problem?
This PR resolves #3584

### How was it solved?
Updated Lisk Client and reverted the quick work around.

### How was it tested?
Manually.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@ledgerhq/hw-transport-node-hid": "5.50.0",
         "@ledgerhq/hw-transport-u2f": "5.33.1-alpha.3",
-        "@liskhq/lisk-client": "5.0.1",
+        "@liskhq/lisk-client": "5.1.1",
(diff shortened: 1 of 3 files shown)


reyraa:
To fix the broken tests we need to update Jest for which at this point we don't have enough time. I'll close this PR until after the RC release.

PR #3619, opened 2021-06-17
Fix reclaim transaction - Closes #3588

### LINKED ISSUE #3588: Reclaim transaction fails
### Expected behavior
Users should be able to reclaim their accounts after the regenesis.

### Actual behavior
1. Login is successful before the reclaim transaction is made
2. I am able to send the reclaim transaction after a small deposit is made to the new account address
3. On reclaim transaction submission, an error popup shows stating "Transaction failed", a while after which a loading screen appears
4. Upon re-login, the loading screen continues to appear. But everything works normally after the client is restarted.
5. The transaction history confirms the successful reclaim transaction for the given account, which can be viewed without logging into the wallet

### Steps to reproduce
https://trello.com/c/EcyNMzrM

isalga:
### What was the problem?
This PR resolves #3588

### How was it solved?
- Fix transaction
- Fix multistep modal crash

### How was it tested?
- Update unit test
- Using alphanet-v5-legacy-service

### DIFF
--- a/src/components/screens/reclaimBalance/modal/status/index.js
+++ b/src/components/screens/reclaimBalance/modal/status/index.js
@@ -4,7 +4,7 @@ import { transactionBroadcasted } from '@actions';
 import Status from './status';
 
 const mapStateToProps = state => ({
-  isMigrated: state.account.info?.LSK?.summary.isMigrated,
+  isMigrated: state.account.info.LSK.summary.isMigrated,
   transactions: state.transactions,
   network: state.network,
 });
--- a/src/components/screens/reclaimBalance/modal/status/index.js
+++ b/src/components/screens/reclaimBalance/modal/status/index.js
@@ -1,15 +1,17 @@
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router';
-import { transactionBroadcasted } from '@actions';
+import { transactionBroadcasted, accountDataUpdated } from '@actions';
 import Status from './status';
 
 const mapStateToProps = state => ({
+  isMigrated: state.account.info.LSK.summary.isMigrated,
   transactions: state.transactions,
   network: state.network,
 });
 
 const mapDispatchToProps = {
   transactionBroadcasted,
+  accountDataUpdated,
 };
 
 export default withRouter(connect(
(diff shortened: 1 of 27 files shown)


src/components/screens/reclaimBalance/modal/status/index.js
```diff
@@ -4,6 +4,7 @@ import { transactionBroadcasted } from '@actions';
 import Status from './status';
 
 const mapStateToProps = state => ({
+  isMigrated: state.account.info?.LSK?.summary.isMigrated,
```
reyraa, line 7: If the account wasn't signed in, i.e. there was no `state.account.info.LSK` we shouldn't even see this page. The null check seems redundant.

src/components/screens/reclaimBalance/modal/summary/index.js
```diff
@@ -31,23 +31,25 @@ const Summary = ({
...
+      amount: account.info.LSK.legacy?.balance,
     },
   });
 
   const onSubmit = async () => {
     const data = {
       network,
+      account: {
+        ...account.info.LSK,
+        sequence: { nonce: 0 },
```
reyraa, line 45: the API seems to return the sequence. Why did you do hard code it here?

src/components/screens/reclaimBalance/modal/summary/index.js
```diff
@@ -31,23 +31,25 @@ const Summary = ({
...
+      },
       passphrase: account.passphrase,
-      account: account.info.LSK,
       transactionObject: {
         moduleAssetId,
-        senderPublicKey,
-        nonce: account.info.LSK.sequence.nonce,
         fee: toRawLsk(minFee.value),
-        amount: account.info.LSK.legacy.balance,
+        amount: account.info.LSK.legacy?.balance,
```
reyraa, line 51: The null check seems redundant. Aside from the fact that if `account.info.LSK.legacy` was `undefined` we wouldn't visit this page, the transaction fails if the amount if not defined. So we have to make sure `account.info.LSK.legacy.balance` if at this point defined and is a number.

src/components/shared/transactionInfo/reclaim.js
```diff
@@ -11,7 +11,7 @@ const Reclaim = ({ account, t }) => (
     <section>
       <label>{t('Balance to reclaim')}</label>
       <LiskAmount
-        val={Number(account.info.LSK.legacy.balance)}
+        val={Number(account.info.LSK.legacy?.balance)}
```
reyraa, line 14: Similar to other null checks.

reyraa (approved):

PR #3645, opened 2021-06-24
Fix new block and new transaction buttons - Closes #3643

### LINKED ISSUE #3643: New block and new transaction buttons do not appear
### Expected behavior
- The new block button in monitor/blocks page should show up when there's a new block.
- The new transaction button in monitor/transactions page should show up when there's a new transaction.

### Actual behavior
They never appear.

### Steps to reproduce
- Visit monitor/blocks and monitor/transactions.
- Keep the page open for a while.

reyraa:
### What was the problem?
This PR resolves #3643

### How was it solved?
The subscriber was outdated and didn't work with the new API structure. It must have been broken since 2.0.0-beta.0. I removed the outdated hooks and used the latest blocks data from the Redux store to display the buttons.

### How was it tested?
Unit tests are updated and the app is tested manually.


src/components/screens/monitor/blocks/blocks.js
```diff
@@ -53,7 +53,7 @@ const Blocks = ({
           <BlockFilterDropdown filters={filters} applyFilters={applyFilters} />
         </BoxHeader>
         <LoadLatestButton
-          event="update.block"
+          entity="block"
```
isalga, line 56: Can we make this "block" string a constant and import it? That would be more scalable and not prone to crash.

isalga (changes requested):

src/components/screens/monitor/blocks/blocks.js
reyraa: I do agree we should prevent mistakes there, but I think propTypes is a better choice than constants.

isalga (approved):

isalga (approved):

PR #3666, opened 2021-06-30
Revalidate multisignature transaction - Closes #3593

### LINKED ISSUE #3593: The sign multisignature modal doesn't re-validate after input value changes
### Expected behavior
If I input an invalid value into the textarea, and then fix its issues or input a new valid transaction, the error message should disappear and the CTA must get activated.

### Actual behavior
The CTA remains disabled and the error message doesn't disappear.

2.0.0-beta.4

isalga:
### What was the problem?
This PR resolves #3593 

### How was it solved?
- Reset state when new transaction is inputted.

### How was it tested?
Copy/paste a invalid transaction, then copy/paste a valid one.

### DIFF
--- a/src/components/screens/signMultiSignTransaction/importData/importData.js
+++ b/src/components/screens/signMultiSignTransaction/importData/importData.js
@@ -23,34 +23,30 @@ const ImportData = ({ t, nextStep }) => {
     }
   };
 
-  const resetState = () => {
-    setTransaction(undefined);
-    setError(undefined);
-  };
-
   const validateAndSetTransaction = (input) => {
     try {
       const parsedInput = JSON.parse(input);
       setTransaction(parsedInput);
+      setError(undefined);
     } catch (e) {
-      setTransaction(input);
+      setTransaction(undefined);
       setError('Invalid transaction');
     }
   };
 
   const onFileInputChange = ({ target }) => reader.readAsText(target.files[0]);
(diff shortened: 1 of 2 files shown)


UsamaHameed (changes requested):
I came across two issues. 

1. When I paste an invalid transaction, the "upload" icon in the background stays. It should not be there as there is already a transaction pasted.

2. After pasting a valid tranaction, if I paste an invalid one, I see the error feedback but the content inside the textarea stays the same i.e. its still the valid transaction that I had before. It should be the invalid transaction instead.

src/components/screens/signMultiSignTransaction/importData/importData.js
```diff
@@ -23,23 +23,31 @@ const ImportData = ({ t, nextStep }) => {
...
+    setTransaction(undefined);
+    setError(undefined);
+  };
+
   const validateAndSetTransaction = (input) => {
     try {
       const parsedInput = JSON.parse(input);
       setTransaction(parsedInput);
     } catch (e) {
+      setTransaction(input);
```
UsamaHameed, line 36: I think it might be better to set this to `setTransaction(undefined)` instead to avoid this behaviour where the clipboard or the textarea component seems to have added some escapes

UsamaHameed (changes requested):

UsamaHameed (approved):

PR #3689, opened 2021-07-14
Fix visual issues in the wallet section - Closes #3654

### LINKED ISSUE #3654: Design issues in wallet page
1. In wallet page, if there are no transaction then we should display a text informing user that there is no transaction.
image

2. In dashboard  page, if there are no transaction then we should display a text informing user that there is no transaction.
image

3. In wallet page, copy address & copy public key panel transparency should be changed
image

4. Incoming and outgoing transaction representation is incorrect for down vote and unlock transactions. This should be applied for  wallet page and dashboard page
image

5. In wallet page --> delegates profile page, can we have '-' if the last forged block  is empty
image

6. In Voting queue page, the text standards are missing
image

7. 'Load more' behaviour is inconsistent in latest votes page. By default it should have 20 records, then load more functionality should add 20 more records.

UsamaHameed:
### What was the problem?
This PR resolves #3654

### How was it solved?

- [x] In wallet page, if there are no transaction then we should display a text informing user that there is no transaction.
- [x] In dashboard page, if there are no transaction then we should display a text informing user that there is no transaction.
- [ ] ~~In wallet page, copy address & copy public key panel transparency should be changed.~~
- [ ] ~~Incoming and outgoing transaction representation is incorrect for down vote and unlock transactions. This should be applied for wallet page and dashboard page~~
- [x] In wallet page --> delegates profile page, can we have '-' if the last forged block is empty
- [ ] ~~In Voting queue page, the text standards are missing~~

### How was it tested?
Manually

### DIFF
--- a/src/components/screens/dashboard/recentTransactions/recentTransactions.js
+++ b/src/components/screens/dashboard/recentTransactions/recentTransactions.js
@@ -40,7 +40,7 @@ const RecentTransactions = ({ className, t, transactions }) => {
   const [isLoaded, setLoaded] = useState(!!transactions.data.length);
   const settings = useSelector(state => state.settings);
   const currentBlockHeight = useSelector(selectCurrentBlockHeight);
-  const activeToken = tokenMap[settings.token.active];
+  const activeToken = tokenMap[settings.token.active].key;
   const host = account.info && account.info[activeToken] ? account.info[activeToken].summary.address : '';
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/components/screens/dashboard/recentTransactions/recentTransactions.js
```diff
@@ -40,7 +40,7 @@ const RecentTransactions = ({ className, t, transactions }) => {
   const [isLoaded, setLoaded] = useState(!!transactions.data.length);
   const settings = useSelector(state => state.settings);
   const currentBlockHeight = useSelector(selectCurrentBlockHeight);
-  const activeToken = tokenMap[settings.token.active];
+  const activeToken = tokenMap[settings.token.active].key;
```
reyraa, line 43: `tokenMap[settings.token.active].key` is exactly the same as `settings.token.active`.

reyraa (changes requested):

reyraa (approved):

PR #3696, opened 2021-07-16
Accept senderAddress and recipientAddress as filters in api/transactions - Closes #3652

### LINKED ISSUE #3652: Filter is not working as expected in transaction monitoring page
### Expected behavior
In transaction monitoring page, wildcard filter (AND condition) should work when we filter with more than one criteria.

### Actual behavior
In transaction monitoring page, select send transaction type and input amount range and input recipient address. When i try to filter with multiple inputs then it is not working as expected.

Mac, v2.0.0-rc0

UsamaHameed:
### What was the problem?
This PR resolves #3652

### How was it solved?

Use filters in api call correctly

### How was it tested?

Manually

### DIFF
--- a/src/components/screens/monitor/transactions/transactions.js
+++ b/src/components/screens/monitor/transactions/transactions.js
@@ -14,12 +14,12 @@ const fields = t => [{
 }, {
   label: t('Address'),
   placeholder: t('Address or public key'),
-  name: 'address',
+  name: 'senderAddress',
   type: 'address',
 }, {
   label: t('Recipient'),
   placeholder: t('Address or public key'),
-  name: 'recipient',
+  name: 'recipientAddress',
   type: 'address',
 }, {
   label: t('Type'),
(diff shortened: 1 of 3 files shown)


reyraa:
Good job Usama. The recipient filter is working, although I detected a couple of issues:
 - Since you changed the functionality from `address` which includes both `senderAddress` and `recipientAddress`. The _address_ title should change to _sender_. There are two places to take care: A. The filter dialog, B. The filter bar above the tx list.
 - The filter dialog doesn't reset when you clean the filter(s) all together or one by one using the filter bar.

UsamaHameed:
> Good job Usama. The recipient filter is working, although I detected a couple of issues:
> 
> * Since you changed the functionality from `address` which includes both `senderAddress` and `recipientAddress`. The _address_ title should change to _sender_. There are two places to take care: A. The filter dialog, B. The filter bar above the tx list.
> * The filter dialog doesn't reset when you clean the filter(s) all together or one by one using the filter bar.

Done ✅

reyraa (approved):

PR #3715, opened 2021-07-26
Recalculate transaction fee when recipientAddress changes - Closes #3712

### LINKED ISSUE #3712: Fee is not being recalculated when the recipient is put as last
### Expected behavior

The fee should be recalculated

### Actual behavior

The fee is not being recalculated

### Steps to reproduce

Go to the _Send LSK_ window, instert the amount first (optionally type a message), insert a recipient address. The fee is not updated.

2.0.0-rc.3

UsamaHameed:
### What was the problem?
This PR resolves #3712

### How was it solved?

Fix the dependency

### How was it tested?

Manually

### DIFF
--- a/src/components/screens/monitor/delegates/delegatesTable/dataColumns.js
+++ b/src/components/screens/monitor/delegates/delegatesTable/dataColumns.js
@@ -138,11 +138,23 @@ export const RoundState = ({
   );
 };
 
+const getDelegateStatus = (key, totalVotesReceived) => {
+  if (key === 'banned' || key === 'punished') {
+    return [key, delegateStatus[key]];
+  }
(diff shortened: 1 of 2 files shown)


src/components/screens/monitor/delegates/delegatesTable/dataColumns.js
```diff
@@ -138,11 +138,23 @@ export const RoundState = ({
   );
 };
 
+const getDelegateStatus = (key, totalVotesReceived) => {
+  if (key === 'banned' || key === 'punished') {
+    return [key, delegateStatus[key]];
+  }
```
reyraa, line 144: Why have your separated this from the default branch in line 149?

reyraa (approved):

PR #3729, opened 2021-08-13
Display a warning message when user sends the entire balance - Closes #3366

### LINKED ISSUE #3366: Display a warning message when user sends the entire balance
### Description
Although "Send entire balance" button makes it easy to send maximum possible fund without having to subtract the tx fee from available balance manually, it increases the chance of unintentionally sending your entire funds to someone else.
Given the importance of the consequences, a warning message would be helpful.

### Motivation
Eliminate gross errors. 

### Acceptance Criteria
Lisk Desktop should display a warning message when users click on "send entire balance".

isalga:
### What was the problem?
This PR resolves #3366

### How was it solved?
- Add warning

### How was it tested?
Add unit test

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -579,6 +579,7 @@
   "Write message": "Write message",
   "Written by": "Written by",
   "YYYY": "YYYY",
+  "You are about to send your entire balance": "You are about to send your entire balance",
   "You are disconnected": "You are disconnected",
   "You can also download, print and store safely your passphrase.": "You can also download, print and store safely your passphrase.",
   "You can learn more": "You can learn more",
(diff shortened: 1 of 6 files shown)


MichalTuleja (changes requested):
The behaviour of that message box is correct, but not very intuitive:
- Making that message disappear also makes the amount disappear
- Changing the amount to lower value does not cause the message disappear

Is it possible to track the amount and display the message based on that input?

sridharmeganathan (dismissed):

sridharmeganathan (approved):

MichalTuleja (approved):
Looks good 👍  Thanks @isalga

PR #3775, opened 2021-09-08
Fix BTC address calculation - Closes #3772

### LINKED ISSUE #3772: Fix BTC address calculation
### Expected behavior
We should calculated all addresses correctly.

### Actual behavior
The BTC address is calculated wrong.

2.0.0

reyraa:
### What was the problem?
This PR resolves #3772

### How was it solved?
Since we added an API response normalizer to the BTC API, we don't need to distinguish the responses in the utility level.

### How was it tested?
unit tests

### DIFF
--- a/src/store/actions/account.js
+++ b/src/store/actions/account.js
@@ -60,22 +60,17 @@ const getAccounts = async ({ network, params }) =>
  */
 export const accountDataUpdated = tokensTypes =>
   async (dispatch, getState) => {
-    const state = getState();
-    const { network, settings, account } = state;
+    const { network, settings, account } = getState();
+
+    // Get the list of tokens that are enabled in settings
     const activeTokens = tokensTypes === 'enabled'
       ? Object.keys(settings.token.list)
         .filter(key => settings.token.list[key])
       : [settings.token.active];
 
+    // Collect their addresses to send to the API
     const params = activeTokens.reduce((acc, token) => {
-      if (token === tokenMap.LSK.key) {
-        acc[token] = { address: account.info[tokenMap.LSK.key].summary.address };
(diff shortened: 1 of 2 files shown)


app/dev-app-update.yml
```diff
@@ -1,3 +1,3 @@
-owner: LiskHQ
+owner: reyraa
```
ManuGowda, line 1: should this be LiskHQ?

app/src/modules/autoUpdater.js
```diff
@@ -12,17 +13,17 @@ const getErrorMessage = (error) => {
 export default ({ // eslint-disable-line max-statements
   autoUpdater, dialog, win, electron,
 }) => {
+  autoUpdater.logger = logger;
```
ManuGowda, line 16: maybe remove the auto updater changes from this PR?

app/dev-app-update.yml
reyraa: For sure. But shouldn't we resolve it in your PR?

ManuGowda: yes, i think we should just keep only BTC changes in this PR

app/src/modules/autoUpdater.js
reyraa: Removed. It was a problem with the branch base.

ManuGowda (commented):
@reyraa the change itself LGTM, although I must say I do not have the full context of codebase, but the changes itself looks good. 

I see you have added in PR `unit tested`, I do not see any unit test, could you please add the tests?

isalga (approved):

ManuGowda (approved):

PR #3838, opened 2021-10-07
Lisk desktop tests fail when lisk-client is update beyond 5.0.1 - Closes #3804

### LINKED ISSUE #3804: Lisk desktop tests fail when lisk-client is update beyond 5.0.1
### Expected behavior

Configure Lisk desktop tests environment to handle `buffer` for both browser and nodejs

### Actual behavior

- When we update `lisk-client` dependency to its latest version the tests starts to fail
- The root cause of failing tests are due to the changes in the way lisk-client is being shipped from version > 5.0.1
  - The global buffer is an instance of browser buffer https://github.com/LiskHQ/lisk-sdk/blob/3f7ef0fdad3f91c01defe91546ba3bb212c6dc8f/elements/lisk-client/src/index.ts#L26-L28
  - Since the jest setup is configured to work with nodejs buffer, if we update lisk-client to its latest version the lisk-client functions starts to fail
  - Also their are some tests related to btc which require nodejs buffer, in this regards we need to check how we can get desktop tests to work for both browser and nodejs buffer instances
  
### Steps to reproduce

- Checkout development branch
- Update `lisk-client` to latest version `5.1.3`
- Run `npm install`
- Run `npm run test`

2.0.x

ManuGowda:
### What was the problem?

This PR resolves #3804

### How was it solved?

With the latest version of lisk-client the buffer issue has been resolved by exporting nodejs verions by default
Update webpack to to ignore os fallback to false since we do not use lisk-api-client

### How was it tested?

- npm ci
- npm run test
- npm run dev

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -18,6 +18,7 @@ const config = {
     fallback: {
       net: false,
       fs: false,
+      os: false,
       crypto: require.resolve('crypto-browserify'),
       stream: require.resolve('stream-browserify'),
       path: require.resolve('path-browserify'),
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@ledgerhq/hw-transport-node-hid": "5.50.0",
         "@ledgerhq/hw-transport-u2f": "5.33.1-alpha.3",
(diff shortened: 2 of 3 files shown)


ManuGowda:
I will merge this PR once the lisk-client dependency is published.

PR #3870, opened 2021-10-19
Warn users when voting a punished delegate - Closes #3768

### LINKED ISSUE #3768: Warn users if they are trying to cast a vote for a currently punished delegate
If a user adds a vote to a delegate that is currently punished there should be a very clear and loud warning that the user's LSK will be locked for X days and that the vote will be counted for the delegate's vote weight AFTER this time and UNTIL then, it has no effect on the delegate's vote weight.

Regarding "X days": `put here the remaining calculated punishment time of the delegate`

### Acceptance criteria

- Meet the proposed [design]( https://www.figma.com/file/FvOfHGyYPUclnv6FNUdumV/Improvements-(September-2021)?node-id=41%3A5960)

gadjacobs:
### What was the problem?

This PR resolves #3768 

### How was it solved?

- Disabled voting option for permanently banned delegates
- Add notification to warn users on punished delegates pages 
- Fetch delegate punishment details from their accounts and display on the warnings

### How was it tested?

Visually


src/components/screens/wallet/overview/index.js
```diff
@@ -53,24 +111,32 @@ const Overview = ({
...
+      <div
+        className={`${grid['col-xs-12']} ${grid['col-md-6']} ${grid['col-lg-6']} ${styles.balanceChart}`}
+      >
         {address && (
           <BalanceChart
             t={t}
             transactions={isWalletRoute ? confirmed : transactions.data.data}
             token={activeToken}
-            isDiscreetMode={discreetMode && host === address}
+            isDiscreteMode={discreteMode && host === address}
```
reyraa, line 139: Discreet means cautious, while discrete means divided. Please  revert this change everywhere.

reyraa (approved):

PR #3905, opened 2021-10-29
Fix login on enter key press - Closes #3877

### LINKED ISSUE #3877: Pressing enter after pasting a passphrase should submit the sign in form
### Expected behavior
Pressing enter after pasting a passphrase should submit the sign in form.

### Actual behavior
Pressing enter after pasting a passphrase currently opens the network switcher modal.

### Steps to reproduce
- Enable network switcher in settings
- Click on Sign In
- Paste a passphrase
- Press enter

Lisk Desktop (version 2.0.2)
MacBook Pro (13-inch, M1, 2020)
MacOs Big Sur (version 11.6)

ikem-legend:
### What was the problem?

This PR resolves #3877

### How was it solved?

- Refactoring passphrase input component to login when enter key is pressed

### How was it tested?

Visually
The app should also be built to ensure it works fine in Electron

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -36,7 +36,8 @@
           ["@shared", "./src/components/shared"],
           ["@screens", "./src/components/screens/"],
           ["@toolbox", "./src/components/toolbox/"],
-          ["@src", "./src"]
+          ["@src", "./src"],
+          ["@fixtures", "./test/constants/"]
         ],
         "extensions": [".js"]
       }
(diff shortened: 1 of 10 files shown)


reyraa (approved):
Some changes will cause conflicts with my PR which I'll resolve there. looks good then.

ikem-legend:
> The fix LGTM but the code cleanup related to network and test imports are unrelated to this issue. I would prefer that those changes goes to other PR and explained there.

@reyraa suggested that I implement the jsconfig.json file in my PR since his was going to take a bit longer. The file then triggered the discovery of the issues in DeepScan which I then fixed in the additional files. So while the additional files are unrelated to the PR, the jsconfig file is important to all of us working on the repo to easily find where components and functions are declared using an IDE, despite the use of aliases in imports

isalga (approved):

PR #3909, opened 2021-10-31
Fix standby status name - Resolves #3908

### LINKED ISSUE #3908: Misalignment in standby delegate status definition
### Expected behavior
Delegates are marked as `standby` by Lisk Service, given they have a minimum delegate weight of 1000 LSK and are none of the following:
- Active
- Punished
- Banned
Lisk Desktop must follow the same logic to mark delegates whose status is not defined by Lisk Service. Also it should mark them exactly `standby`.

### Actual behavior
Lisk Desktop marks such delegates `standBy`. This causes some visual issues.

### Steps to reproduce
Refer to the [PerformanceView](https://github.com/LiskHQ/lisk-desktop/blob/development/src/components/screens/wallet/delegateProfile/performanceView.js#L19) component.

2.1.0

reyraa:
### What was the problem?

This PR resolves #3908

### How was it solved?
Changed `standBy` status to `standby`.
We are patching the API response to rename the delegate status from non-eligible to ineligible.  I've created[ a ticket](https://github.com/LiskHQ/lisk-desktop/issues/3910) to revert this.

### How was it tested?
Visually.


src/components/screens/wallet/delegateProfile/performanceView.js
```diff
@@ -18,7 +18,7 @@ export const getStatus = (data) => {
     return data.status;
   }
   if (data.voteWeight >= DEFAULT_STANDBY_THRESHOLD) {
-    return 'standBy';
+    return 'standby';
   }
   return 'inEligible';
```
ManuGowda, line 23: does this also apply to `inEligible`? would be nice to case them similar

reyraa: I can see that in Lisk Service we use `non-eligible`, although according to our last meeting we decided `ineligible` sounds more familiar to native ears. I think we should fix it there. but as for camelCase-lowercase, I can change it to `ineligible`.

ManuGowda (approved):

PR #3920, opened 2021-11-05
Make blocks and transactions visible on scroll - Closes #3807

### LINKED ISSUE #3807: New blocks and transactions buttons should be visible even if the user scrolls the page
### Description

- Make sure the new blocks and new transaction buttons are static 
- Make sure the buttons always visible even if the user scrolls the page

image

### Motivation

In order for users to be able to refresh when their is new blocks or transaction at the same time to be able to scroll and see the transactions and blocks.

### Acceptance criteria

- The new block and new transaction button should be visible after navigating
- On click on new block and new transaction button the user should be navigated to top of the page
- Test on Mac, Win, Linux

ikem-legend:
### What was the problem?

This PR resolves #3807

### How was it solved?

Implementing sticky behaviour for headers and button as well as scrolling to top of page on button click

### How was it tested?

Visually


reyraa (approved):

ann-vashchenko:
Screenshot 2021-11-11 at 14 32 38

Looks good, but the button should be higher and shouldn't cover the column title

PR #3935, opened 2021-11-16
Remove BTC support - Closes #3928

### LINKED ISSUE #3928: Remove BTC support
### Description
To regain our focus and emphasis on developing the Lisk ecosystem, we are discontinuing the support for Bitcoin wallets. This will also reduce overall application complexity and maintenance.

### Acceptance Criteria
- The token selector at the top bar must be removed.
- The BTC switch in the settings page must be removed.
- The application should automatically activate LSK and disable BTC at the start time.

ikem-legend:
### What was the problem?

This PR resolves #3928

### How was it solved?

- Removing the token selector at the top bar
- Removing the BTC switch in the settings page
- Updating the middleware to activate LSK and disable BTC at start time

### How was it tested?

Manually

### DIFF
--- a/src/store/actions/service.js
+++ b/src/store/actions/service.js
@@ -13,8 +13,8 @@ const tickerReducer = (acc, key) => ({
 
 // eslint-disable-next-line import/prefer-default-export
 export const pricesRetrieved = () => (dispatch, getState) => {
-  const { settings: { token }, network } = getState();
-  const activeToken = token.active;
+  const { network } = getState();
+  const activeToken = 'LSK';
 
   return getPrices({ network })
     .then(({ data }) => {
--- a/src/store/actions/service.js
+++ b/src/store/actions/service.js
@@ -1,6 +1,6 @@
 import i18next from 'i18next';
 import { toast } from 'react-toastify';
-import { actionTypes } from '@constants';
+import { actionTypes, tokenMap } from '@constants';
 import { getPrices } from '@api/market';
 
 const tickerReducer = (acc, key) => ({
@@ -14,7 +14,7 @@ const tickerReducer = (acc, key) => ({
 // eslint-disable-next-line import/prefer-default-export
 export const pricesRetrieved = () => (dispatch, getState) => {
   const { network } = getState();
-  const activeToken = 'LSK';
+  const activeToken = tokenMap.LSK.key;
 
   return getPrices({ network })
     .then(({ data }) => {
--- a/src/store/actions/settings.js
+++ b/src/store/actions/settings.js
@@ -10,7 +10,16 @@ export const settingsRetrieved = () => (dispatch) => {
   getFromStorage('settings', initialState, (data) => {
     dispatch({
       type: actionTypes.settingsRetrieved,
-      data,
+      data: {
+        ...data,
+        token: {
+          active: 'LSK',
+          list: {
+            BTC: false,
+            LSK: true,
+          },
+        },
+      },
     });
   });
 };
--- a/src/store/actions/settings.js
+++ b/src/store/actions/settings.js
@@ -1,4 +1,4 @@
-import { actionTypes } from '@constants';
+import { actionTypes, tokenMap } from '@constants';
 import { getFromStorage } from '@utils/localJSONStorage';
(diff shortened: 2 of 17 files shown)


src/store/reducers/settings.js
```diff
@@ -33,7 +36,7 @@ export const initialState = {
   discreetMode: false,
   token: {
     active: tokenKeys[0],
-    list: tokenKeys.reduce((acc, key) => { acc[key] = true; return acc; }, {}),
+    list: tokenList,
```
isalga, line 39: As we are only going to support LSK can we hardcode all of this to simply:
```
{
  active: tokenKeys.LSK.key,
  list: {
    tokenKeys.LSK.key: true,
    tokenKeys.BTC.key: false,
}
```
Reducing tokenKeys to generate the list looks fine in the case we have more tokens to support, but given that we only support LSK seems overengineering to me.

src/store/middlewares/settings.js
```diff
@@ -19,7 +19,16 @@ const settings = store => next => (action) => {
...
+      setInStorage('settings', {
+        ...store.getState().settings,
+        token: {
+          active: 'LSK',
+          list: {
+            BTC: false,
+            LSK: true,
+          },
+        },
+      });
```
isalga, line 31: At this place token settings should already equal to this. If there is some case were it is not in my opinion would be better to set the correct store during startup. At `settingsRetrieved` in middleware if `BTC=true` I would override the localstorage and get sure reducer does not receive `BTC=true`.

isalga (changes requested):

src/store/reducers/settings.js
ikem-legend: @isalga Well I left that possibility in the event we want to support another coin in the near future but what you suggested makes sense and I'll update accordingly

src/store/middlewares/settings.js
ikem-legend: Fixed. However, I implemented it in actions since the settings are retrieved there and it might be better to override it there rather than dispatching the same action in the middleware

reyraa (approved):

src/store/actions/service.js
```diff
@@ -13,8 +13,8 @@ const tickerReducer = (acc, key) => ({
 
 // eslint-disable-next-line import/prefer-default-export
 export const pricesRetrieved = () => (dispatch, getState) => {
-  const { settings: { token }, network } = getState();
-  const activeToken = token.active;
+  const { network } = getState();
+  const activeToken = 'LSK';
```
isalga, line 17: Can you use `tokenMap.LSK.key`?

src/store/actions/settings.js
```diff
@@ -10,7 +10,16 @@ export const settingsRetrieved = () => (dispatch) => {
   getFromStorage('settings', initialState, (data) => {
     dispatch({
       type: actionTypes.settingsRetrieved,
-      data,
+      data: {
+        ...data,
+        token: {
+          active: 'LSK',
```
isalga, line 16: Same. I would prefer all the 'LSK' strings replaced by `tokenMap.LSK.key`

test/unit-test-utils/fakeStore.js
```diff
@@ -31,7 +31,7 @@ const defaultStore = {
     token: {
       active: 'LSK',
       list: {
-        BTC: true,
+        BTC: false,
         LSK: true,
```
isalga, line 35: Same:
```
[tokenMap.LSK.key]: true,
[tokenMap.BTC.key]: false,
```
On this one and the previous.

isalga (changes requested):

src/store/actions/service.js
ikem-legend: Done

src/store/actions/settings.js
ikem-legend: Fixed

test/unit-test-utils/fakeStore.js
ikem-legend: Fixed

isalga (approved):

PR #3990, opened 2021-12-01
Display no standby delegates found message and inelegible delegates - Closes #3948

### LINKED ISSUE #3948: Display 'No standby delegates found' instead of 'Nothing found'
### Description

Given the case that there are no delegates with voteWeight over 1000 LSK to be eligible as standby, API returns 404 to the call for standby delegates (https://testnet-service.lisk.com/api/v2/accounts?isDelegate=true&limit=20&status=standby) resulting on table to display Nothing found.

### Motivation

As a user perspective this result seems like an error when it is an expected behaviour.

isalga:
### What was the problem?

This PR resolves #3948

### How was it solved?

- I needed to pass `status: 'standby,non-eligible'` instead of `status: 'standby',`. notice that Service uses `non-eligible` instead of `ineligible`. Therefore this is not a mistake.
- Also fixed the empty template of our generic table to display the custom error message correctly.

### How was it tested?
- Updated unit tests
- Checked the delegates page to make sure the standby and ineligible delegates are listed correctly.
- Check several pages to make sure the empty state work.


ikem-legend:
Looks good so far. However I found two issues:
- Empty message for delegates outside round displays twice (I used the filter to trigger empty state for such delegates)


- Active delegates no longer display

reyraa:
@ikem-legend Seems like the `emptyState` property wasn't used correctly. I fixed it.

ikem-legend (approved):

PR #3992, opened 2021-12-01
Fix bookmark after tx send - Closes #3991

### LINKED ISSUE #3991: Add to bookmark after sending transaction is broken
### Expected behavior
Label should be editable when receiver is not a delegate

### Actual behavior
Label is not editable when receiver is not a delegate

### Steps to reproduce
- Go to a non delegate account
- Send LSK there
- click Add to bookmark in transaction result dialog

macOs, release 2.2.0

isalga:
### What was the problem?

This PR resolves #3991 

### How was it solved?

Set isDelegate parameter correctly

### How was it tested?

Manually


ikem-legend (approved):
Works fine now

reyraa (approved):

PR #3994, opened 2021-12-02
Fix banned delegate profile errors - Closes #3977 

### LINKED ISSUE #3977: Banned delegate profile throws an error
### Expected behavior

It shouldn't cause any errors

### Actual behavior

Opening a delegate profile that was banned  throws an error
image

### Steps to reproduce

Open the wallet, sign in (without signing in it won't happen https://github.com/LiskHQ/lisk-desktop/issues/3976), go to delegate section, select any `banned` delegate from the `sanctioned` tab.

2.2.0-beta.0

reyraa:
### What was the problem?

This PR resolves #3977

### How was it solved?
- We had a missing null check
- React was complaining about a missing property that was removed form the component. I removed it from the propType.

### How was it tested?
Visually.

### DIFF
--- a/src/components/screens/wallet/overview/index.js
+++ b/src/components/screens/wallet/overview/index.js
@@ -65,7 +65,7 @@ const Overview = ({
     if (!isWalletRoute && host && address && (isBanned || pomHeights?.length)) {
       addWarningMessage({
         isBanned,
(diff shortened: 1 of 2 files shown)


ikem-legend (approved):
Looks good to me

PR #3995, opened 2021-12-03
Fix negative value in punished delegates warning and forged blocks broken link - Closes #3974

### LINKED ISSUE #3974: Incorrect information in delegates profile page
### Expected behavior
- We should display only positive values or zero value in the caution panel. If punishment ends, the caution panel will be removed
-  No hyperlink for Last forged block, if there is nothing found

### Actual behavior
- Caution panel is displaying even the punishments ends 
- Last forged block information is clickable

### Steps to reproduce
- Go to delegates monitoring page --> Sanctioned tab and select any delegates who is already punished

v2.2.0-beta.0

gadjacobs:
### What was the problem?

This PR resolves #3974 

### How was it solved?

- Show link only when there's been a forged block
- Check that the number of days left being punished is greater than zero before displaying a warning

### How was it tested?

Visually.


ikem-legend (approved):
Great job!

src/components/shared/warnPunishedDelegate/index.js
```diff
@@ -49,15 +49,17 @@ const Warning = ({ vote, ...props }) => {
 };
 
 export const DelegateProfileWarning = ({ daysLeft, punishmentStartDate, ...props }) => (
-  <WarnPunishedDelegate
-    daysLeft={daysLeft}
-    punishmentStartDate={punishmentStartDate}
-    {...props}
-  />
+  daysLeft >= 1 ? (
```
reyraa, line 52: Why mount the component in the first place? If the punishment is already over, the warning component must not be mounted. I think this condition must be moved to the `editVote` and `wallet/overview` components.

reyraa (changes requested):

reyraa (approved):

PR #4000, opened 2021-12-07
Updated voting validation in the case of locked balance - Closes #3968

### LINKED ISSUE #3968: Vote amount validation is incorrect
The fundamentals of updating votes in 2.2.0-beta.0 are broken.

I have an testnet account with a total balance of ~3.5 million LSK. Around 2 million LSK are already used to vote myself. Around 1.5 million are left. I tried to change my self-vote by 10 LSK to 2000010 LSK but it doesn't work because Lisk Desktop doesn't consider that I only want to change by 10 LSK. It thinks I want to add 2000010 LSK, which is wrong of course. See here:

voting_bugged

ikem-legend:
### What was the problem?

This PR resolves #3968

### How was it solved?

Updated the validation to account for voting above balance if the user has previously voted for self in locked balance

### How was it tested?

Visually


reyraa (changes requested):
I tried this from the e2e stand point, and there were some issues:
 - I tried to edit a vote from 30 to 1060, given that I have 1037, I should be able to do that, but I wasn't. And it didn't even show an error message.
 - If I add 1030 to a new delegate, it works fine. but 1040 invalidates the form, which is expected, but it doesn't show an error message.

reyraa (approved):

PR #4024, opened 2021-12-16
Improve e2e test suite - Closes #3734

### LINKED ISSUE #3734: Improve e2e test suite
### Description

The below points should be addressed to fix the unstable and failing e2e test

- We have duplicated step definitions and many step definitions are defined for a specific feature while they can be in common feature to prevent duplication
- We have step definitions that perform several assertions, increasing the chance of code duplication.
- Accounts and transactions are used in common, thus changing one feature can affect the features that run next.
- All tests run against a single machine. Parallelising can decrease the run times by about 50%
- Wallet filters and Send feature are unstable

### Acceptance Criteria

- We should move as many step definitions as possible to common steps.
- Each step should only assert one goal. We should refactor the complex step definitions down to several steps with a single assertion.
- We should create accounts specific to each feature so the result of feature don't affect the next features.
- We should determine if Cypress allows running features with parameter. If not, we should break cypress config into two and run e2e test only using Jekins parameter and against master
- We should investigate running e2e test on multiple machines in parallel.
- We check if defining feature specific accounts help, if not, we refactor these features and resolve fragile tests.

eniolam1000752:
### What was the problem?

This PR resolves #3734

### How was it solved?

I was solved by removing duplicated step definitions on most of the feature tests and adding new test cases to voting.feature

### How was it tested?

it was tested by running `npm run cypress:open` to run all test cases in an interactive UI


test/redump-blockchain.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/zsh
+
+LISK_CORE_PATH="../../lisk-service/"
+NODE_VERSION="v12.22.7"
```
fchavant, line 4: Why hardcode this instead of reading from the `.nvmrc` file?

test/redump-blockchain.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/zsh
+
+LISK_CORE_PATH="../../lisk-service/"
```
fchavant, line 3: It seems to me this variable should be called `LISK_SERVICE_PATH`.

test/redump-blockchain.sh
```diff
@@ -0,0 +1,36 @@
...
+NODE_VERSION="v12.22.7"
+
+cd "$LISK_CORE_PATH/jenkins/redis"
+
+source ~/.bash_profile
```
fchavant, line 8: The shebang points to zsh but we read the *bash* profile?

test/redump-blockchain.sh
```diff
@@ -0,0 +1,36 @@
...
+pm2 stop ../../ecosystem.core3.config.js
+
+make down
+cd ../mariadb
+make down
+
+LISK_CORE_PID=$(pgrep -u $USER lisk-core);
```
fchavant, line 15: Shouldn't pm2 be used? (Won't pm2 restart core if it's killed?)

test/redump-blockchain.sh
```diff
@@ -0,0 +1,36 @@
...
+
+lisk-core blockchain:import snapshots/unlocked-tokens/blockchain.db.tar.gz --force
+lisk-core forger-info:import snapshots/forger.db.tar.gz --force
+
+cd ../mariadb
+make up
+cd ../redis
+make up
+
+lisk-core start --network=devnet --api-ws --api-ws-host=0.0.0.0 --api-ws-port=8888 --enable-forger-plugin --port=5050
```
fchavant, line 36: Is there a reason for `--port=5050`?

eniolam1000752: port 5000 was already used by an internal service on mac so I had to use 5050

test/redump-blockchain.sh
eniolam1000752: nope I don't think pm2 kills lisk-core if it is stopped or killed @fchavant

ManuGowda:
The unstable tests will be resolved in https://github.com/LiskHQ/lisk-desktop/issues/4058

fchavant (commented):
I would still recommend dropping the `test/*.sh` scripts as they have some issues and probably only work on your local machine as-is.

reyraa (approved):

PR #4027, opened 2021-12-26
Delegate search points to transaction tab - Closes #4023

### LINKED ISSUE #1234: Add send all amount button to send transaction form
### Expected behaviour

 - The amount field should display a link/button "Set max. amount" on the right when focused and empty
- Once the amount field is not empty, the button is replaced with the USD/EUR conversion rate
- If the amount field is not focused and empty, then nothing is displayed on the right of it.
- Clicking the "Set max. amount" button fills in account balance minus 0.1 LSK fee.

screen shot 2018-10-19 at 10 46 35

### LINKED ISSUE #4023: Delegate search points to transaction tab
### Expected behavior
When you search for a delegate from the search box, you should be redirected to the delegates profile tab below:
Screenshot 2021-12-15 at 15 06 58

### Actual behavior
Searching for a delegate from the search box redirects to the transactions tab of the delegate.
Screenshot 2021-12-15 at 15 06 43

### Steps to reproduce
- On the lisk desktop app, click on the search icon and type in a delegates username.
- Select the user and you will be redirected to their transactions tab.

v2.2.0-Beta.1

eniolam1000752:
### What was the problem?

This PR resolves #4023

### How was it solved?

This issue was solved by adding the `tab` query param to the accounts route when a delegate search item was clicked

### How was it tested?

- Search for a delegate
- Click on any of the delegates that appear in the suggestion list
- you should be routed to the delegate profile page not delegate transactions


gadjacobs:
@eniolam1000752 you may need to sign the CLA first before the branch can be properly built

gadjacobs (approved):

reyraa (approved):
Good job @eniolam1000752

PR #4028, opened 2021-12-29
Fix missing latest vote properties - Closes #4018

### LINKED ISSUE #4018: The latest vote weight is missing
### Expected behavior
I should be able to see the vote weight of my latest vote and edit it.

### Actual behavior

I can't see my vote weight (the second row) and 

Screenshot 2021-12-14 at 15 49 25

release/2.2.0

gadjacobs:
### What was the problem?

This PR resolves #4018 

### How was it solved?

- Aligning the table header and row
- Triggering a re-render when the vote list is updated.

### How was it tested?

Visually.

### DIFF
--- a/src/components/screens/wallet/votes/votes.js
+++ b/src/components/screens/wallet/votes/votes.js
@@ -37,11 +37,11 @@ const Votes = ({
 
   // Fetch delegate profiles to define rank, productivity and delegate weight
   useEffect(() => {
-    if (isEmpty(accounts.data) && votes.data.length) {
-      const addressList = votes.data.map(vote => vote.address);
+    const addressList = Object.keys(hostVotes);
+    if (isEmpty(accounts.data) && addressList.length) {
       accounts.loadData({ addressList, isDelegate: true });
     }
-  }, [votes.data]);
+  }, [hostVotes]);
 
   const areLoading = accounts.isLoading || votes.isLoading;
   const filteredVotes = votes.data.filter((vote) => {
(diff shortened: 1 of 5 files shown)


src/components/screens/wallet/votes/votes.js
```diff
@@ -37,11 +37,11 @@ const Votes = ({
 
   // Fetch delegate profiles to define rank, productivity and delegate weight
   useEffect(() => {
-    if (isEmpty(accounts.data) && votes.data.length) {
-      const addressList = votes.data.map(vote => vote.address);
+    const addressList = Object.keys(hostVotes);
```
ManuGowda, line 40:
```suggestion
    const addressList = Object.keys(sentVotes);
```

src/components/screens/wallet/votes/votes.js
```diff
@@ -37,11 +37,11 @@ const Votes = ({
...
+    if (isEmpty(accounts.data) && addressList.length) {
       accounts.loadData({ addressList, isDelegate: true });
     }
-  }, [votes.data]);
+  }, [hostVotes]);
```
ManuGowda, line 44:
```suggestion
  }, [sentVotes]);
```

ManuGowda (commented):
Please add tests to cover the scenarios

src/components/screens/wallet/votes/votes.js
gadjacobs: @ManuGowda `hostVotes` returns both sent and received votes for the selected delegate, so it might be a more appropriate name.

reyraa: @gadjacobs This prop is defined as `hostVotes: state.voting` which only includes the votes sent by the account. not the ones they received.

reyraa (changes requested):

reyraa (approved):

PR #4031, opened 2021-12-31
Prevent login without passphrase - Closes #4025

### LINKED ISSUE #4025: Hitting enter on the sign in screen without a passphrase logs into an account
### Expected behavior
Hitting the enter key with empty, incomplete or incorrect passphrase should throw a validation error 
Screenshot 2021-12-17 at 14 42 15

### Actual behavior
Hitting enter with an empty passphrase logs into the account below:
Screenshot 2021-12-17 at 14 53 34

### Steps to reproduce
Press the enter key without typing a passphrase on the sign in screen.

v2.2.0

gadjacobs:
### What was the problem?

This PR resolves #4025 

### How was it solved?

Ensuring passphrase is not empty before triggering log in.

### How was it tested?

Visually.

### DIFF
--- a/i18n/locales/en/common.json
+++ b/i18n/locales/en/common.json
@@ -282,6 +282,7 @@
   "Min": "Min",
   "Minimize": "Minimize",
   "Missed slot": "Missed slot",
+  "Modify recovery phrase derivation path": "Modify recovery phrase derivation path",
   "More filters": "More filters",
   "Multisignature account details": "Multisignature account details",
   "Multisignatures": "Multisignatures",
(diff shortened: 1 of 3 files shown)


ManuGowda (approved):

eniolam1000752 (changes requested):
Was think it would be better if by hitting on the `enter` when an empty passphrase is inputted, a message just like this one in the snapshot below shows, telling the user that the field is required. Instead, there is no feedback from the action telling the user what is happening.


image

eniolam1000752 (approved):
Looks good to me

PR #4049, opened 2022-01-07
Fix account index discovery - Closes #4019 

### LINKED ISSUE #4019: Can't sign transactions using a new account on hardware wallet
### Expected behavior
I should be able to sign and broadcast delegate registration using a HW device.

### Actual behavior
When I sign a tx using an old account that has outgoing and incoming txs, the signature is created successfully and I can broadcast the tx to the network. but if I try the same thing with a new account with sufficient balance but no outgoing tx, the tx gets aborted immediately.

First I thought we have an issue with the last account, so I created two new accounts with tough balance (and no outgoing tx) but they both fail to sign a tx.

### Steps to reproduce
- Create a new account.
- Send token to this account.
- Try sending tokens (or create any other tx type)
- The tx will fail immediately and the error is unhandled:

> signUsingHW 2 Error: Error: hwCommand failed: Error: TransportStatusError: Ledger device: UNKNOWN_ERROR (0x6802)
    at _callee2$ (hwManager.js:64)
    at tryCatch (runtime.js:63)
    at Generator.invoke [as _invoke] (runtime.js:294)
    at Generator.throw (runtime.js:119)
    at asyncGeneratorStep (bundle.app.a63c6201e26f0f934ca2.js:64446)
    at _throw (bundle.app.a63c6201e26f0f934ca2.js:64448)

release/2.2.0

reyraa:
### What was the problem?

This PR resolves #4019

### How was it solved?

HW refers to accounts using their index. When we refactored the sign-in page and created new tabs, the index of the accounts in second and third tabs got wrong.
I passed the actual index along side the account info to use for account selection.

### How was it tested?
Unit tests

### DIFF
--- a/src/components/screens/hwWalletLogin/selectAccount/selectAccount.js
+++ b/src/components/screens/hwWalletLogin/selectAccount/selectAccount.js
@@ -13,12 +13,12 @@ const Tab = ({
   tabName, tabId, accountsList,
   onSaveNameAccounts, onSelectAccount,
 }) => (
-  <div tabName={tabName} tabId={tabId} className={`${styles.deviceContainer} ${`tab-${tabId}`} hw-container`}>
-    {accountsList.map((account, index) => (
+  <div name={tabName} id={tabId} className={`${styles.deviceContainer} ${`tab-${tabId}`} hw-container`}>
+    {accountsList.map((data) => (
       <AccountCard
-        key={`hw-account-tabId-${index}`}
-        account={account}
-        index={index}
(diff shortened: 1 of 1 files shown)


ManuGowda (commented):
@reyraa how was it tested mentions unit tests, did you miss to push the unit tests?

ManuGowda (commented):
Passing the index code changes LGMT. Tested using hw, able to create and send LSK

ManuGowda (approved):

PR #4052, opened 2022-01-10
Fix invalid URL - Closes #3887

### LINKED ISSUE #3887: Trailing slash makes URL invalid
### Expected behavior

Desktop connects to Lisk Service with and without the trailing slash at the end of the URL.

### Actual behavior

Trailing slash makes the URL invalid in the custom service node dialog.

Compare: 

`https://mainnet-service-staging.lisk.com/` vs `https://mainnet-service-staging.lisk.com`

### Steps to reproduce

Screenshot 2021-10-25 at 12 24 01
Screenshot 2021-10-25 at 12 23 52

`2.0.2`

isalga:
### What was the problem?

This PR resolves #3887 

### How was it solved?

- Adding function to remove trailing slash

### How was it tested?

Manually


src/components/screens/login/networkSelector/customNode/editMode.js
```diff
@@ -22,6 +22,16 @@ const validateNode = async (address) => {
...
 
+const removeTrailingSlash = (input) => {
+  const bypassStrings = ['http:/', 'http://', 'https:/', 'https://'];
+
+  if (input.charAt(input.length - 1) !== '/' || bypassStrings.includes(input)) {
+    return input;
+  }
+
+  return input.substring(0, input.length - 1);
+};
```
reyraa, line 33: Why not use a regex?

```suggestion
/**
 * Removes the trailing slash
 *
 * @param {string} url - A URL that might end in a slash
 * @returns {string} A URL without a trailing slash
 */
const removeTrailingSlash = url => url.replace(/\/$/, '');
```

reyraa (approved):

PR #4071, opened 2022-01-13
 Prevent double broadcasting of the delegate registration transaction - Closes #4054

### LINKED ISSUE #4054: Prevent double broadcasting of the delegate registration transaction
### Description
In the delegate registration screen:
The tx status page displays the Send button since all signatures are collected, but when I press send, the screen remains on the previous/current state until the tx is approved. This is not a desired behavior since the user might press the Send button several times.

We had an animation here which is replaced by the multisignature illustration. Therefore we should change to result page to not wait until the tx is confirmed.

### Steps to reproduce
- Try registering a delegate using a mutlsig account.
- Click Send to broadcast the transactions.
- Look at the tx result screen.

2.0.0-beta.1

reyraa:
### What was the problem?

This PR resolves #4054

### How was it solved?

- Removed unnecessary animations
- Adapted the tx status component with the generic illustration. I did this since there's no reason to keep the user in the tx status screen till the tx is confirmed. we do not do this for any other tx, and we shouldn't do it for delegate reg either.
- Also fixed the error report button for mutlsig txs in addition to the ticket title.

### How was it tested?
Manually


eniolam1000752 (approved):

eniolam1000752:
LGTM

PR #4091, opened 2022-01-20
Gracefully handle exception been thrown by /votes_sent - Closes #4089

### LINKED ISSUE #4089: The application throws when the account does not have sent votes
### Expected behavior
Many accounts do not have sent votes. If the `/votes_sent` API returns 404, Lisk Desktop should handle this error gracefully.

### Actual behavior
The application throws an error that breaks the [login e2e](https://github.com/LiskHQ/lisk-desktop/blob/development/test/cypress/features/login.feature#L3,L8) tests.

### Steps to reproduce
- Use an account without sent votes.
- Check the dev console. you'll see the following error:

image

2.2.0

### More information
The mainnet e2e tests is disabled due to the above issue. Once the issue is resolved, the mainnet [login e2e](https://github.com/LiskHQ/lisk-desktop/blob/development/test/cypress/features/login.feature#L3,L8)tests should be reinstated.

eniolam1000752:
### What was the problem?

This PR resolves #4089

### How was it solved?

- [x] wrapped `await getVotes` call in a try-catch block

### How was it tested?

on Jenkins


src/store/actions/voting.js
```diff
@@ -117,10 +117,12 @@ export const votesRetrieved = () =>
...
-      data: votes.data,
-    });
+    try {
+      const votes = await getVotes({ network, params: { address } });
+      dispatch({
+        type: actionTypes.votesRetrieved,
+        data: votes.data,
+      });
+      return 0;
+    } catch (exp) { return 0; }
```
reyraa, line 127: The action should always dispatch. Although the action type and data might be different according to the API response, there should always be an action dispatched. This brings consistency to the behavior of the action function.

reyraa (approved):

PR #4107, opened 2022-01-25
Implemented e2e test for transactions monitor screen - Closes #4066

### LINKED ISSUE #4066: Add e2e tests for transactions monitor screen
### Description
As a part of  #4065 we'd like to introduce e2e test coverage for the transactions monitor screen. The e2e tests should test the following user journeys. 

### Acceptance criteria
- Load latest transactions.
- Filter the transactions based on the given filtering criteria.
- Navigate to specific transaction details page
   - The transaction id in the address must be correct.
   - The transaction details should show expected transaction fields.
- If Lisk Desktop throws an unwanted error that fails the e2e test in a random basis, it should be resolved.

ikem-legend:
### What was the problem?

This PR resolves #4066

### How was it solved?

Writing the end to end tests for the page

### How was it tested?

Visually


test/cypress/features/transaction.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Transactions 
+
+  Background:
+    Given I login as testnet_guy on testnet
```
eniolam1000752, line 4: I think this can be tested against dev net

test/cypress/features/transaction/transaction.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable */
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps';
+import { ss } from '../../../constants';
+
+Given(/^Given I am on Transactions page$/, function () {
+	cy.visit('/transactions');
+})
```
eniolam1000752, line 7: there already exists a step implementation that takes care of page/screen visitation in `common.js` it would be better to either extend or use that step definition

image

test/cypress/features/transaction/transaction.js
```diff
@@ -0,0 +1,11 @@
...
+
+Then(/^I should see more than ([^\s]+) transactions in table$/, function (trnxNumber) {
+  cy.get(ss.transactionRow).should('have.length.greaterThan', Number(trnxNumber));
+});
```
eniolam1000752, line 11: there already exists a step definition for this too in `common.js` also since this test can be run against devnet instead of using greaterThan, a defined length can be used since the length of the transaction to be displayed is always known.

test/cypress/features/transaction.feature
```diff
@@ -0,0 +1,39 @@
...
+    And I wait 5 seconds
+    Given I am on Transactions page
+
+  Scenario: Load Latest Transactions
+    Then I should see 20 transactions in table
+    When I click on showMoreButton
+    Then I should see more than 20 transactions in table
```
eniolam1000752, line 11: if devnet is used, clicking more would render a definite value of 40 transactions

test/cypress/features/transaction.feature
```diff
@@ -0,0 +1,39 @@
...
+    Then I should see 20 transactions in table
+    When I click on filterTransactionsBtn
+    Then I should see filterDropdown
+    When I fill 1000 in amountToInputFilter field
+    And I click on applyFilters
+    Then I should see 20 transactions in table
+    When I click on filterTransactionsBtn
+    Then I should see filterDropdown
+    When I fill lskgpfj26e3uogvr5wbono75hzn5myds4pncv3zm8 in senderAddressFilter field
+    Then I should see more than 5 transactions in table
```
eniolam1000752, line 27: I think a deterministic number of transactions can also be known here if the test is ran against devnet

test/cypress/features/transaction.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Transactions 
+
+  Background:
+    Given I login as testnet_guy on testnet
+    And I wait 5 seconds
+    Given I am on Transactions page
```
eniolam1000752, line 6: If the step definition in `common.js` is used then this feature step can be done as

`Given I am on transactions page`

src/components/shared/transactionsTable/index.js
```diff
@@ -73,7 +73,7 @@ const TransactionsTable = ({
         clearFilter, clearAllFilters, filters, formatters, t,
       }}
       />
-      <BoxContent className={styles.content}>
+      <BoxContent className={`${styles.content} transaction-results`}>
```
eniolam1000752, line 76: `transaction-results` was added as a selector class for cypress though it wasn't used any were in the PR

eniolam1000752 (changes requested):

eniolam1000752:
I think with `cypress.intercept`, it is also possible to check the response from local service(i.e if the test is running against devnet) in order to test that the data rendered in the Dognut Chart or grouped bar chart is correct.

image

src/components/shared/transactionsTable/transactionRow.js
```diff
@@ -18,7 +18,7 @@ const TransactionRow = ({
   data, className, t, currentBlockHeight,
 }) => (
   <DialogLink
-    className={`${grid.row} ${className}`}
+    className={`${grid.row} ${className} transactions-row`}
```
isalga, line 21: Same as `transactions-results`

test/cypress/features/transaction.feature
```diff
@@ -0,0 +1,39 @@
...
+    When I click on filterTransactionsBtn
+    Then I should see filterDropdown
+    When I fill 100 in amountFromInputFilter field
+    When I fill 1000 in amountToInputFilter field
+    When I click on moreLessSwitch
+    When I fill 15424425 in heightFilter field
+    And I click on applyFilters
+    Then I should see 1 transactions in table
+    And I click on clearAllFiltersBtn
+    Then I should see 20 transactions in table
```
isalga, line 39: Can you add additional filters like transaction types and height?

isalga (changes requested):

isalga (commented):
I don't see this case being cover:
`Once clicked on each item, it should show the correct tx details page:
- The tx id in the address must be correct.
- The info should be displayed correctly.`

That should come too in this PR

src/components/shared/transactionsTable/transactionRow.js
ikem-legend: @isalga I don't think so. `transaction-results` refers to the table container and as a result it can't be used directly to count the number of rows in the table, that's why I needed the `transactions-row`

test/cypress/features/transaction.feature
ikem-legend: @isalga I initially tried but couldn't get the right identifiers for the dropdown but I'll try again

test/cypress/features/transaction/transaction.js
ikem-legend: @eniolam1000752 While I was aware of using exact length, there are cases when such user might make further transactions in the future so using an exact length would work for the time being but break tests in the future. In such case, since the minimum number of transactions is known at the time of writing the test, then expecting such number or more can future proof the test especially in a case where the expected result is less than 20 for the time being

test/cypress/features/transaction.feature
ikem-legend: @eniolam1000752 I've explained my thought process behind this in the step definition. I'm not quite certain that running against devnet might affect the process too much but I'll look into it

src/components/shared/transactionsTable/index.js
ikem-legend: @eniolam1000752 This was done for uniformity to ensure that such selector can be used in any future test where a specific value is to be checked for in the table. For example, `cy.get(ss.transactionsTable).contains('lskgpfj26e3uogvr5wbono75hzn5myds4pncv3zm8').should('exist')`. However I can remove it for the time being

test/cypress/features/transaction.feature
ikem-legend: @eniolam1000752 Fixed

test/cypress/features/transaction.feature
ikem-legend: @isalga Done

test/cypress/features/transaction.feature
ikem-legend: Done

test/cypress/features/transaction/transaction.js
ikem-legend: Fixed

test/cypress/features/transaction.feature
ikem-legend: Fixed

test/cypress/features/transaction/transaction.js
```diff
@@ -0,0 +1,15 @@
...
+  cy.get(ss[field]).click()
+  cy.get('.options').contains(value).click();
+});
+
+Then(/^I should be on transaction details page$/, function () {
+  cy.location().should((location) => {
+    const hasAddress = /\?transactionId=a1c5521f466ae5476d3908cc8d562444d45adf4ac3af57e77f1f9359999ab9ca/.test(location.href);
+    expect(hasAddress).true;
+  });
+});
```
eniolam1000752, line 15: I think the step definition `I should be on transaction details page` isn't an appropriate name for this scenario since its a modal that gets opened when a transaction is clicked.

eniolam1000752 (changes requested):

test/cypress/features/common/common.js
```diff
@@ -80,6 +80,9 @@ Given(/^I am on (.*?) page$/, function (page) {
     case 'delegates':
       cy.visit(urls.delegates);
       break;
+    case 'transactions':
+      cy.visit(urls.transactions);
+      break;
```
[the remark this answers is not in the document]
ikem-legend: @eniolam1000752 Fixed

test/cypress/features/transaction/transaction.js
ikem-legend: @eniolam1000752 Fixed

test/cypress/features/common/common.js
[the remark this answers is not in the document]
ikem-legend: @eniolam1000752 Fixed

eniolam1000752 (approved):

isalga (approved):

ikem-legend:
Closed due to errors in version history

PR #4121, opened 2022-01-31
Fix account retrieval logic in the votes tab of the wallet page - Closes #4115

### LINKED ISSUE #4115: Can't load votes for other delegates
### Expected behavior
You should be able to fetch both sent and received votes for every delegate.

### Actual behavior
It returns an empty list every time, even if votes have clearly been sent and/or received.
Screenshot 2022-01-26 at 11 41 07

### Steps to reproduce
- open the Lisk Desktop app
- Navigate to any delegates profile other than yourself and select the votings tab

Lisk Desktop 2.2.0

gadjacobs:
### What was the problem?

This PR resolves #4115 

### How was it solved?

- Fetch `sentVotes` from the correct endpoint.

### How was it tested?

Added unit tests

### DIFF
--- a/src/components/screens/wallet/votes/votes.js
+++ b/src/components/screens/wallet/votes/votes.js
@@ -17,7 +17,7 @@ const getMessages = t => ({
 });
 
 const Votes = ({
-  votes, accounts, address, t, history, sentVotes = {},
+  votes, accounts, address, t, history,
 }) => {
   const [filterValue, setFilterValue] = useState('');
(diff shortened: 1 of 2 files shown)


reyraa (approved):
There's a non-blocking issue. The component makes API calls to votes_sent and votes_received twice. This is not optimal. Would be cool if we resolved it. but given the 2.2.0 release plan, we can create an issue to fix this later.

ManuGowda (approved):

PR #4152, opened 2022-02-10
Fix multi signature transaction signing - Closes #4141

### LINKED ISSUE #4141: Unable to send a transaction from a specific multisignature account
### Expected behavior
User should be able to send a transaction from any multisignature account
### Actual behavior
User is unable to send a transfer transaction from a multisingature account that has one required key that doesn't belong to senders account (example of the account https://testnet-service.lisk.io/api/v2/accounts?address=lsko3qy92jx8g9h4zhghdbey6n753k434wmpeoeaw)
### Steps to reproduce
Create a multisignature account with one required key that doesn't belong to senders account (example https://testnet-service.lisk.io/api/v2/accounts?address=lsko3qy92jx8g9h4zhghdbey6n753k434wmpeoeaw). After creating it, try sending a transfer transaction with the account
Windows Lisk Desktop 2.2.0, doesn't affect 2.1.0.

ikem-legend:
### What was the problem?

This PR resolves #4141

### How was it solved?

Updated the multisignature transaction validation

### How was it tested?

- Create a one member multisig account with a different account as the only signatory
- Create transactions with the multisig account


ManuGowda (commented):
@ikem-legend The changes work well in non hardware wallet scenario.

@reyraa On hardware wallet it still fails to sign the transaction, Ikem doesn't have the HW, so if you can look into it would be nice.

isalga (dismissed):

reyraa (approved):

ManuGowda (approved):

PR #4154, opened 2022-02-10
Fix sorting and filtering exceptions on certain transaction types - Closes #4118

### LINKED ISSUE #4118: Filtering and sorting by amount on vote transactions returns  empty results
### Expected behavior
- The recipient filter input should be active only for the transfer type.
- The amount from/to range should be active only for the transfer, unlock and reclaim types.
- Once a tx types are selected
   - The recipient and amount rage inputs must be disabled according to the above criteria.
   - The previously selected recipient and amount range filters must be removed if irrelevant.
- You should make the above changes dynamically based on the existing module asset schemas.
- The above criteria should also work for sorting.

### Actual behavior
- Sorting vote transactions by amount doesn't render any vote transactions.
- Filtering vote transactions  by amount doesn't render any vote transactions.

### Steps to reproduce
- Navigate to the transactions page.
- Filter transactions by vote type and by a given minimum and maximum amount (no vote transaction is rendered even for a valid amount range).
- Clear minimum and maximum amount filter values (all list of votes should be rendered)
- Sort the list by amount (no vote transaction is rendered)

- v2.2.0 
- all OS

eniolam1000752:
### What was the problem?

This PR resolves #4118 

### How was it solved?

- [x] implement exceptions on sorting by amount for `Vote`, `Register Delegate`, `Report Delegate Misbehaviour` and `Register Multisignature group` transaction types.

- [x] Disable min/max amount fields and recipient address fields on filter when `Vote`, `Register Delegate`, `Report Delegate Misbehaviour` and `Register Multisignature group` transaction type is selected

- [x] Write unit test

### How was it tested?

- Visually


test/fixtures/transactions_data.js
```diff
@@ -0,0 +1,175 @@
+export default [
```
ikem-legend, line 1: @eniolam1000752 I believe filename should be renamed to camel case for consistency, preferably `transactions.js`

ikem-legend (commented):
Looks good generally, just one comment to be resolved. Great job!!

ikem-legend (approved):

ManuGowda (commented):
LGTM

ManuGowda (approved):

PR #4167, opened 2022-02-18
Show warning tooltip for old bookmark addresses - Closes #3793

### LINKED ISSUE #3793: Change bookmark accounts from old to new addresses
Currently old bookmarks still shows the old addresses. It should automatically switch to new address system. Otherwise people have to remove their bookmarks and add them new. Not a drama, but would be nice to have this done automatically.

gadjacobs:
### What was the problem?

This PR resolves #3793 

### How was it solved?

-Increasing the height of the container to fill the parent so scroll only starts when it's at 100% height.

### How was it tested?

Visually.

If you have no bookmarked legacy accounts to test with, you can toggle `!bookmark.disabled` on line 229 in `src/components/screens/bookmarks/list/list.js` to show the warning on the regular addresses.


ikem-legend (dismissed):
Looks good to me

eniolam1000752 (dismissed):
LGTM

ManuGowda:
@gadjacobs i think the changes actually introduced bug? please find the attachment with old and new changes

OLD
Screenshot 2022-02-18 at 8 07 47 PM


NEW
Screenshot 2022-02-18 at 8 06 36 PM

ManuGowda:
> @ManuGowda did you return  `!bookmark.disabled` on line 229 in `src/components/screens/bookmarks/list/list.js` back to normal? That's only to test what it would look like on a legacy addresses in case you don't have one already bookmarked.

Yes I did that, that's how I was able to capture the results 🙂... We can discuss Monday if you want

PR #4171, opened 2022-02-21
Disable send button for an incorrect second passphrase - Closes #4169

### LINKED ISSUE #4169: No error response when providing an incorrect second passphrase
### Expected behavior
Lisk desktop should show an error that the provided second passphrase is not valid
### Actual behavior
Lisk desktop doesn't show any error, instead shows a broken modal and makes the app unresponsive 
image

### Steps to reproduce
Login to a multisig account that is recognized as a "second passphrase account", make a transaction, use "Send using second passphrase right away" feature, type in an incorrect second passphrase that is bip39 compliant.
Windows 10, Lisk Desktop 2.2.0

eniolam1000752:
### What was the problem?

This PR resolves #4169 

### How was it solved?

- [x] Disabling send button when there is a validation error on the second passphrase

### How was it tested?

- Try initiating a transition on a multi-signature account by providing a second passphrase
- providing an invalid passphrase disables the send button

### DIFF
--- a/src/components/shared/transactionSummary/footer.js
+++ b/src/components/shared/transactionSummary/footer.js
@@ -39,7 +39,7 @@ const SecondPassInput = ({
 
   useEffect(() => {
     if (secondPass.error === 0) {
-      secondPassphraseStored(secondPass);
+      secondPassphraseStored(secondPass.data);
       setInputStatus('valid');
     } else if (secondPass.error > 0) {
       setInputStatus('invalid');
--- a/src/components/shared/transactionSummary/footer.js
+++ b/src/components/shared/transactionSummary/footer.js
@@ -24,7 +24,7 @@ const Actions = ({
     )}
     <PrimaryButton
       className="confirm-button"
-      disabled={confirmButton.disabled || inputStatus === 'visible'}
+      disabled={confirmButton.disabled || inputStatus === 'visible' || inputStatus === 'invalid'}
       onClick={confirmButton.onClick}
     >
       {isMultisignature ? t('Sign') : confirmButton.label}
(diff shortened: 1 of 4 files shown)


ManuGowda (approved):

src/components/shared/transactionSummary/footer.js
```diff
@@ -57,7 +61,10 @@ const SecondPassInput = ({
     <div className={styles.secondPassphrase}>
       <PassphraseInput
         t={t}
-        onFill={set2ndPass}
+        onFill={(passPhrase, validationError) => {
+          setIsValid(validationError);
+          set2ndPass(passPhrase);
+        }}
```
reyraa, line 67: Please replace the state hook with a custom hook that validates the second passphrase and stored the `secondPassphrase` as:

```
{
   data: '',
   error: 0,
   feedback: [],
}
```

Where error is `-1` for empty, `0` for valid, and a positive number for the number of errors discovered.
And feedback is an array of strings, each representing an error message.

reyraa (changes requested):

src/hooks/setSecondPassphrase.js
```diff
@@ -0,0 +1,46 @@
+import { useState } from 'react';
+import { useSelector } from 'react-redux';
+import { selectActiveTokenAccount } from '@store/selectors';
+import { extractPublicKey } from '@utils/account';
+
+const empty2ndPass = {
+  data: '',
+  error: -1,
```
ManuGowda, line 8: maybe either `status` or `errorStatus`?

ManuGowda (approved):

reyraa (approved):

PR #4173, opened 2022-02-22
Fix dynamic voting balance issue - Closes #4143

### LINKED ISSUE #4143: Available balance for voting should consider votes that are in queue already
I have 2000 LSK and I add a vote of 500 LSK to delegate A (vote in queue, not executed) and when I go to delegate B now and add a vote, the pop up shows that 2000 LSK are avaiable which is right but also wrong. It should show 1500 LSK available. This would be more user-friendly.

ikem-legend:
### What was the problem?

This PR resolves #4143

### How was it solved?

- Changed balance display for voting to be based on max vote amount
- Implemented info tooltip

### How was it tested?

Manually

### DIFF
--- a/src/components/screens/editVote/index.js
+++ b/src/components/screens/editVote/index.js
@@ -7,7 +7,6 @@ import { compose } from 'redux';
 import { selectSearchParamValue, removeSearchParamsFromUrl } from '@utils/searchParams';
 import { tokenMap } from '@constants';
 import { voteEdited } from '@actions';
-import { selectAccountBalance } from '@store/selectors';
 import { toRawLsk, fromRawLsk } from '@utils/lsk';
 import Dialog from '@toolbox/dialog/dialog';
 import Box from '@toolbox/box';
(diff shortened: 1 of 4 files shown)


src/components/screens/editVote/index.js
```diff
@@ -97,11 +95,11 @@ const AddVote = ({
             <p className={styles.balanceTitle}>{t('Available balance')}</p>
             <div className={styles.balanceDetails}>
               <span className={styles.lskValue}>
-                <LiskAmount val={balance} token={tokenMap.LSK.key} />
+                <LiskAmount val={maxAmount} token={tokenMap.LSK.key} />
```
ManuGowda, line 98: this logic will replace the account balance? e.g it will only display whats the maximum possible amount to send rather than balance available. The modal shows `Available balance` and we should display `Available balance`

ManuGowda (changes requested):
@ikem-legend the logic changes the behaviour please take a look at screen shoot
Screenshot 2022-02-23 at 10 00 16 AM

src/components/screens/editVote/index.js
ikem-legend: @ManuGowda That's one of the few ways to fix the issue as it was described but I'll align with @reyraa to find the best solution

ManuGowda (approved):

reyraa (approved):

PR #4175, opened 2022-02-23
Implement new HW select account view - Closes #4012 

### LINKED ISSUE #4012: HW can not create new account because there is always a non-initialised one
The current design of the HW login page seems to confuse some users. We should consider improving the user experience and resolve the following issues:

- Users must be able to create only one new account. A new account is not an account with balance = 0LSK, but an account that has no incoming or outgoing transaction.
- Users must notice that they have to take immediate action about their reclaimable balance.
- Users must be able to filter out the accounts with balance = 0LSK. This is not the new account, but the legacy accounts that could have empty balance. With Core v3, this is not possible since an account should have at least 0.5LSK after making a transaction.
- A new account must be easily distinguishable. 
- Users must have only one way to create a new account.

Mac, 2.2.0-beta.1

isalga:
### What was the problem?

This PR resolves #4012 

### How was it solved?

- Implement checkbox
- Update styles

### How was it tested?

Add unit test
run electron build locally, you will need a mock legacy account to test tooltip. You can do it like [this](https://github.com/LiskHQ/lisk-desktop/pull/4175/commits/d77eccbe15218b86e930fa3c227888fec2223d6a)


src/components/screens/hwWalletLogin/selectAccount/selectAccount.js
```diff
@@ -134,63 +117,53 @@ class SelectAccount extends React.Component {
...
     return (
-      <div>
+      <div className={styles.selectAccountWrapper}>
         <h1>{t('Lisk accounts on {{WalletModel}}', { WalletModel: device.model })}</h1>
         <p>
           {t('Please select the account you’d like to sign in to or')}
           <TertiaryButton
             className={`${styles.createAccountBtn} create-account`}
             onClick={this.onAddNewAccount}
+            disabled={hwAccounts.some(account => account.summary?.balance === 0)}
```
ManuGowda, line 130: In case of hardware wallets the last account balance will always be zero and the `create account` button will be disabled all the time, should we consider not showing it if its not functional? @isalga @reyraa

ManuGowda (changes requested):

src/components/screens/hwWalletLogin/selectAccount/selectAccount.js
isalga: Agree it should not be displayed.

ann-vashchenko (approved):

reyraa:
@ManuGowda  I made the following changes to the PR:

- The `getAccountsFromDevice` now fetches 10 public keys at a time, and makes only 1 API call per 10 public key. This will help notably decrease the time of displaying the account.
- Also, the `getAccountsFromDevice` function does not return a new account at the end of the list, therefore there's no need for imperative conditional statements for showing/hiding the new account. 
- By clicking on the create account button, the `getNewAccountByIndex` function returns a new account. If you don't use it, the button will be active the next time, and it will generate the exact same account it created before (because it creates by index).

ManuGowda (approved):
Nice job @isalga @reyraa the fetching of all accounts once improves the overall experience. Unfortunately I do not have the  HW to test, but the code changes itself looks good.

ann-vashchenko (approved):

PR #4184, opened 2022-02-25
Fixed rerender issue on clicking entire balance warning - Closes #4176

### LINKED ISSUE #4176: Clicking on send entire balance warning causes a rerender
### Expected behavior

When the warning is clicked, nothing should happen.

### Actual behavior

Clicking anywhere on the warning causes a re-render.

### Steps to reproduce

- Go to any user account and try to send some LSK
- Click on 'send entire balance' on the send modal
- Now click on the warning that appears below the input field

v2.3.0 beta.0

ikem-legend:
### What was the problem?

This PR resolves #4176

### How was it solved?

Changing the click setup for the warning

### How was it tested?

Manually

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - security
+  - discussion
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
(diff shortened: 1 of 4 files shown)


eniolam1000752 (commented):
I think a unit test for this scenario would be needed.

ikem-legend:
> I think a unit test for this scenario would be needed.

Fixed

eniolam1000752 (changes requested):
There a commits from other branches that I think are not suppose to be in this PR

ikem-legend:
> There a commits from other branches that I think are not suppose to be in this PR

Since the PR needed to be rebased from release, rather than development, some extra commits became part of it but I'll review them

ikem-legend:
Similar PR https://github.com/LiskHQ/lisk-desktop/pull/4187 was created with cleaner history

PR #4192, opened 2022-03-03
Update nvm to use node.js LTS - Closes #3765 

### LINKED ISSUE #3765: Update nvm to use node.js LTS
### Expected behavior

Update nvm to use LTS version for stability

_Note: We should always use even-numbered releases (10, 12, 14, 16 etc.) for production, as these releases tend to be Active LTS status and are ready for general use._

### Actual behavior

- We should use LTS version for nvm, in this case we are using `15.x.x` which is not a stable version.

### Steps to reproduce

Check `nvm` file for node.js version

master, development

reyraa:
### What was the problem?

This PR resolves #3765

### How was it solved?

Updated the node version in .nvmrc file.

### How was it tested?

on Jenkins


gadjacobs (dismissed):

ikem-legend (dismissed):

ikem-legend (approved):
LGTM

gadjacobs (dismissed):

gadjacobs (commented):
I don't think the commits are signed @reyraa

PR #4193, opened 2022-03-03
Implement e2e test for blocks monitor screen - Closes #4099

### LINKED ISSUE #4099: Add e2e tests for blocks monitor screen
### Description
As a part of #4065 we'd like to introduce e2e test coverage for the blocks monitor screen. The e2e tests should test the following user journeys.

### Acceptance criteria
- Loads blocks.
- Load more blocks.
- Retrieves and prepares the new blocks to be prepended.
- Sorting should work as expected. (by height)
- The filter feature works and you can filter the blocks based on all existing inputs.
- Once clicked on an item, it should navigate to the block details page and:
   - Display the details of the block correctly.
   - Display the list of tx of the block.
   - Tx items must be clickable. The page should open the tx details modal with correct tx id. There's no need to test the tx details info, since it's being tested by #4066
- If Lisk Desktop throws an unwanted error that fails the e2e test in a random basis, it should be resolved.

ikem-legend:
### What was the problem?

This PR resolves #4099

### How was it solved?

Writing the e2e tests for the different scenarios
- [x] Loads blocks
- [x] Load more blocks
- [x] Retrieves and prepares the new blocks to be prepended
- [x] Sorting should work as expected (by height)
- [x] The filter feature works and you can filter the blocks based on all existing inputs
- [x] Once clicked on an item, it should navigate to the block details page and:
      - Display the details of the block correctly
      - Display the list of tx of the block
      - Tx items must be clickable. The page should open the tx details modal with correct tx id

### How was it tested?

Manually


eniolam1000752:
I think you would need to fix the commit history because there are commit that I don't think are supposed to be in this PR

ikem-legend:
> I think you would need to fix the commit history because there are commit that I don't think are supposed to be in this PR

Fixing

src/components/screens/monitor/blockDetails/blockDetails.js
```diff
@@ -57,49 +57,49 @@ const getFields = (data, token, t, currentHeight) => ({
...
+    value: data?.version,
   },
   generator: {
     label: t('Generated by'),
     value: (
       <Generator
-        generatorAddress={data.generatorAddress}
-        generatorUsername={data.generatorUsername}
+        generatorAddress={data?.generatorAddress}
+        generatorUsername={data?.generatorUsername}
```
eniolam1000752, line 84: I think instead of having optional chaining on `data` in many places within the getFields function block, I think on line 42 this can be done so that even if data I undefined, it falls back to an empty object

`const getFields = (data = {}, token, t, currentHeight) => ({`

test/cypress/features/blocks.feature
```diff
@@ -0,0 +1,80 @@
...
+		When I click on filterTransactionsBtn
+    Then I should see filterDropdown
+		When I fill 3271 in heightFilter field
+		And I click on applyFilters
+		Then I should see 1 blocks in table
+		When I click on clearAllFiltersBtn
+		Then I should see 20 blocks in table
+		When I click on filterTransactionsBtn
+    Then I should see filterDropdown
+		When I fill 04.01.22 in dateFromInputFilter field
```
eniolam1000752, line 39: there are no consistent tab spacing on this file

test/cypress/features/blocks/blocks.js
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable */ 
+import { Then } from 'cypress-cucumber-preprocessor/steps';
+import { ss } from '../../../constants';
+
+let oldHeight, currentHeight
+
+Then(/^I should see (\d+) blocks in table$/, function (number) {
+  cy.wait(500);
```
eniolam1000752, line 8: I think instead of putting the wait command here, it can be defined in the feature file just before this test case is defined something like 

`When I wait 0.5 seconds`

eniolam1000752 (changes requested):

src/components/screens/monitor/blockDetails/blockDetails.js
```diff
@@ -57,49 +57,49 @@ const getFields = (data, token, t, currentHeight) => ({
...
     label: t('Height'),
-    value: <CopyToClipboard value={data.height} />,
+    value: <CopyToClipboard value={data?.height} />,
   },
   date: {
     label: t('Date'),
     value: (
       <DateTimeFromTimestamp
-        time={data.timestamp * 1000}
+        time={data?.timestamp * 1000}
```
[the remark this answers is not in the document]
ikem-legend: @eniolam1000752 As long as there was a null check in line 73 for data.height, DeepScan suggests making null checks everywhere else data was used. Without the change in line 73, there'd be a NaN error when confirmations first renders. Therefore all the checks are for data (which may or not exist when the component first loads) and not the properties and aside from line 73, the optional checks make no difference

test/cypress/features/blocks/blocks.js
ikem-legend: @eniolam1000752 A lot of the feature tests are set up like this because it prevents verbosity when waiting for a very short period (< 1 second) before proceeding with a test. Therefore writing things like `When I wait for 0.5 seconds` or `When I wait for 0.1 seconds` unnecessarily adds more lines to the tests. In cases, where a common action like clicking on a button has a minor wait time too, writing the extra line in all the features where an element is clicked would be a lot of work without a lot of obvious value

eniolam1000752 (dismissed):
LGTM

reyraa (approved):

eniolam1000752 (approved):
LGTM

reyraa (approved):

PR #4194, opened 2022-03-06
Resolve wrong maximum amount issue - Closes #4190

### LINKED ISSUE #4190: Send maximum amount fails in some cases
If you hit “Send maximum amount”, and then entering the address or adding anything into the data field, the fee calculation will be wrong and fail.

gadjacobs:
### What was the problem?

This PR resolves #4190 

### How was it solved?

- By updating `amount` every time `maxAmount` changes and the user has selected the option to use their entire balance. 

### How was it tested?

Visually


eniolam1000752 (approved):
LGTM

PR #4205, opened 2022-03-14
Support block height search - closes #4132

### LINKED ISSUE #4132: Add support for searching block height
### Description
At the moment we can search for the following values on the blockchain:
 - Address
 - Public key
 - Username
 - Block ID
 - Transaction ID

We should add support for searching block height.

### Motivation
Enable users to search values on the blockchain

### Acceptance criteria
- Users should be able for block height
- The result should be marked as block
- Add sufficient unit test coverage

vibratious:
### What was the problem?

This PR resolves #4132

### How was it solved?

I added support for searching block height in search API function

### How was it tested?

- manually


src/constants/regex.js
```diff
@@ -6,6 +6,7 @@ export default {
   delegateName: /^[a-z0-9!@$&_.]{3,20}$/,
   transactionId: /^[0-9a-z]{64}/,
   blockId: /^[0-9a-z]{64}/,
+  blockHeight: /[0-9]+/,
```
reyraa, line 9: Your changes break the delegate search because the regex you created captures a delegate name that has numbers too.  I think you should change it to

```suggestion
  blockHeight: /^[0-9]+$/,
```
So it only captures numbers and nothing else.

reyraa (changes requested):

src/constants/regex.js
ManuGowda: @vibratious could you please consider Ali's input and update the PR
- Code changes
- Signing the commits

Thank you so much for your contributions 🙏🏼

reyraa (approved):
Good job. I hope to see more contribution from you :raised_hands:

reyraa (approved):

gadjacobs (approved):

eniolam1000752 (approved):
LGTM

vibratious:
thank you for approving, sorry my computer was crashed.

PR #4240, opened 2022-03-31
Refactor settings domain to become blockchain agnostic - Closes #4210

### LINKED ISSUE #4210: Refactor the Setting domain to match the proposed architecture
### Description

As a part of the Make Lisk Desktop blockchain agnostic, we should create a domain for user Settings. Said domain is responsible for collecting, storing locally, and displaying user settings.

### Acceptance Criteria
- Implement togglers that can be used by other components such as navbars, sign-in page and the main settings view.
- Display main settings in a dedicated view.
    - Said view should use commonly reusable togglers and selectors explained above.
- Update the local storage once a setting has been changed.
- Inform users about saving settings.
- Update unit and e2e tests

gadjacobs:
### What was the problem?

This PR resolves #4210 

### How was it solved?

- [ ] Create proposed folder structure
- [ ] Create a reusable checkbox component in the toggles folder
- [ ] Move light/dark mode to toggles folder
- [ ] Move the discrete mode toggle to the toggles folder
- [ ] Move the sidebar toggle to the toggles folder
- [ ] Create a selectors folder for network selector and currency switcher
- [ ] Change `settings.js` to `config.js` (should be developed with dynamism in mind)
- [ ] Create custom hook within the manager that manages how settings are updated and observed
- [ ] Implement the connected network display in the info items folder
- [ ] Move the settings modal to info items

### How was it tested?

Unit tests

### DIFF
--- a/packages/network/store/middleware.js
+++ b/packages/network/store/middleware.js
@@ -1,5 +1,5 @@
 // eslint-disable-next-line no-unused-vars
-import settings from '@settings/configuration/settings';
+import config from '@settings/configuration/settingConstants';
 import { getAutoLogInData, shouldAutoLogIn } from '@common/utilities/login';
 // eslint-disable-next-line no-unused-vars
 import { networkConfigSet, login, settingsUpdated } from '@common/store/actions';
--- a/packages/settings/setters/selectors/currencySelector.js
+++ b/packages/settings/setters/selectors/currencySelector.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import { toast } from 'react-toastify';
+import Select from '@basics/inputs/select';
+import settingConstants from "@settings/configuration/settingConstants"
+import useSettings from "@settings/managers/useSettings"
+import Piwik from '@common/utilities/piwik';
+
+function CurrencySelector({t}) {
+    const {currency, toggleSetting} = useSettings(settingConstants.keys.currency)
+    const onChangeCurrency = (currency) => {
+        Piwik.trackingEvent('Settings', 'button', 'Update settings');
+        toggleSetting(currency)
+        toast(t('Settings saved!'));
+    }
+
+    return(
+        <Select
+            options={settingConstants.currencies.map(currency => ({
+                label: currency, value: currency,
+            }))}
+            selected={currency || settingConstants.currencies[0]}
+            onChange={onChangeCurrency}
+            className="currency"
+            placeholder="Currency"
+        />
+    )
+}
+
+export default CurrencySelector
--- a/packages/settings/setters/toggles/discreteModeToggle.js
+++ b/packages/settings/setters/toggles/discreteModeToggle.js
@@ -1,0 +1,15 @@
+import React from 'react'
+import settingConstants from "@settings/configuration/settingConstants";
+import Toggle from "./toggle";
+
+function DiscreateModeToggle(){
+    return ( 
+        <Toggle
+            setting={settingConstants.keys.discreetMode}
+            icons={['toggleSidebarActive', 'toggleSidebar']}
+            tips={[t('Collapse sidebar'), t('Expand sidebar')]}
+        />
+    )
+}
+
+export default DiscreateModeToggle
--- a/packages/settings/store/middleware.js
+++ b/packages/settings/store/middleware.js
@@ -9,6 +9,7 @@ const settings = store => next => (action) => {
   switch (action.type) {
     case actionTypes.networkConfigSet:
       // store.dispatch(pricesRetrieved());
+      if(!action.data) break;
       store.dispatch(settingsUpdated({
         network: {
           name: action.data.name,
--- a/packages/views/shared/converter/index.js
+++ b/packages/views/shared/converter/index.js
@@ -1,6 +1,6 @@
 /* istanbul ignore file */
 import { connect } from 'react-redux';
-import settings from '@settings/configuration/settings';
+import config from '@settings/configuration/settingConstants';
 import { tokenMap } from '@token/configuration/tokens';
 import Converter from './converter';
 
@@ -10,8 +10,8 @@ const mapStateToProps = state => ({
   priceTicker: (state.service && state.service.priceTicker)
     ? state.service.priceTicker
     : {
-      LSK: settings.currencies.reduce((acc, key) => ({ ...acc, [key]: '0' })),
-      BTC: settings.currencies.reduce((acc, key) => ({ ...acc, [key]: '0' })),
+      LSK: config.currencies.reduce((acc, key) => ({ ...acc, [key]: '0' })),
+      BTC: config.currencies.reduce((acc, key) => ({ ...acc, [key]: '0' })),
     },
 });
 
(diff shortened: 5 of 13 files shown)


packages/views/shared/navigationBars/topBar/topBar.js
```diff
@@ -6,8 +6,10 @@ import Icon from '@basics/icon';
 import DialogLink from '@basics/dialog/link';
 import { PrimaryButton } from '@basics/buttons';
 import Tooltip from '@basics/tooltip/tooltip';
-import Toggle from '@settings/setters/toggles/toggle';
 import VoteQueueToggle from '@settings/setters/toggles/voteQueueToggle';
+import DiscreateModeToggle from '@settings/setters/toggles/discreteModeToggle';
```
ikem-legend, line 10: @eniolam1000752 @gadjacobs Minor typo in `DiscreateModeToggle`

packages/settings/setters/toggles/lightDarkModeToggle.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import settingConstants from '@settings/configuration/settingConstants';
+import Toggle from './toggle';
+
+function LightDarkToggle({ t }) {
```
soroushm, line 5: the translation shouldn't pass from the parent, use the translation hook
```
```
https://react.i18next.com/latest/usetranslation-hook

packages/settings/setters/toggles/toggle.js
```diff
@@ -1,29 +1,40 @@
...
  * Toggles boolean values on store.settings
  *
  * @param {String} setting The key to update in store.settings
  * @param {Array} icons [activeIconName, normalIconName]
  * @param {Array} tips [activeTip, normalTip]
+ * @param {boolean} isCheckbox show checkbox or tooltip
  */
 const Toggle = ({
-  setting, icons, tips,
+  setting, icons, tips, isCheckbox, t,
```
soroushm, line 18: we always need the default props
```
Toggle.defaultProps = {
  isCheckbox: false // or true depend on the logic on this case because the backward compatibility should be false
}
  
  export default Toggle
  ```

File: packages/settings/store/middleware.js [hunk 4]
```diff
@@ -9,6 +9,7 @@ const settings = store => next => (action) => {
   switch (action.type) {
     case actionTypes.networkConfigSet:
       // store.dispatch(pricesRetrieved());
+      if(!action.data) break;
```
soroushm, line 12: it isn't better to write "break" in the second line so eyes can catch it easily?
```suggestion
       if(!action.data) {
            break;
       }
```

soroushm (changes requested):
init review

packages/settings/setters/toggles/discreteModeToggle.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import settingConstants from '@settings/configuration/settingConstants';
+import Toggle from './toggle';
+
+function DiscreateModeToggle({ t }) {
```
reyraa, line 5: Discreet is the correct word. Please update the module and file names.

packages/settings/store/middleware.js
```diff
@@ -9,6 +9,7 @@ const settings = store => next => (action) => {
   switch (action.type) {
     case actionTypes.networkConfigSet:
       // store.dispatch(pricesRetrieved());
```
reyraa, line 11: When will this be reinstated?

packages/settings/store/middleware.js
reyraa, line 12: Although using `if` condition in a `switch` statement is valid, but it kind of means the cases could be designed better. In your case, `networkConfigSet` is always expected to have data. I'd recommend finding the root cause and fixing it instead of pushing bugs under the rug.

reyraa (changes requested):

packages/settings/store/middleware.js
gadjacobs: This change was made by @isalga in the general refactor PR

packages/network/store/middleware.js
```diff
@@ -1,5 +1,5 @@
 // eslint-disable-next-line no-unused-vars
-import settings from '@settings/configuration/settings';
+import config from '@settings/configuration/settingConstants';
```
soroushm, line 2:
```suggestion
import settingConstants from '@settings/configuration/settingConstants';
```

packages/views/shared/converter/index.js
```diff
@@ -1,6 +1,6 @@
 /* istanbul ignore file */
 import { connect } from 'react-redux';
-import settings from '@settings/configuration/settings';
+import config from '@settings/configuration/settingConstants';
```
soroushm, line 3:
```suggestion
import settingConstants from '@settings/configuration/settingConstants';
```

tests/cypress/support/commands.js
```diff
@@ -24,7 +24,7 @@
 // -- This is will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
 import networks from '../../../packages/network/configuration/networks';
-import settings from '../../../packages/settings/configuration/settings';
+import config from '../../../packages/settings/configuration/settingConstants';
```
soroushm, line 27:
```suggestion
import settingConstants from '../../../packages/settings/configuration/settingConstants';
```

packages/settings/setters/selectors/currencySelector.js
```diff
@@ -0,0 +1,34 @@
...
+  const onChangeCurrency = (value) => {
+    Piwik.trackingEvent('Settings', 'button', 'Update settings');
+    toggleSetting(value);
+    toast(t('Settings saved!'));
+  };
+
+  const currencies = useMemo(() => settingConstants.currencies.map((currencyName) => ({
+    label: currencyName,
+    value: currencyName,
+  })), [settingConstants.currencies]);
```
soroushm, line 21: static variable dosnt need to add as dependencies
https://stackoverflow.com/questions/64561538/why-does-passing-a-static-dependency-to-an-useeffect-causes-it-to-be-called-ever
```suggestion
  const currencies = useMemo(() => settingConstants.currencies.map((currencyName) => ({
    label: currencyName,
    value: currencyName,
  })), []);
```

packages/settings/setters/selectors/currencySelector.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <Select
+      options={currencies}
+      selected={currency || settingConstants.currencies[0]}
```
soroushm, line 26: it may couse the exeption in  case currecncies doesn't exist
```suggestion
      selected={currency || settingConstants?.currencies?.[0]}
```

soroushm (changes requested):

packages/settings/setters/selectors/currencySelector.js
eniolam1000752: though since `settingsConstants` is an object of constant values and currencies  would always exist I think the only way an exception can occur is if the constants are deleted

ikem-legend (dismissed):

reyraa (approved):
It looks better now. One thing I noticed though is that the token is still stored in settings. We should move this the Token domain. I think this migration should be accomplished by #4245. Good job.

soroushm (dismissed):

PR #4255, opened 2022-04-07
Fix issues discovered with HWs - Closes #4244

### LINKED ISSUE #4244: Issues discovered while using Ledger Nano S hardware wallet
### Expected behavior
- The hide zero balances option should be visible in the dark mode.
- We should make the login process stable for all HWs.
- The created account should remain there. Probably a good idea to to remove the create new account button and by default create 1 used account. If we reverse the list, this unused account is always at the top and with some visual differences, it should be easy to locate.

### Actual behavior
- The hide zero balances option is not visible in the dark mode.
- The login process seems glitchy when the HW doesn't have any used accounts.
- The last new account goes away and that could be confusing when you have only 1 account. Since the user might not have saved their address or public key, they can't make sure that the account create again is the same as before.

2.3.0-rc.1

reyraa:
### What was the problem?

This PR resolves #4244

### How was it solved?

- The hide empty accounts button was invisible in dark mode
   - Fixed the styles
- The recently created unused account would disappear next time you signed in
   - I removed the create account button, and changed the title to point the user to the first account in the list which is always an unused account. For this, I add an unused account to the list of accounts that I discover in the HW, then reversed the list so the unused account in at the top.
- If the HW doesn't have an account in the current network, the app will get stuck in the loading mode.
   - We were checking against the length of the accounts list to exist loading state, which obviously didn't work when there was no used account in the network. I added a `loaded` flag instead.

### How was it tested?
- Updated unit tests
- Tested manually against mainnet (which I don't have accounts) and testnet (which I do have accounts).


ManuGowda (changes requested):
Still having the same spinning issue, also I observed other errors this time.

Screenshot 2022-04-08 at 9 56 12 AM
Screenshot 2022-04-08 at 9 54 15 AM

src/utils/hwManager.js
```diff
@@ -26,6 +26,23 @@ const getAccountBundle = async (deviceId, network, offset) => {
   return accounts.data.filter(item => item.summary?.address);
 };
 
+const getNewAccountByIndex = async ({ deviceId, index }) => {
+  console.log('getNewAccountByIndex', deviceId, index);
```
ManuGowda, line 30: please remove

ManuGowda (commented):
@reyraa the changes LGTM, please address the comment.

ManuGowda (approved):

Balanced02 (approved):

PR #4241, opened 2022-04-01
Refactor views domain to blockchain agnostic design - Closes #4202

ikem-legend:
### What was the problem?

This PR resolves #4202

### How was it solved?

- [x] Basic components must be in the right place
- [x] Shared components must be in the right place
- [x] Unused components must be removed
- [x] Router should perform correctly and display the right views
- [x] Child components must be imported correctly
- [x] React Context API must be used in place of prop drilling

### How was it tested?

Manually

PR #4214, opened 2022-03-16
General Architecture Refactor - Closes #4201

isalga:
### What was the problem?

This PR resolves #4201 

### How was it solved?

- [x] Renamed files
- [x] Refactored components
- [x] Removed unused components
- [x] Merged or split components and modules
- [x] Updated configs
- [x] Updated tests
- [x] Updated file imports

### How was it tested?

- Running tests
- Building the app in dev and production modes

PR #4261, opened 2022-04-16
Refactor the Network domain to match the proposed architecture - Closes #4208

### LINKED ISSUE #4208: Refactor the Network domain to match the proposed architecture
### Description

As a part of the Make Lisk Desktop blockchain agnostic, we should create a domain for Network. Said domain is responsible for collecting, displaying nodes on the network.

### Acceptance Criteria

- List all nodes on the network.
   - No filtering is required.
   - Apply sorting as currently exists.
- Display the geographical distribution of the loaded nodes on a map.
- Display statistical information about the nodes on the network, including:
   - Peer versions distribution pie chart.
   - Peer height distribution pie chart.
   - Peer connectivity distribution pie chart.

vibratious:
### What was the problem?

This PR resolves #4208

### How was it solved?

- [x] Restructured the network directory according to the changes I noticed in #4214, #4240, and #4241.
- [x] Moved network to packages directory
- [x] Updated unit test

### How was it tested?
Tested visually and updated unit tests


reyraa (dismissed):

packages/network/statistics/index.js
```diff
@@ -0,0 +1,18 @@
+import { compose } from 'redux';
+import withData from "@common/utilities/withData";
+import { withTranslation } from "react-i18next";
+import Statistics from './statistics';
+import { getNetworkStatistics } from '@network/utilities/api';
+import { tokenMap } from '@token/configuration/tokens';
+
+export default compose(
+  withData({
```
isalga, line 9: Should Api call not be moved to a separated manager file?

packages/network/nodesList/manager/manager.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
isalga, line 1: That file should be a repeater and not a manager.

packages/network/nodesList/manager/index.js
```diff
@@ -0,0 +1,29 @@
...
+import withData from "@common/utilities/withData";
+import { DEFAULT_LIMIT } from '@views/configuration';
+import { withTranslation } from "react-i18next";
+import withLocalSort from "@common/utilities/withLocalSort";
+import { sortByVersion } from '../../utilities/helpers';
+import NodeList from "./manager";
+import { tokenMap } from '@token/configuration/tokens';
+
+export default compose(
+  withData({
```
isalga, line 12: Same, Should not API calls be moved to managers?

isalga (changes requested):

ManuGowda:
@vibratious can you please take a look at the minor review comments and address them? after this we can accept your contribution 🙏🏼 .

packages/network/statistics/index.js
vibratious: Hi, The API call is already in a dedicated file. And the API utility is already moved to the network domain directory. I'm not sure what you are aiming to change?

packages/network/nodesList/manager/manager.js
vibratious: The Table is performing the task of a repeater. I don't have distinct repeater myself.

packages/network/nodesList/manager/index.js
vibratious: Similar to the other file. I think it's fine as it is. but please give more info if you think I have to change it.

packages/network/nodesList/manager/manager.js
soroushm: Dear Elnaz,

we try to separate API call logic from the component so the composer is moved to another file called manager so in future we can change the way of API call or reuse the manager

soroushm (approved):

reyraa (approved):

PR #4266, opened 2022-04-27
Restructure the bookmarks domain based on the TECH proposal document - Closes #4209

### LINKED ISSUE #4209: Refactor the Bookmark domain to match the proposed architecture
### Description

As a part of the Make Lisk Desktop blockchain agnostic, we should create a domain for Bookmarks. Said domain is responsible for collecting, displaying bookmarks stored by the user for different chains.

### Acceptance Criteria

- List all bookmarks stored locally.
    - No sorting is required.
    - Add edit capability.
- List all bookmarks as address input suggestions list. (used in the send token form)
    - No sorting is required.
    - No filtering is required.
    - Read-only mode.
- Add bookmark creation form
    - Users must be able to add new bookmarks with address and title for any given chain.
    - Users must not be able to choose a title if the given address is a validator.
- All unit tests must be updated. Also add missing unit test coverage to logical functions.

soroushm:
### What was the problem?
This PR resolves #4209

### How was it solved?
- move bookmark file to own directory
- move searchParams utils to global utils

### How was it tested?
end to end test will resolve with https://github.com/LiskHQ/lisk-desktop/issues/4263


setup/config/webpack.config.js
```diff
@@ -5,12 +5,13 @@ const config = {
   mode: 'development',
   resolve: {
     alias: {
+      src: resolve('./src'),
```
ikem-legend, line 8: @soroushm I believe this should be `'@src'`

src/modules/bookmark/components/BookmarksListModal/BookmarkModalWrapper.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import styles from './bookmarks.css';
+import styles from '@bookmark/components/BookmarksListModal/BookmarkListModal.css';
```
ikem-legend: @soroushm This is in the same folder as the CSS import and should be a relative import

soroushm: good catch

src/modules/bookmark/manager/BookmarkListManager.js
```diff
@@ -2,7 +2,7 @@
 import { connect } from 'react-redux';
 import { withTranslation } from 'react-i18next';
 import { bookmarkRemoved, bookmarkUpdated } from '@common/store/actions';
-import BookmarkListModal from './modal';
+import BookmarkListModal from '@bookmark/components/BookmarksListModal/BookmarkListModal';
```
ikem-legend: @soroushm Since the modal is in the same folder, it should be relative import instead

src/modules/bookmark/utils.js
```diff
@@ -0,0 +1,73 @@
+import { tokenKeys, tokenMap } from '@token/configuration/tokens';
```
gadjacobs, line 1: Can these be split into separate utilities?

soroushm: as i said before if make it as directory they need to be split if its just few of them keep it
don't wast the time for such things we need to finish movement as soon as posible if we have extra time we will spend on it later on this sprint

setup/config/webpack.config.js
soroushm: we can have the src because its not a modules its a root project

ikem-legend (approved):

gadjacobs (approved):

PR #4276, opened 2022-05-02
Move search to the search domain to match the new director structure

eniolam1000752:
### What was the problem?

There is the need to move the search components and its utilities to its own dedicated domain folder

### How was it solved?

- [x] Moved and created a Search component for search.js
- [x] Move and create a SearchBar component for searchBar.js 
- [x] Created TransactionResultList and BlockResultList components and consumed them into the newly created SearchBar component.

### How was it tested?

- Tested visually
- Ran unit tests on the search domain


soroushm (approved):

gadjacobs (approved):
LGTM

PR #4278, opened 2022-05-03
Restructure the wallet domain to match the consented scaffold

reyraa:
### How was it solved?
- [x] Moved the components
- [x] Moved relevant utils
- [x] Updated aliases and the router
- [x] Adjusted styles
- [x] Fixed linting and other issues


.eslintrc
```diff
@@ -43,7 +43,7 @@
           ["@token", "./packages/token/"],
           ["@transaction", "./src/modules/transaction"],
           ["@updater", "./packages/updater/"],
-          ["@wallet", "./packages/wallet/"],
+          ["@wallet", "./src/modules/wallet/"],
```
soroushm, line 47: jsconfig also needs to be fix

soroushm (dismissed):

PR #4285, opened 2022-05-04
Restructure common and utils to match the tech proposal

soroushm:
- [x] move common files
- [x] move utils files


src/modules/common/components/FormattedNumber/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './index';
```
ikem-legend: @soroushm The index file exports itself rather than the FormattedNumber component

src/theme/tabs/switcher/demo.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { MemoryRouter } from 'react-router-dom';
-import DemoRenderer from '@basics/demoRenderer';
+import DemoRenderer from '@views/../../demo/demoRenderer';
```
ikem-legend: @soroushm Import seems incorrect here as it's both an aliased and relative route

ikem-legend (approved):

gadjacobs (approved):

PR #4286, opened 2022-05-05
Restructure the file left in packages to match the tech proposal 

soroushm:
- [ ] split store to the match domain
- [ ] move common file
- [x] move delegate


reyraa (changes requested):
I'll summarize changes on the feature branch so it hopefully helps with merging your PR: 
- editVote is already moved.
- multisignature is moved and renamed.
- BlockDetails is moved and adapted to use the transactionRow from the transaction domain
- BlockRow is moved
- Block middleware is moved. I've also split the logic into their own domains according to new structure
- block API and block subscribe are also moved
- The send form is fully moved. I've also made sure all types of transactions work.
- VoteQueue is fully moved and it's now working
- The Message module (sign an confirm views) is fully moved.
- Search functionality is moved and ready.
- Network selector is fixed. I've also removed the redundant "dev net" and updated unit tests and e2e tests.

Probably there's more that has been moved already and I overlooked in your PR. but for now, please discard your changes to the above directories to prevent unnecessary conflicts. 
Let me know if I can help.

PR #4313, opened 2022-05-17
Implement enter account password form - Closes #4298

### LINKED ISSUE #4298: Implement enter account password form
### Description
Enable user to decrypt the json backup schema by entering their password

### Acceptance Criteria
- Implement sign in form according to [design](https://www.figma.com/file/GbrsJiZlsh2OzVKqYDD4zI/Account-management-improvements?node-id=243%3A15640)
- Allow user to reset password by redirect to the reset password flow
- Allow user to decrypt encryption by the password 
- Add unit test to make sure decryption working by entering password
- Show error if decryption gets failure

### Additional Information

- Component will receive the json backup schema as props
- Component will return on success a encrypted recovery phrase / private key
- Title and subtitle should receive from the props 
- This component also will use on remove account, sign a transaction, sign a message

isalga:
### What was the problem?

This PR resolves #4298

### How was it solved?

Implement design

### How was it tested?

Add unit test


src/modules/auth/components/EnterPasswordForm/enterPasswordForm.js
```diff
@@ -0,0 +1,60 @@
...
+        />
+        <PrimaryButton
+          className={styles.button}
+          onClick={onSubmit}
+          disabled={!password}
+        >
+          {t('Continue')}
+        </PrimaryButton>
+      </BoxContent>
+    </Box>
```
eniolam1000752: I think this should be managed by a form so that the perks of being able to hit enter and an onsubmit callback is triggered can be preserved.

Also as a suggestion, you can use useForm from `react-hook-form` to manage the form.

eniolam1000752 (changes requested):

src/modules/auth/components/EnterPasswordForm/enterPasswordForm.js
soroushm: @eniolam1000752  is just a input without validation, we can keep it like that for now

src/modules/auth/components/EnterPasswordForm/enterPasswordForm.js
```diff
@@ -0,0 +1,60 @@
...
+      <BoxContent className={styles.content}>
+        <h1>{t('Enter your password')}</h1>
+        <p className={styles.subheader}>
+          {t('Please provide your device password to backup the recovery phrase.')}
+        </p>
+        <WalletVisual
+          className={styles.avatar}
+          address={accountSchema?.metadata?.address}
+        />
+        <p className={styles.accountName}>{accountSchema?.metadata?.name}</p>
```
soroushm: name is optional 
{accountSchema?.metadata?.name &&  <p className={styles.accountName}>{accountSchema?.metadata?.name}</p>}

src/modules/auth/components/EnterPasswordForm/enterPasswordForm.test.js
```diff
@@ -0,0 +1,40 @@
...
+        value: 'qwerty',
+      },
+    });
+    wrapper.find(styles.button).first().simulate('click');
+    expect(props.onEnterPasswordSuccess).toHaveBeenCalledWith({
+      privateToken: 'private-token-mock',
+      recoveryPhrase: 'target cancel solution recipe vague faint bomb convince pink vendor fresh patrol',
+    });
+  });
+});
```
soroushm, line 71: we need to have test for unhappy path
like when the password is wrong `onEnterPasswordSuccess` shouldn't trigger

soroushm (dismissed):

src/modules/auth/components/EnterPasswordForm/enterPasswordForm.test.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    ));
+    wrapper = mount(<EnterPasswordForm {...props} />);
+
+    wrapper.find('input').at(0).simulate('change', {
+      target: {
+        value: 'qwerty',
+      },
+    });
+    wrapper.find(styles.button).first().simulate('click');
```
soroushm: dear, could you please change the find method,
in front-end, we will change the styles often so it's will end up with a broken test

eniolam1000752 (dismissed):

src/modules/auth/components/EnterPasswordForm/enterPasswordForm.test.js
isalga: Done. No need to call me dear.

eniolam1000752 (dismissed):

soroushm (approved):

eniolam1000752 (approved):

Issue #4263, opened 2022-04-22
Restore end to end tests


soroushm:
## Description
we are moving all files and components to the new structure base on domain driven after refactoring the project structure  by moving the components and relevant styles, assets, utility functions, configuration files  into domain directories need to end test is failed because the file pass is going wrong in order to get built from Jenkins we need to pass the test

## Problem
- Jenkins CI is failed because the test not running correctly

## Acceptance Criteria
- The Test structure must match the TECH design proposal document.
- All end to end tests should be pass
- CI should work properly

## Additional information

- Remove [this](https://github.com/LiskHQ/lisk-desktop/pull/4292/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R61)
- Uncomment [this pipeline](https://github.com/LiskHQ/lisk-desktop/pull/4292/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R69)
- Branch you need to work on is [here](https://github.com/LiskHQ/lisk-desktop/tree/fix/e2e-tests)
- the related pull request for the issue #4318

PR #4318, opened 2022-05-20
Restore end to end tests - Closes #4323, #4263

### LINKED ISSUE #4323: Investigate root cause e2e failing
## Description
we are moving all files and components to the new structure base on domain driven after refactoring the project structure  by moving the components and relevant styles, assets, utility functions, configuration files  into domain directories need to end test is failed because the file pass is going wrong in order to get built from Jenkins we need to pass the test

## Problem
- Jenkins CI is failed because the test not running correctly

## Acceptance Criteria
- Find the reason why its not working
- Fix cypress config
- we are not looking to pass all the testes

## Additional information

- Remove [this](https://github.com/LiskHQ/lisk-desktop/pull/4292/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R61)
- Uncomment [this pipeline](https://github.com/LiskHQ/lisk-desktop/pull/4292/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R69)
- Branch you need to work on is [here](https://github.com/LiskHQ/lisk-desktop/tree/fix/e2e-tests)
- the related pull request for the issue #4318

ikem-legend:
### What was the problem?

This PR resolves #4323 

#### Resolved the following
- Changed alias importation in tests/constants/networks.js
- Changed alias importation in tests/cypress/features/login/login.js

#### Observed the following issues that needs to be resolved in #4263 to get e2e up and running.
- Accounts gherkin script can’t run d due to the fact that there is a change in the route `/accounts`
- For every new browser instances, there is a `TypeError: Cannot read properties of undefined (reading 'networks')` exception been thrown which means any gherkin script with a login step fails on cypress.
this issue occurs because there in no network parameter in the settings object when `readStoredNetwork` in `src/modules/network/store/middleware.js` tries to get the selected stored network in the localStroage.


tests/constants/networks.js
```diff
@@ -1,6 +1,5 @@
 /* global Cypress */
-// import networks from '@network/configuration/networks';
-import networks from '@network/configuration/networks';
+import networks from '../../src/modules/network/configuration/networks';
```
soroushm, line 2: is the alias path working here like below?
```suggestion
import networks from '@network/configuration/networks';
```

soroushm (commented):
looks good,

reyraa (approved):
Great job @eniolam1000752

tests/constants/networks.js
eniolam1000752: nope it doesn't I have also tried adding the webpack plugin that enables cypress to extend its build configurations to our own webpack config but that still didn't work.

soroushm (approved):

Issue #4334, opened 2022-06-02
Fetch account data for account management


soroushm:
### Description
we introduce login by encryption JSON and remove the recovery phrase from the redux we need to update the redux and API call to match with a new structure

### Acceptance Criteria

- on add /switch account, data should be fetched
  - data from the dashboard should fetch
  - data from the wallet should fetch
  - sign multi-signature/message and votes should be available 
- adopt unit test

ManuGowda:
@soroushm the issue is not upto date, we need to add all the RPC endpoints required for account management

soroushm:
we will integrate the endpoint on #4424, #4429, #4400 tickets

PR #4337, opened 2022-06-07
Implement useAccounts hooks - closes #4308

### LINKED ISSUE #4308: Implement useAccounts hook and persist data to the local store
### Description
Enable the application have access to user accounts stored in the local store

### Acceptance Criteria
- Implement a hook in account domain allow application get list and set an account
     ```javascript
  const {
    accounts,
    setAccount,
    deleteAccountWithAddress,
    getAccountWithAddress,
  } = useAccounts()
  
- accounts needs to return list of account from redux
- setAccount need to be set account  in redux
- deleteAccountWithAddress need to remove the specific account from redux
- getAccountWithAddress needs to return the  specific account from redux
- account list in redux needs to store in localStorage and sync with it on updates

### Additional Information
[redux-persist](https://github.com/rt2zz/redux-persist) is useful to store data from redux in localStorage or any other storage

soroushm:
### What was the problem?

This PR resolves #4308
- Implement useAccounts hook and persist data 

### How was it solved?
- implement a custom hook working with redux to store data
- config redux persists to store data on local storage

### How was it tested?
- add unit test for it


gadjacobs:
@soroushm we need two more functions, one for removeAccount(address) and another for getAccount(address)

src/modules/account/hooks/useAccounts.js
```diff
@@ -1,69 +1,14 @@
...
+import { useSelector, useDispatch } from 'react-redux';
+import { selectAccounts } from '@account/store/selectors';
+import { addAccount } from '../store/action';
 
 // eslint-disable-next-line
-export function useAccounts(){
-  const [accounts, setAccount] = useState(mockAccounts);
+export function useAccounts() {
+  const dispatch = useDispatch();
+  const setAccount = (accountSchema) => {
```
ManuGowda: bit confused here why accountSchema variable used here? we should choose the naming appropriately.
- Schema is required to encode or decode account or to do validation, i see here nothing of that sort is done here

packages/common/store/index.js
```diff
@@ -1,17 +1,20 @@
 import {
   createStore, combineReducers, applyMiddleware, compose,
 } from 'redux';
-
+import storage from 'redux-persist/lib/storage';
+import { persistStore } from 'redux-persist';
```
ManuGowda, line 5:
```suggestion
import { persistStore, storage } from 'redux-persist';
```

ManuGowda: also could you explain bit about usage of `redux-persist` and what we are trying to persist? so that other devs also know about it 🙏🏼

src/modules/account/components/AddAccountBySecretRecovery/AddAccountBySecretRecovery.js
```diff
@@ -6,26 +6,28 @@ import SetPasswordForm from 'src/modules/auth/components/SetPasswordForm/SetPass
...
 
   const onSetPassword = (account) => {
     setCurrentAccount(account);
+    setAccount(account);
     multiStepRef.current.next();
   };
 
   const onPasswordSetComplete = () => {
-    login(passphrase); // Todo this login method is depricated
+    login(passphrase); // Todo this login method is deprecated
```
ManuGowda: can we not use login, signin or any other terminology to avoid further confusion? if you don't the terminology we can brainstorm with the team?

following are some initial thoughts:
- authenticate? : when a user choose an account and navigates to wallet
- authorize?: when a user choose to use his wallet (any action associated with token usage, whereby requiring password)

ManuGowda (commented):
@soroushm i have left minor comments, you need to delete the root package-lock file and regenerate, because both root and app/lock file should be updated

packages/common/store/index.js
soroushm: storage isn't export in the root component 
we can use [Window.localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) but in any application is forbidden to interact global object like the window so we access in from the libraries 
redux persist is the libraries to persist redux state to any storage such as localStorage, react-native async storage [more info](https://github.com/rt2zz/redux-persist)

src/modules/account/components/AddAccountBySecretRecovery/AddAccountBySecretRecovery.js
soroushm: this method should complete remove after #4334 is done 
i just fix the typo on comment

src/modules/account/hooks/useAccounts.js
ManuGowda: @soroushm why was this marked resolved? I think the usage of `accountSchema` naming is not correct.

soroushm:
> @soroushm we need two more functions, one for removeAccount(address) and another for getAccount(address)

@gadjacobs the functions is implemented on the pr

src/modules/account/hooks/useAccounts.js
soroushm: @ManuGowda the name is already changed (perhaps the commit wasn't pushed at the time)

packages/common/store/index.js
```diff
@@ -21,5 +24,10 @@ if (module.hot) {
...
 }
 
-export * from './selectors';
+export {
+  storage,
+  store,
+  persistedStore,
+};
+
 export default store;
```
ikem-legend: @soroushm Since store is exported as a named component above, why is it still exported as a default here, when it's not imported by default anywhere else

src/modules/account/hooks/useAccounts.js
```diff
@@ -1,69 +1,27 @@
...
-  const [accounts, setAccount] = useState(mockAccounts);
+export function useAccounts() {
+  const dispatch = useDispatch();
+  const accountsObject = useSelector(selectAccounts);
+  const setAccount = useCallback((account) => dispatch(addAccount(account)), []);
+  const deleteAccountWithAddress = useCallback(
+    (address) => dispatch(removeAccount(address)),
+    [],
+  );
+  const getAccountWithAddress = (address) => accountsObject[address];
```
reyraa: Grammatically incorrect. Rename this to `getAccountByAddress`.

src/modules/account/hooks/useAccounts.js
```diff
@@ -1,69 +1,27 @@
...
+import { addAccount, removeAccount } from '../store/action';
 
 // eslint-disable-next-line
-export function useAccounts(){
-  const [accounts, setAccount] = useState(mockAccounts);
+export function useAccounts() {
+  const dispatch = useDispatch();
+  const accountsObject = useSelector(selectAccounts);
+  const setAccount = useCallback((account) => dispatch(addAccount(account)), []);
+  const deleteAccountWithAddress = useCallback(
```
reyraa: Grammatically incorrect. Rename this to `removeAccountByAddress`.

src/modules/account/store/selectors.js
```diff
@@ -1,2 +1,3 @@
 // eslint-disable-next-line import/prefer-default-export
 export const selectCurrentAccount = state => state.account.current;
```
reyraa, line 2: What do you want to do with `selectActiveTokenAccount` then?

packages/common/store/index.js
```diff
@@ -1,17 +1,20 @@
...
-const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+export * from './selectors';
 
-const store = createStore(App, composeEnhancers(applyMiddleware(...middleWares)));
+const rootReducer = combineReducers(reducers);
+const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 
+const store = createStore(rootReducer, composeEnhancers(applyMiddleware(...middleWares)));
 store.dispatch({ type: actionTypes.storeCreated });
+const persistedStore = persistStore(store);
```
reyraa, line 17: These changes are not relevant. Please open a ticket, explain your goal and submit changes via the corresponding PR.

reyraa (changes requested):

src/modules/account/components/ManageAccounts/ManageAccounts.js
```diff
@@ -68,4 +72,9 @@ const ManageAccounts = ({ onSelectAccount, onAddAccount, onRemoveAccount }) => {
   );
 };
 
+ManageAccounts.defaultProps = {
+  onSelectAccount: () => null,
```
reyraa: Why `null`? what is the situation where it makes sense to not pass the `onSelectAccount` function, and let it call a pointless function that literally don't nothing?

src/modules/account/components/ManageAccounts/ManageAccounts.js
```diff
@@ -68,4 +72,9 @@ const ManageAccounts = ({ onSelectAccount, onAddAccount, onRemoveAccount }) => {
...
+  onAddAccount: () => null,
```
reyraa: Same here.

reyraa (changes requested):

src/modules/account/store/reducer.js
```diff
@@ -1,19 +1,55 @@
...
     default:
       return state;
   }
 };
 
+/**
+ *
+ * @param {Object} state
+ * @param {type: String, accountSchema: Object} action
+ */
```
reyraa: Please complete this documentation.

src/modules/account/store/reducer.js
```diff
@@ -1,19 +1,55 @@
 import { combineReducers } from 'redux';
+import { persistReducer } from 'redux-persist';
+import { storage } from '../../../../packages/common/store/index';
 import actionTypes from './actionTypes';
 
 /**
  *
  * @param {Object} state
  * @param {type: String, accountSchema: Object} action
  */
```
reyraa: Please complete this documentation.

src/modules/account/store/reducer.js
```diff
@@ -1,19 +1,55 @@
 import { combineReducers } from 'redux';
+import { persistReducer } from 'redux-persist';
+import { storage } from '../../../../packages/common/store/index';
```
reyraa, line 3: Please use the relevant alias.

reyraa (changes requested):

packages/common/store/index.js
soroushm: @reyraa I already explain it to the team and its my pleasure to briefly explain to you, I believe this change is part of the acceptance criteria!

src/modules/account/store/selectors.js
soroushm: its gonna be remove! 
we have the new selector for it, also we have the useCurrentAcccount hook for the account
 by the way those change is not related to this PR!!

src/modules/account/components/ManageAccounts/ManageAccounts.js
soroushm: that point less function is avoid the break application and its the correct way to do

src/modules/account/components/ManageAccounts/ManageAccounts.js
soroushm: same answar

src/modules/account/store/reducer.js
soroushm: do we have one? by the way its need to be change in different task

src/modules/account/hooks/useAccounts.js
soroushm: while we use get, set its need to be called delete like rest method

soroushm: while we use get, set its need to be called delete like rest method

src/modules/account/components/ManageAccounts/ManageAccounts.js
reyraa: Please remove the default prop and make the property required.

packages/common/store/index.js
reyraa: no need. If others agreed, it's fine.

src/modules/account/hooks/useAccounts.js
reyraa: It's fine to call it remove or delete. Pick a choice. The problem is `With`. It should be `By`. Please change it.

src/modules/account/components/ManageAccounts/ManageAccounts.js
reyraa: Please remove the default value, and make the property required.

src/modules/account/hooks/useAccounts.js
reyraa: The problem is `With`. It should be `By`. Please change it.

src/modules/account/store/reducer.js
reyraa: Please do not resolve the comment without fixing it, or leaving a comment. Otherwise why do we bother with reviewing the PRs?

src/modules/account/store/reducer.js
reyraa: I unresolve this as it's not fixed.

reyraa (changes requested):

src/modules/account/hooks/useAccounts.js
soroushm: its already fixed

src/modules/account/store/reducer.js
soroushm: what you expect to have here?

src/modules/account/components/ManageAccounts/ManageAccounts.js
soroushm: you need to go over the changes those function are removed from props

src/modules/account/store/action.js
```diff
@@ -8,7 +8,17 @@ import actionTypes from './actionTypes';
...
+  encryptedAccount,
+});
+
+export const addAccount = (encryptedAccount) => ({
+  type: actionTypes.addAccount,
+  encryptedAccount,
+});
+
+export const deleteAccount = (address) => ({
+  type: actionTypes.removeAccount,
```
ikem-legend: @soroushm I feel like there should be consistency in the name of the action and the action type just like `addAccount`

src/modules/account/store/action.test.js
```diff
@@ -18,4 +19,14 @@ describe('actions: account', () => {
...
+
+  describe('add account', () => {
+    it('should create an action to add account', () => {
+      const expectedAction = {
+        type: actionTypes.addAccount,
+      };
+
+      expect(addAccount()).toEqual(expectedAction);
+    });
+  });
```
ikem-legend, line 36: @soroushm `deleteAccount` is not tested here

soroushm: done

ikem-legend (approved):

reyraa (approved):

PR #4351, opened 2022-06-23
Update initial routes to new account management - Closes #4350

### LINKED ISSUE #4350: Adopt initial routes with new account management
### Description
we don't have login or logout functionality any more, so while user open the application its need to chose or add an account to the device.
we need to open account management on the bootstrap, just if user doesn't have any application it should redirect to the add account flow

### Acceptance Criteria
- open application management as first init route "/account"
- Route "/account" must have an condition, if there isn't any account on the list it should be redirect to the "/account/add"
- While user will redirect to the add flow the browser history should be replace by new url

### Additional Information
   here is a great example how you can handle conditional route [redirects-in-react-router](https://stackabuse.com/redirects-in-react-router/)

gadjacobs:
### What was the problem?

This PR resolves #4350 

### How was it solved?

- [x]  Update root URL component to manage account
- [x] Update dashboard page route to '/dashboard'
- [x] Add conditional logic to display add account page on launch

### How was it tested?

Visually


src/modules/common/components/customRoute/index.js
```diff
@@ -32,17 +33,24 @@ const CustomRoute = ({
...
   const { search = '' } = history.location;
+  const { accounts } = useAccounts();
 
   Piwik.tracking(history, token);
 
   if (forbiddenTokens.indexOf(token.active) !== -1) {
     return <Redirect to={`${routes.dashboard.path}`} />;
   }
 
+  if (!isAuthenticated && path === routes.dashboard.path) {
```
soroushm: `!isAuthenticated && path === '/'`

src/modules/common/components/customRoute/index.js
```diff
@@ -32,17 +33,24 @@ const CustomRoute = ({
   const isNetworkSet = useSelector(checkNetwork);
   const isAuthenticated = !!wallet.summary;
```
soroushm:
```suggestion
const [account] = useCurrentAccount()
  const isAuthenticated = !!account
```

soroushm (changes requested):

src/modules/common/components/customRoute/index.js
```diff
@@ -30,19 +31,30 @@ const CustomRoute = ({
...
-  const isAuthenticated = !!wallet.summary;
+  const [account] = useCurrentAccount();
+  const isAuthenticated = Object.keys(account).length > 0;
   const { search = '' } = history.location;
+  const { accounts } = useAccounts();
+  const hasAccounts = accounts.length > 0;
+
+  console.log(!isAuthenticated, path === routes.manageAccounts.path, !hasAccounts);
+
+  console.log('current acc--', account);
```
ManuGowda: Please remove

src/modules/common/components/customRoute/index.js
```diff
@@ -30,19 +31,30 @@ const CustomRoute = ({
   const wallet = useSelector(state => selectActiveTokenAccount(state));
   const token = useSelector(state => state.token);
   const isNetworkSet = useSelector(checkNetwork);
-  const isAuthenticated = !!wallet.summary;
+  const [account] = useCurrentAccount();
+  const isAuthenticated = Object.keys(account).length > 0;
   const { search = '' } = history.location;
+  const { accounts } = useAccounts();
+  const hasAccounts = accounts.length > 0;
```
ManuGowda: if the logic is not reused, its better to use them directly to avoid unnecessary allocation

packages/views/screens/router/routes.js
```diff
@@ -71,9 +71,10 @@ export default {
     forbiddenTokens: [],
   },
   manageAccounts: {
-    path: '/account',
+    path: '/',
```
ManuGowda, line 74: Informational: @soroushm the default route is not finalised yet, for now i understand we do not have sign in anymore and this is ok. But we might want to change this at later point in time based on new design changes to dashboard.

ManuGowda (changes requested):
@gadjacobs left few minor comments.

ikem-legend (approved):

src/modules/common/components/customRoute/index.js
```diff
@@ -30,15 +31,21 @@ const CustomRoute = ({
...
   const { search = '' } = history.location;
+  const { accounts } = useAccounts();
 
   Piwik.tracking(history, token);
 
   if (forbiddenTokens.indexOf(token.active) !== -1) {
     return <Redirect to={`${routes.dashboard.path}`} />;
   }
 
+  if (!isAuthenticated && path === routes.manageAccounts.path && !(accounts.length > 0)) {
```
ManuGowda, line 45: this is redundant? `!(accounts.length > 0)` it is same as `!isAuthenticated`

soroushm: @ManuGowda accounts is the list of account from the account manager

ManuGowda (approved):

ManuGowda (approved):

ikem-legend (approved):

soroushm (approved):

PR #4387, opened 2022-07-05
Create useCurrentApplication hooks - Closes #4373

### LINKED ISSUE #4373: Create useCurrentApplication Hooks
### Description
A Custom hooks for the current application is an interface to handle all the logic related current application, this enables the component/hooks to use applications logic from one single place

### Acceptance Criteria

- Allow to set current application on redux
- Allow to get current application from redux
- Set lisk Application as default application on bootstrap
- Allow user to select the node URL if there is more than one
- Add required unit tests

	
### Additional Information
Suggestion component name `useCurrentApplication`

```
function useCurrentApplication() {
 const dispatch = useDispatch();
 const currentApplication = useSelector(selectCurrentApplication);
 const setCurrentApplication = useCallback((application) => dispatch(setCurrentApplication(application)), []);
  return [currentApplication, setCurrentApplication];
}
```

soroushm:
### What was the problem?
This PR resolves #4373
- [x] Create useCurrentApplication hooks
- [x] Move files to match the domain structure

### How was it solved?
By create a custom hooks and related redux store

### How was it tested?
by unit test


src/modules/blockchainApplication/explore/api/index.js
```diff
@@ -1,27 +1,2 @@
...
-  params = {},
-  baseUrl,
-}) => http({
-  path: httpPaths.blockchainAppsStatistics,
-  network,
-  params,
-  baseUrl,
-});
+export * from './applicationExplore';
+export * from './applicationStatistics';
```
isalga, line 2: I don't think that creating 3 files for just one function is good architecture. I think decoupling files should be done when needed (Separation of concerns), but not for nesting exports and creating single function files, that results on an over-scaffolded file structure.

src/modules/blockchainApplication/explore/manager/BlockchainApplicationStatistics.js
```diff
@@ -1,8 +1,8 @@
 /* istanbul ignore file */
 import { compose } from 'redux';
 import withData from 'src/utils/withData';
-import ManageBlockchainApplicationsView from '../../components/ManageBlockchainApplications';
-// import { getApps, getStatistics } from '../../../explore/utils/api';
+import ManageBlockchainApplicationsView from '../components/BlockchainApplications';
+// import { getStatistics } from '../../../explore/api';
```
isalga, line 5: This file seems outdated from `feature/blockchain-application`. Please check `src/modules/blockchainApplucation/manage/manager/ManageBlockchainApplications/BlockchainApplicationsStatistics.js` and move it to here if need. `getStatistics` functions was unmocked and uncomented

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
```diff
@@ -0,0 +1,15 @@
+import { useCallback } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import { selectCurrentApplication } from '../store/selectors';
+import { setCurrentApplication } from '../store/action';
+
+// eslint-disable-next-line
```
isalga: Which is the reason eslint disabled here?

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.test.js
```diff
@@ -0,0 +1,38 @@
...
+    const [, setCurrentApplication] = result.current;
+    const expectedAction = {
+      type: actionTypes.setCurrentApplication,
+      application: mockApplications[0],
+    };
+    act(() => {
+      setCurrentApplication(mockApplications[0]);
+    });
+    expect(mockDispatch).toHaveBeenCalledTimes(1);
+    expect(mockDispatch).toHaveBeenCalledWith(expectedAction);
```
isalga, line 41: Would be good to add a case where it test that calling `useCurrentApplication` returns the correct application.

src/modules/blockchainApplication/explore/api/index.js
soroushm: so usually is the best way its have separate a file for each utils so over time they don't get fat and ugly, 
in this spacial case because we mock the HTTP and it has difference between them its needs to be split into two file

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
soroushm: we try to use normal export, (i remember two month ago we discuss about it) the reason is easy can export all hooks or utils, and because the import name is equal to the export name its much easier to find the function

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.test.js
soroushm: good catch its updated

src/modules/blockchainApplication/explore/manager/BlockchainApplicationStatistics.js
soroushm: @isalga so API directory is on the root of each domain so I just fix the comment path and I do blive its correct the correct path of not please mention where and why

src/modules/blockchainApplication/explore/api/index.js
isalga: That I agree, but we did not reach that point and from requirements does not look we will. If files get too large thats why we hace a lint rule of maximum 300 lines. I agree with reducing the amount of lines if need, but right now for me this is counterproductive decoupling. We will end up spending way too much time creating files for just 1 function for no gain. I don't see why mocking HTTP call affects the architecture, specially because it is mocked in HOC component level and not on api util

src/modules/blockchainApplication/explore/manager/BlockchainApplicationStatistics.js
isalga: Ok, my bad, it was unmocked and uncomented in another ticket in progress (https://github.com/LiskHQ/lisk-desktop/pull/4382) and not in feature branch

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
isalga: `import { useCurrentApplication } from `[smth]/useCurrentApplication.js` looks strange to me. I think `import useCurrentApplicatio from `[smth]/useCurrentApplication.js` is more correct as function name is same as file name. I would favour `export default useCurrentApplication` in this case. Avoid eslint expections is a good practise in my opinion

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.test.js
```diff
@@ -0,0 +1,38 @@
...
+  });
+});
```
eniolam1000752, line 43: I think we can also have a test to assert if it returns the right current application data

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
soroushm: then we can talk about it in style guideline process please skip it, for now, we have already implementation on this way
by the way, You do the same, please check getStatistics export, its also not a default export

src/modules/blockchainApplication/explore/manager/BlockchainApplicationStatistics.js
soroushm: then please after merge this branch correct the other one

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
eniolam1000752: I think the exact rule to disable can be used i.e
```suggestion
// eslint-disable-next-line import/prefer-default-export
```

src/modules/blockchainApplication/manage/store/selectors.js
```diff
@@ -1,2 +1,3 @@
 // eslint-disable-next-line import/prefer-default-export
```
eniolam1000752: I think this eslint ignore can be removed now since there are two named exports here

eniolam1000752 (changes requested):

src/modules/blockchainApplication/explore/api/index.js
soroushm: here is the reason on the test file 

application list
```
jest.mock('src/utils/api/http', () =>
  jest.fn().mockImplementation(() => Promise.resolve({ data: [{ type: 0 }] })));
```

application statics
```

jest.mock('src/utils/api/http', () =>
  jest.fn().mockImplementation(() => Promise.resolve({
    data: {
      registered: 2503,
      active: 2328,
      terminated: 35,
      totalSupplyLSK: '5000000',
      stakedLSK: '3000000',
      inflationRate: '4.50',
    },
  })));

```

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.test.js
soroushm: you'r late :) iris already mention that one look at line 26

src/modules/blockchainApplication/manage/store/selectors.js
soroushm: the rule say its need to have export default it doesn't care how many exports it has

eniolam1000752: 🤔  I think if there are more than one named exports in a file, there would be no need for this rule `import/prefer-default-export` excpetion.

soroushm: you are right

eniolam1000752 (approved):
LGTM

isalga (approved):

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
isalga: That function is that way because you in a PR review told me to do it that way.

src/modules/blockchainApplication/explore/api/index.js
isalga: That is an unit test mock. I was talking about mocking in runtime. I don't see neither how mocking HTTP module affects file structure still

PR #4394, opened 2022-07-11
Add support for delegate key management - Closes #4383

### LINKED ISSUE #4383: Update delegate registration process to accept delegate keys
### Description
Currently the delegate registration form only accepts the delegate username. The recent changes of Lisk SDK require delegates to also register
- blsPublicKey
- generatorPublicKey
- proofOfPossession
The changes are captured in the [new design guides](https://www.figma.com/file/Krew1j8uTJVfb6RuojImTK/Introduce-delegate-key-management-%2F-Implement-interoperable-token-usage?node-id=1%3A3).

Users should be able to run

```
Lisk keys:create
```
To create a new account. The result will include the required keys. Users could pass an existing secret recovery phrase using the `--passphrase` parameter, and store the result in a file using the `-f` parameter. Users then use the above keys from the result to fill in the delegate registration form.

### Acceptance Criteria
- The form should:
   - Enable users to paste the following values:
      - blsPublicKey
      - generatorPublicKey
      - proofOfPossession
   - Validate each value and display potential error messages.
   - Display tooltips next to each input to guide users on how to collect the corresponding value.
- Users should be able to:
   - Review the values in the transaction summary step.
   - Submit the transaction and register a new delegate.
- Add unit tests

### Additional information
Since the API is not ready, we should use mock values. Also, the e2e tests can only be updated when real API calls are possible. Therefore we created a new ticket to tackle this problem.

reyraa:
### What was the problem?
This PR resolves #4383

### How was it solved?
- Added `useDelegateKey` hook to validate and store delegate keys.
- Added `useDelegateName` hook to validate, check existence/uniqueness and store delegate name.
- update the registration form and tx summary screens to work with the new values

NOTE:

1. After aligning with the designer, I've implemented the typography of the summary screen similar to other existing transactions.
2. The transaction won't go through by default
   a. You should change the mock to an account that is not a delegate in your network
   b. You should mock the tx to return 200 even though the new schema is not supported yet. 
   c. Once you did the above and tested the PR, you have to revert your changes so that other features work.

### How was it tested?
- Updated unit tests
- Added e2e tests but commented out since the transactions endpoint doesn't support it.


isalga (dismissed):

src/modules/dpos/validator/components/RegisterDelegateForm/InputLabel.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import Tooltip from 'src/theme/Tooltip';
+import styles from './form.css';
+
+const InputLabel = ({ field }) => {
```
soroushm: as component should be reusable could you pass title and tip from the parent so it can be reuse in the other cases

src/modules/dpos/validator/components/RegisterDelegateForm/index.js
```diff
@@ -90,39 +64,68 @@ const RegisterDelegateForm = ({
...
-              {t('Your username')}
-              <Tooltip position="right">
-                <p>
-                  {t(
-                    'Max. 20 characters, a-z, 0-1, no special characters except !@$_.',
-                  )}
-                </p>
-              </Tooltip>
-            </label>
+            <InputLabel field="name" />
```
soroushm: for an example pass values from the parent
```suggestion
            <InputLabel {...config.name} />
```

src/modules/dpos/validator/components/RegisterDelegateForm/index.js
```diff
@@ -1,82 +1,56 @@
...
   prevState,
 }) => {
-  const timeout = useRef();
   const { t } = useTranslation();
-  const network = useSelector(state => state.network);
-  const [error, setError] = useState('');
-  const [loading, setLoading] = useState(false);
-  const [username, setUsername] = useState(prevState?.rawTx?.asset.username ?? '');
+  const [name, setName] = useDelegateName(prevState?.rawTx?.asset.username);
+  const [generatorPublicKey, blsPublicKey, proofOfPossession, setKey] = useDelegateKey(
```
soroushm: recently we introduce [react-hooks-form](https://react-hook-form.com) please use that one so it dosn't need to create custom validation you can use [yup validation](https://github.com/jquense/yup#stringmatchesregex-regex-message-string--function-schema)

src/modules/dpos/validator/hooks/useDelegateKey.js
```diff
@@ -0,0 +1,45 @@
...
+    message: '',
+  });
+  const [proofOfPossession, setProofOfPossession] = useState({
+    value: initPop ?? '',
+    error: false,
+    message: '',
+  });
+  const { t } = useTranslation();
+
+  const setters = {
```
soroushm: if you need a hooks for validation and keep the state its need to reusable
```javascript
const useValidation = (initValue, regex, errorMessage) => {
 const [value, setValueState] = useState({
    value: initGeneratorKey ?? '',
    error: false,
    message: '',
  });
  
const setValue = (value) => {
   const isValid = regex.test(value);
   setValueState({
      value,
      error: !isValid,
      message: isValid ? '' : errorMessage,
    })
  };
  return [value, setValue];
  };
  ```
  usage
  ```javascript
  const message = t('Please enter a valid value')
const [BLSKey, setBLSKey] = useValidation('', regex.blsPublicKeyValidator, message)
const [pubKey, setPubKey] = useValidation('', regex.generatorPublicKeyValidator, message)
const [proofOfPossession, setProofOfPossession] = useValidation('', regex.proofOfPossessionValidator, message)
  ```
is it more readable?

File: src/modules/transaction/utils/transaction.js [hunk 5]
```diff
@@ -111,6 +114,9 @@ const getElementsTxAsset = (desktopAsset, moduleAssetId) => {
     case registerDelegate: {
       return {
         username: desktopAsset.username,
```
soroushm, line 116: the form  changed to name please also change from here
```suggestion
        name: desktopAsset.name,
```

soroushm (commented):
init

src/modules/transaction/utils/transaction.js
reyraa: This is not something I choose, Lisk SDK gets to choose this parameter name.

src/modules/transaction/api/index.test.js
```diff
@@ -18,6 +22,10 @@ const {
 } = MODULE_ASSETS_NAME_ID_MAP;
 const { network } = getState();
 
+const genKey = 'dcad7c69505d549803fb6a755e81cdcb0a33ea95b6476e2585149f8a42c9c882';
+const blsKey = '830ce8c4a0b4f40b9b2bd2f16e835676b003ae28ec367432af9bfaa4d5201051786643620eff288077c1e7a8415c0285';
+const pop = '722b19e4b302e3e13ef097b417b651feadc8e28754530119911561c27b9478cdcd6b7ada331037bbda778b0b325aab5a79f34b31ea780acd01bf67d38268c43ea0ea75a5e757a76165253e1e20680c4cfd884ed63f5663c7b940e67162d5f715';
+
```
eniolam1000752: I think it would be nicer if these test data i.e `blsKey`, `genKey`, `pop` are like placed in a fixture or constants because it seems there were all referenced to  in like 6 occurrences.

eniolam1000752 (changes requested):

eniolam1000752:
@reyraa Unit tests are failing

soroushm (approved):

eniolam1000752 (approved):
LGTM

PR #4395, opened 2022-07-13
Implement adding blockchain application - Closes #4374

### LINKED ISSUE #4374: Implement blockchain application addition
### Description

1. 	As a user I visit the “blockchain application management” and click on “Add application” to input the blockchain application name or blockchain application service URL in the relevant fields.
- Extension condition: When a user visits the “Add blockchain application” page he should be prompted with a list of blockchain apps available from the registry (which are not operated by Lightcurve) for easy addition. Along with this user case still enter the exact service URL and then we can ping the service node to check if its ready for connection or show warning if it not reachable

2. “Add blockchain application” fetches the information about blockchain application from Lisk Service or provided blockchain application service URL

3. Lisk Wallet connects to blockchain application and confirms addition to the user
- Extension condition: By default Lightcurve will operate certain blockchain applications and these can be fetched via a service endpoint and list them in blockchain management, when we add any application which is not maintained by Lightcurve and the user wants to add these applications to wallet we need to inform user regarding the addition with a warning “The service node is not maintained by Lightcurve, please be aware” this is just an example we need to decide the message

### Acceptance Criteria

- Implement add flow according to defined 
- Step 1: Allow users to fetch an application by name or node URL (ping the application before api call and show status by icon)
- Enable valid url by showing an icon in search box after ping the url
- Step 2: Allow users to show application details such as Application name, avatar, owner address, application status, chainId. Last update, Last Certificate Height, deposit, application URL, custom background (user can add or back to step one)
- Step 3: Show success page and redirect user to dashboard
- Ensure [UI/UX design](https://www.figma.com/file/uStjr4eQ37wPm9bDt8JyXP/Applications-management?node-id=149%3A4536) requirements are met
- Add required unit tests
- Add required new e2e tests

	
### Additional Information

- Suggested route “application/add?chainId=XXXX”
- Suggested component name for multi step  `AddApplicationFlow`
- Suggested component name for search `AddApplicationSearch`
- Suggested component name for search `AddApplicationDetails`
- Suggested component name for search `AddApplicationSuccess`

Blocked by #4371

ikem-legend:
### What was the problem?

This PR resolves #4374

### How was it solved?

- [x] Allow users to fetch an application by name or node URL
- [x] Display application details
- [x] Show success page and redirect user to dashboard
- [x] Unit tests

### How was it tested?

 - [x] Visually
 - [x] Unit tests


src/modules/blockchainApplication/explore/api/applicationStatus.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Returns network config to use for future API calls.
+ *
+ * @param {Object} network
+ * @param {String} network.serviceUrl - a valid URL pointing to a running node
+ * @returns {Promise}
+ */
+export const getApplicationConfig = ({ serviceUrl }) =>
+  getApplicationStatus({ baseUrl: serviceUrl, network: serviceUrl })
```
eniolam1000752, line 32: isn't network suppose to be of any of these types `testnet`, `mainnet`, or `custom`

src/modules/blockchainApplication/explore/components/BlockchainApplicationAddSuccess/BlockchainApplicationAddSuccess.test.js
```diff
@@ -0,0 +1,26 @@
...
+    push: jest.fn(),
+  },
+};
+
+describe('BlockchainApplicationAddSuccess', () => {
+  it('displays properly', () => {
+    renderWithRouter(BlockchainApplicationAddSuccess, props);
+
+    expect(screen.getByText('Perfect! Application has now been added')).toBeTruthy();
+    expect(screen.getByText('You can see a list of your applications on the dashboard.')).toBeTruthy();
```
eniolam1000752, line 16: I think you can add a test to assert if the illustration image renders using `getByAltText`

eniolam1000752 (changes requested):
**UI concerns**

- padding here is 16px the UI designs says 24px
image

- heading doesn't seem to be aligned with the close button
image

- search Icon doesn't seem to be properly horizontally signed (an 8px margin-top should do the trick)
image


- there is a little offset on the highlight per row
image

- this button doesn't look like what is on the designs
image

- font-family should be `Gilroy` not `basier circle`
image

- in the add application success page, font-size and font-weight should be `15px` and `400` respectively instead we have `16px` and `300`
image

- continue button doesn't have an horizontal padding as seen in the UI designs
image

- it seems the illustration image on the add application success page looks bigger than what is depicted on the UI designs

- The back button isn't on the review application details page modal
image

- If I have added an application to my list of managed apps, I don't think I should have the ability to select this same application on the add application list modal.

- You missed out the `showHeader` props on the `All Account` table ( suggestion would been to inverse the functionality of the props which is `noHeader` meaning any use case that doesn't need the header would have the `noHeader` prop while the other tables would still maintain their default behaviour).

tests/cypress/features/blockchainExplorer.feature
```diff
@@ -17,8 +17,24 @@ Feature: BlockchainExplore
...
+    Then I should be on add blockchain application modal
+    When I click on addApplicationRow
+    And I wait 1 seconds
+    Then blockchain details should be accurately displayed
+    Then blockchain details should be in add application mode
+    When I click on addApplicationButton
+    Then I should be on add blockchain application success modal
+    When I click on addApplicationSuccessButton
+    Then I should be on dashboard page
+    Then application list should have Test App
```
eniolam1000752: I think this should be in `blockchainManagment` not `blockchainExplorer`  because its an application management flow

eniolam1000752 (changes requested):

src/modules/blockchainApplication/explore/api/applicationStatus.js
ikem-legend: @eniolam1000752 If an external application is being connected to, then network won't be any of those. The custom node for the application would most likely be different from that of the application being connected to

src/modules/blockchainApplication/explore/components/BlockchainApplicationAddSuccess/BlockchainApplicationAddSuccess.test.js
ikem-legend: @eniolam1000752 The current assertions should suffice. Similar pages don't have such assertions

tests/cypress/features/blockchainExplorer.feature
ikem-legend: @eniolam1000752 Done

eniolam1000752 (approved):
LGTM

src/theme/table/list.js
```diff
@@ -26,7 +26,13 @@ const List = ({
   if ((data.length === 0 || error) && !showHeader) return null;
   return (
     <>
-      <Header data={header} currentSort={currentSort} headerClassName={headerClassName} />
+      {showHeader ? (
```
soroushm, line 29: for new changes while its involve with changes from existing source code we can have a flag for hiding the header, its just a way for the backward compatibility! there is a thousand way to do,
I do like the below one, custom header can be empty element
``` js
customHeader ?? (<Header
          data={header}
          currentSort={currentSort}
          headerClassName={headerClassName}
        />)
```
  by the way as this change has all the requirement it's fine, just keep in mind

soroushm (approved):

PR #4433, opened 2022-08-02
Update readme for Python requirements

soroushm:
### What was the problem?
Python is a new development environment so developer needs to know 
### How was it solved?

by writing it down on the read me 

### How was it tested?
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,19 @@ If you have licensed copies of `Basier Circle` and `Gilroy`, you can add them to
 
 ### Setup environment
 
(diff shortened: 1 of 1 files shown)


soroushm:
do we need to add `.python-version` file with value of 2.7.18 to the project?

README.md
```diff
@@ -30,7 +30,19 @@ If you have licensed copies of `Basier Circle` and `Gilroy`, you can add them to
 
 ### Setup environment
 
-The development environment currently depends on [Node.js version 16 (lts/gallium)](https://nodejs.org/download/release/latest-v16.x/). The below instructions assume [nvm](https://github.com/nvm-sh/nvm) is being used to manage Node.js versions.
+The development environment currently depends on:
```
ManuGowda, line 33: Here we can directly refer https://github.com/LiskHQ/lisk-sdk/tree/development/sdk#dependencies
As this is dependency of SDK and they might resolve this later point in time, also this PR is created without any issue association

PR #4461, opened 2022-08-31
Remove delegate status patch - Closes #3910

### LINKED ISSUE #3910: Remove delegate status patch
### Description 
Since Lisk Service marks certain delegates as non-eligible, while they should be called ineligible. we are patching the status we receive from Lisk Service. Once Lisk Service [renames](https://github.com/LiskHQ/lisk-service/issues/843) non-eligible to ineligible we can remove the patch.

ikem-legend:
### What was the problem?

This PR resolves #3910

### How was it solved?

- Updating all instances of `non-eligible` to `ineligible`

### How was it tested?

- Manually


eniolam1000752 (approved):

reyraa (approved):
Looks good.

PR #4462, opened 2022-09-02
Create reusable table - Closes #4452

### LINKED ISSUE #4452: Create reusable table can work with query hooks (useInfiniteQuery, useQuery) and webSocket
### Description
we have a generic table but its need to implement logic for handling query and new data from web socket so we need to make it reusable component to avoid duplicate code

### Acceptance Criteria
- Handle load more data on press load more
- Show loading state as skeleton
- Show new data arrived on web socket
- hide load more if its doesn't have more data
- show loading on fetching next page

soroushm:
### What was the problem?

duplication on bunch of table witch they need to use react-query custom hooks - Closes #4452

### How was it solved?

 we gather all the logic to queryTable component with handle logic for load more and loading state

### How was it tested?
by unit test


src/theme/table/stickyHeader/index.js
```diff
@@ -1,34 +1,12 @@
...
   title,
-  button,
-  error,
   filters,
-  scrollToSelector,
 }) => (
   <div className={styles.header}>
     <h1>{title}</h1>
-    {// istanbul ignore next
-      button && !error && (
```
soroushm, line 15: @ikem-legend it's optional by passing the button, so we can have it for now till we remove the usage from all place

ikem-legend: @soroushm It's only used in two places `blocks` and `transactions` and we've already implemented the changes in the `blocks` domain. Once we work on the `transactions` domain, we'll update it there too

src/modules/block/components/blocks/blocks.js
```diff
@@ -122,21 +111,24 @@ const Blocks = ({
...
-            t,
           }}
         />
         <BoxContent className={`${styles.content} block-results`}>
-          <Table
+          <QueryTable
             showHeader
-            data={blocks.data}
-            isLoading={isFetching}
+            isLoading
```
ikem-legend: @soroushm @eniolam1000752 `isLoading` seems to be set to true rather than being set inside the QueryTable based on the queryHook. Therefore it should be removed

src/theme/table/index.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import Loading from './loading';
+// import Loading from './loading';
```
ikem-legend: @soroushm @eniolam1000752 Please remove comment

src/theme/QueryTable/QueryTable.js
```diff
@@ -0,0 +1,55 @@
...
+    hasNextPage,
+    hasUpdate,
+  } = queryHook(queryConfig);
+
+  const handleClick = () => {
+    // When the header is fixed at the top, the position is 50px
+    // Therefore the page should only scroll into the view if the header is not at the top
+    if (document.querySelector(`.${styles.header}`).getBoundingClientRect().top - window.scrollY <= 50) {
+      document.querySelector(scrollToSelector).scrollIntoView(true);
+    }
```
reyraa, line 30: 1. This is not the react way of doing it. You could forward the reference instead. 
2. Why hard-coded 50? Should we remember to change the JS every time CSS changes? If you use the reference, you can get the header height dynamically.

src/theme/QueryTable/QueryTable.js
```diff
@@ -0,0 +1,55 @@
...
+    // Update the table with new data
+    // button.onClick();
```
reyraa: If this updates the table with the new data, how come it is commented out? Is it not needed any more?

src/modules/block/components/blocks/blocks.js
```diff
@@ -32,34 +33,30 @@ const Blocks = ({
...
-  } = useBlocks({ config });
-
+  const [params, setParams] = useState();
   const formatters = {
     height: (value) => `${t('Height')}: ${value}`,
     /* istanbul ignore next */
     generatorAddress: (value) => `${t('Generated by')}: ${value}`,
   };
 
-  const handleLoadMore = () => {
```
reyraa, line 46: - Is this handled with another method and we have to remove it?
- Is this temporarily disabled and we have to add a todo to revert it?

src/modules/block/components/blocks/blocks.js
```diff
@@ -32,34 +33,30 @@ const Blocks = ({
...
-  const [config, setConfig] = useState({ params: {} });
-  const {
-    data: blocks, error, isFetching, fetchNextPage, hasNextPage,
-  } = useBlocks({ config });
-
+  const [params, setParams] = useState();
   const formatters = {
     height: (value) => `${t('Height')}: ${value}`,
     /* istanbul ignore next */
     generatorAddress: (value) => `${t('Generated by')}: ${value}`,
```
reyraa, line 39: I see you haven't changed this line, but filters are prone to regression, and you have changed so many things. Would be helpful to add a test to cover this line.

src/theme/table/stickyHeader/index.js
```diff
@@ -1,34 +1,12 @@
 import React from 'react';
-import LoadLatestButton from './loadLatestButton';
 import styles from './stickyHeader.css';
 
 const StickyHeader = ({
   title,
-  button,
-  error,
   filters,
```
reyraa, line 6: Since you removed everything except filters, this is stripped down to a simple `h1` wrapped in `div`. We at least can either move the `FilterBar` component here to make sense out of it. Otherwise I'd suggest to add filters as `children` so if we had to pass any other component in another page, it still makes sense.

reyraa (changes requested):

src/theme/table/stickyHeader/index.js
soroushm: @reyraa personally i'm not fan of children for it, but filter name can be rename

src/theme/QueryTable/QueryTable.js
soroushm: @reyraa I guess 50 is the height of the header, its old implementation, what is you suggestion?

src/modules/block/components/blocks/blocks.js
soroushm: @reyraa its moved to query table, then we don't need to have duplication

src/theme/QueryTable/QueryTable.js
ikem-legend: @reyraa The function for updating the table isn't available yet so I guess the `button.onClick` line should be removed since the other comment is there

src/modules/block/components/blocks/blocks.js
reyraa: then please delete it.

src/theme/QueryTable/QueryTable.js
reyraa: If it's old, we can ignore it. We don't have to introduce another refactoring.

reyraa (changes requested):

src/modules/block/components/blocks/blocks.js
ikem-legend: @reyraa TODO added for future refactor

src/modules/block/components/blocks/blocks.js
ikem-legend: @reyraa TODO added

src/modules/blockchainApplication/explore/components/BlockchainApplicationSkeleton/BlockchainApplicationSkeleton.js
```diff
@@ -23,7 +22,7 @@ const BlockchainApplicationSkeletonRow = () => (
 
 const BlockchainApplicationSkeleton = () => (
```
reyraa, line 23: We have to get rid of this at some point. But the rest look very well. Good job!

reyraa (approved):

ikem-legend (approved):

PR #4492, opened 2022-09-29
Add useDeprecatedAccount hook - Closes #4484

### LINKED ISSUE #4484: Adopt account/wallet information on redux
### Description
Since only the tx flow is using the old Wallet object in the Redux store, we can fetch and dispatch selected account data to be used by components and utilities for Creating, validating and displaying txs
We can create a hook to consume the existing hooks for retrieving account information, and then dispatch it to the Redux store immediately.

### Acceptance Criteria
- Create custom hooks to collect data from endpoint and migrate data to like what we had previously
- Retrieve and dispatch wallet data at the time of Mounting TxComposer and dispatch it
- Adopt unit Test

### Additional Information
```javascript
{
 summary: {
   address: 'lskusco5s2hx4myvdv8h....s85r',
   publicKey: 'c498ef96.......8fb978893',
   // privateKey: 'should remove'
 
   legacyAddress: '1246100....494L',
   balance: '1104980000',  // Create this value at the time of data retrieval
   username: '',  // Create this value at the time of data retrieval
   isMigrated: true, // Create this value at the time of data retrieval
   isDelegate: false, // Create this value at the time of data retrieval
   isMultisignature: false, // Create this value at the time of data retrieval
 },
 legacy: {
   address: '',
   balance: '73723736'
 },
 // balances
 token: {
   balance: '1104980000'
 },
 // auth
 sequence: {
   nonce: '28'
 },
 // auth
 keys: {
   numberOfSignatures: 0,
   mandatoryKeys: [],
   optionalKeys: []
 },
 dpos: {
   // dpos -> specific /v3/dpos/delegates. We can use useDelegates hook
   delegate: {
     username: '',
     consecutiveMissedBlocks: 0,
     lastForgedHeight: 14075260,
     isBanned: false,
     totalVotesReceived: '0'
   },
   // dpos -> specific /v3/dpos/votes/sent
   sentVotes: [
     {
       delegateAddress: 'lsk2a2pb5yp267nudcajund7ksepwnubptp4fu429',
       amount: '1000000000',
       // name (optional)
     }
   ],
   // dpos -> /v3/dpos/unlocks
   unlocking: [
     {
       delegateAddress: 'lsknfggfq34u78vmrts7hdhtk57ocw7e96yp8nzh4',
       amount: '1000000000',
       height: {
         start: 16569388,
         end: 16571388
       }
     }
   ]
 }
}

```

soroushm:
### What was the problem?

This PR resolves #4484 

### How was it solved?
by adding a hook to create data structure for deprecated account

### How was it tested?
unit test


src/modules/account/__fixtures__/mockLegacy.js
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line import/prefer-default-export
+export const mockLegacy = {
+  data: {
+    legacyAddress: '3057001998458191401L',
+    balance: '0',
```
ManuGowda: if your intention is to migrate legacy account and reclaim? then would be better to have some balance to initiate the reclaim?

src/modules/account/hooks/queries/useLegacy.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param {object} configuration - the custom query configuration object
+ * @param {Object} configuration.config - the query config
+ * @param {Object} configuration.config.params - the query config params
+ * @param {string} configuration.config.params.publicKey - auth address
+ * @param {string} configuration.options - the query options
+ *
+ * @returns the query object
+ */
+// eslint-disable-next-line import/prefer-default-export
```
ManuGowda: why dont we update the eslint?

src/modules/account/hooks/queries/useLegacy.js
```diff
@@ -0,0 +1,33 @@
+import { LEGACY } from 'src/const/queries';
+import {
+  API_VERSION,
+} from 'src/const/config';
+import { useCustomQuery } from 'src/modules/common/hooks';
+
+/**
+ * Creates a custom hook for command parameters schemas queries
```
ManuGowda: I think the general comment doesn't make so much sense? if its generic it can be part of `useCustomQuery`, here the comment can be more specific to `useLegacy`?

src/modules/account/hooks/useDeprecatedAccount.js
```diff
@@ -0,0 +1,165 @@
...
+  } = useAuth({ config: { params: { address } } });
+  useEffect(() => {
+    if (isAuthSuccess) {
+      setAccount((state) => ({
+        ...state,
+        summary: {
+          ...state.summary,
+          publicKey: auth?.meta?.publicKey || pubkey,
+          username: auth?.meta?.name || '',
+          isMultisignature: (auth?.data?.mandatoryKeys || []).length > 1,
```
ManuGowda: This logic is not valid, one can setup a multisignature account with only `optionalKeys`, instead of this you can use property `numberOfSignatures`?

src/modules/account/__fixtures__/mockLegacy.js
soroushm: it can be work, I'm try to cover happy path, for later we can use `mswjs/data`, it's  kind a database to keep the state also we can generate dyta dynamically or conditionally

src/modules/account/hooks/queries/useLegacy.js
soroushm: good point, I will make it on the difrent pr

ManuGowda: Ok please create an issue so that we dont forget 🙏🏼

src/modules/account/hooks/useDeprecatedAccount.js
```diff
@@ -0,0 +1,178 @@
...
+export const useDeprecatedAccount = () => {
+  const [currentAccount] = useCurrentAccount();
+  const dispatch = useDispatch();
+  const { pubkey, address } = currentAccount.metadata;
+  const [account, setAccount] = useState({
+    summary: {
+      address,
+      publicKey: '',
+      legacyAddress: '',
+      // balance: @todo account has multiple balance now
```
ManuGowda, line 20: why is this todo? we can store the balances in the state and later we can adjust the implementation?

src/modules/legacy/hooks/queries/index.js
```diff
@@ -0,0 +1 @@
+export * from './useLegacy'
```
ManuGowda, line 1: `useLegacy` file doesn't exits 🤔

ManuGowda (changes requested):

src/modules/account/hooks/useDeprecatedAccount.js
soroushm: @ManuGowda then witch blanace we need to store here? i think we need adjust implementation other than put some random balance here

ManuGowda: maybe @reyraa can make better judgement here.

reyraa (approved):

ManuGowda (approved):

PR #4493, opened 2022-10-03
Switch between supported networks - Closes #4477

### LINKED ISSUE #4477: Switch between supported networks
### Description

Users should be able to switch between supported networks (mainnte, testnet, betanet, custom-network-url, etc..) for a given chain (application). 

Switching between supported network can be achieved with application meta data endpoint information.

### Acceptance Criteria

- Allow user to switch between  supported networks (mainnte, testnet, betanet, custom-network-url, etc..)
- Change application base URL

Balanced02:
### What was the problem?

This PR resolves #4477 

### How was it solved?
- [x] Allow user to switch between supported networks (mainnet, testnet)
- [x] Change application base URL to include query with `network=networkType`

### How was it tested?
- [x] Unit test for redux change
- [x] Browser


src/modules/common/components/bars/topBar/networkName.js
```diff
@@ -1,42 +1,45 @@
...
-            }
-          </p>
-        </Tooltip>
-      </div>
-    </section>
+    <>
+      <section className={styles.wrapper} onClick={() => setShowDropdown(true)} >
+        <span className={`${styles.status} ${statusColor}`} />
+        <div className={styles.message}>
+          <span >{activeNetwork}</span>
```
ikem-legend: @Balanced02 Please remove space before span closing tag (`<span >`) and other similar cases (` >`)

src/modules/common/components/bars/topBar/networkName.js
```diff
@@ -1,42 +1,45 @@
...
   const statusColor = network.status.online ? styles.online : styles.offline;
 
+  const closeDropdown = () => {
+    setShowDropdown(false)
+  }
+
+  const setActiveNetwork = net => {
+    dispatch(networkSelected({ name: net }))
+    closeDropdown();
+  }
```
eniolam1000752:
```suggestion
  const setActiveNetwork = name => {
    dispatch(networkSelected({ name }))
    closeDropdown();
  }
```

src/modules/common/components/bars/topBar/networkName.js
```diff
@@ -1,42 +1,45 @@
...
-  const networksList = {
-    [networkKeys.mainNet]: t('Mainnet').toLowerCase(),
-    [networkKeys.testNet]: t('Testnet').toLowerCase(),
-    [networkKeys.customNode]: t('Custom node').toLowerCase(),
-  };
+const Network = ({ t }) => {
+  const [showDropdown, setShowDropdown] = useState(false)
+  const network = useSelector(state => state.network);
+  const dispatch = useDispatch()
   const activeNetwork = getNetworkName(network);
```
eniolam1000752:
```suggestion
  const activeNetworkName = getNetworkName(network);
```

I think using `activeNetworkName` would be nicer so that anywhere we use it, we know we are using the network's name (which is of type string)

src/modules/common/components/bars/topBar/networkName.js
```diff
@@ -1,42 +1,45 @@
...
+    <>
+      <section className={styles.wrapper} onClick={() => setShowDropdown(true)} >
+        <span className={`${styles.status} ${statusColor}`} />
+        <div className={styles.message}>
+          <span >{activeNetwork}</span>
+          <Icon name="dropdownArrowIcon" />
+        </div>
+      </section>
+      <Dropdown showDropdown={showDropdown} active={activeNetworkIndex} className={styles.dropdown} closeDropdown={closeDropdown} title={t('Select network')} >
+        {Object.values(networkKeys).map(net => <button key={net} className={styles.networkItem} onClick={() => setActiveNetwork(net)} >{t(net)}</button>)}
```
eniolam1000752:
```suggestion
        {Object.values(networkKeys).map(networkName => <button key={networkName} className={styles.networkItem} onClick={() => setActiveNetwork(networkName)} >{t(networkName)}</button>)}
```

src/modules/common/components/bars/topBar/topBar.js
```diff
@@ -16,7 +16,7 @@ import styles from './topBar.css';
 import Network from './networkName';
 import NavigationButtons from './navigationButtons';
 
-const TopBar = ({ t, history, network, token, noOfVotes, location }) => {
+const TopBar = ({ t, history, noOfVotes, location }) => {
```
eniolam1000752: you can remove t from prop and consume it like this

`const { t } = useTranslation();`

src/modules/common/hooks/useCustomInfiniteQuery.js
```diff
@@ -29,13 +31,17 @@ export const useCustomInfiniteQuery = ({
...
+  const networkName = getNetworkName(network)
+
   return useInfiniteQuery(
     [chainID, config, APPLICATION, METHOD, ...keys],
     async ({ pageParam }) => client[METHOD]({
       ...config,
       params: {
         ...(config.params || {}),
         ...pageParam,
+        network: networkName
```
eniolam1000752, line 44: @soroushm @Balanced02 this is giving me concerns, because now, I am thinking, how are we going to handle the scenario if i want to switch to devnet. Which is a node I am running locally(meaning I can reach my node on localhost or 127.0.0.1). because we are passing the network name as a parameter. Are we expecting the service's server we are hitting to redirect the call based on the network params been passed. 

I think I need more clarification here

src/modules/common/hooks/useCustomQuery.js
```diff
@@ -28,9 +30,17 @@ export const useCustomQuery = ({
   client = defaultClient
 }) => {
   const [{ chainID }] = useCurrentApplication();
+  const network = useSelector(state => state.network);
+  const networkName = getNetworkName(network)
+
+  const axiosConfig = {
+    ...config,
+    params: { ...(config?.params ?? {}), network: networkName }
```
eniolam1000752: I think instead of doing this you can make `config` on line 28 this `config = {}` then here you can do this

params: { ...config.params , network: networkName }

eniolam1000752 (changes requested):

src/modules/common/hooks/useCustomInfiniteQuery.js
Balanced02: From the explanation from @soroushm, we need to pass the query of network being either mainnet or testnet to the api calls. So, calls would have the query of `network=mainnet` or otherwise.

ikem-legend (dismissed):

eniolam1000752 (approved):

eniolam1000752 (changes requested):
you've got failing unit tests you need to resolve @Balanced02

eniolam1000752 (changes requested):
You still have some coverage issues to resolve

image

setup/jest.config.js
```diff
@@ -305,6 +305,8 @@ module.exports = {
     'src/modules/account/hooks/useCurrentAccount.js',
     'src/modules/transaction/store/actions.js',
     'src/modules/transaction/hooks/useCreateTransaction.js',
+    'src/modules/common/hooks/useOutsideClick.js',
+    'src/modules/common/hooks/useCustomQuery.js',
```
eniolam1000752: was there a reason for excluding coverage checks for `useOutsideClick` and `useCustomQuery` @Balanced02

setup/jest.config.js
```diff
@@ -307,6 +307,7 @@ module.exports = {
     'src/modules/transaction/hooks/useCreateTransaction.js',
     'src/modules/common/hooks/useOutsideClick.js',
     'src/modules/common/hooks/useCustomQuery.js',
+    'src/theme/Dropdown/dropdown.js',
```
eniolam1000752: Also why excluding it here @Balanced02

eniolam1000752 (changes requested):

setup/jest.config.js
Balanced02: > Also why excluding it here @Balanced02

Added more tests but couldn't get it above the threshold.

ikem-legend (dismissed):

eniolam1000752 (approved):

ikem-legend (approved):

PR #4540, opened 2022-10-25
Fix for lisk-client Require undefined issues  - Closes #4520

### LINKED ISSUE #4520: Application fails to load modules defined in lisk-api-client
### Expected behavior
after @liskhq/lisk-client released version [6.0.0-alpha.32064](https://npm.lisk.com/-/web/detail/@liskhq/lisk-api-client/v/6.0.0-alpha.32064) wallet desktop complain about some of the packages is not exist in web browser, web-pack/babel should use browser export instead of node version

### Actual behavior
```js
index.js:6 Uncaught ReferenceError: require is not defined
    at ./node_modules/node-gyp-build/index.js (index.js:6:22)
    at __webpack_require__ (bootstrap:21:1)
    at fn (hot module replacement:61:1)
    at ./node_modules/zeromq/lib/native.js (native.js:6:18)
    at __webpack_require__ (bootstrap:21:1)
    at fn (hot module replacement:61:1)
    at ./node_modules/zeromq/lib/index.js (index.js:3:16)
    at __webpack_require__ (bootstrap:21:1)
    at fn (hot module replacement:61:1)
    at ./node_modules/@liskhq/lisk-api-client/dist-node/ipc_channel.js (ipc_channel.js:7:18)
```

### Steps to reproduce
NPM install latest version of SDK 

Version 3.0.0

soroushm:
### What was the problem?

This PR resolves #4520
### How was it solved?
By change import to the web version

### How was it tested?
unit test


src/locales/en/common.json
```diff
@@ -336,6 +336,7 @@
   "New blocks": "New blocks",
   "New bookmark": "New bookmark",
   "New transactions": "New transactions",
+  "New wallet address": "New wallet address",
```
eniolam1000752: This seems weird this text was added by i18n even though there are no changes to the JSX in this PR

src/locales/en/common.json
```diff
@@ -361,7 +362,6 @@
   "Number of transactions / Volume (LSK)": "Number of transactions / Volume (LSK)",
   "Official guidelines": "Official guidelines",
   "Ok": "Ok",
-  "Old account": "Old account",
```
eniolam1000752: like wise this, it seems weird this text was removed by i18n even though there are no changes to the JSX in this PR

eniolam1000752 (changes requested):

src/locales/en/common.json
soroushm: so how changed it they didn't addet to the common js there is transaltion for it

src/locales/en/common.json
soroushm: its not exist in project

eniolam1000752 (approved):

clemente-xyz:
@soroushm would be nice also to include in the PR description the command you ran for generating the patching of `lisk-client`. In the case of mobile (where we use `yarn`), we had to run:

```
yarn --registry https://npm.lisk.com  patch-package @liskhq/lisk-client --use-yarn --include package\\.json$ --exclude
```

I say this cause there is an issue when patching `package.json` files, since by default they are excluded. Also, we need to fetch the package from Lisk registry, for which we need to specify it on the command.

PR #4541, opened 2022-10-26
Implement Dpos constants query hook - Closes #4502

### LINKED ISSUE #4502: Integrate tokenID and other related dpos constants to be used for dpos governance
### Description

In the delegate feature, there is a need to know the token and some other dpos related constants that would be used for dpos governance on a particular chain since a user can have more than one token balance for that given chain.

### Motivation

Service would be exposing an endpoint which would be able to return these constants(this would be addressed in this [service 's issue](https://github.com/LiskHQ/lisk-service/issues/1281)

### Additional Information
- Create a `useDposConstants` query hook to get dpos constants
- Consume `useDposConstants`  to get the `tokenIDDPoS` and `roundLength` and integrate into the edit vote feature
- Implement unit test for `useDposConstants`.

eniolam1000752:
### What was the problem?

This PR resolves #4502

### How was it solved?

- [x] Implement useDposConstants query hook
- [x] implement MSW handler for the dpos constants query
- [x] implement unit tests for useDposConstants
- [x] replace dposTokenId with actual api call using useDposConstants query hook
- [x] fix filling unit tests.

### How was it tested?

visually
jest


src/modules/dpos/validator/components/VoteForm/voteForm.test.js
```diff
@@ -10,8 +10,11 @@ import flushPromises from '@tests/unit-test-utils/flushPromises';
 import { useTokensBalance } from '@token/fungible/hooks/queries';
 import VoteRow from './VoteRow';
 import Form from './VoteForm';
+import { useDposConstants } from '../../hooks/queries';
+import { mockDposConstants } from '../../__fixtures__/mockDposConstants';
 
 jest.mock('@token/fungible/hooks/queries');
+jest.mock('../../hooks/queries');
```
ikem-legend: @eniolam1000752 Avoid mocking entire folders to prevent unexpected issues

ikem-legend (approved):

PR #4547, opened 2022-10-31
Update token transfer form

### LINKED ISSUE #4528: Adopt transfer token form
### Description
Transfer token form needs to adopt new changes

### Acceptance Criteria
- From application shouldn't be changed while submitting the transaction
- Recipient application should fetch from `/api/v3/blockchain/apps`. It also needs merge application data from metadata
- Tokens list should refetch from recipient application and make sure that while switching application it's fetching again
- Loading state should be handled while its fetching application or token list
- Make sure transaction object will be updated with application and token switch
- Fill the form by query params from deep link
- Adopt unit and e2e test

ikem-legend:
### What was the problem?

This PR resolves #4528

### How was it solved?

- Updating the UI based on the new design

### How was it tested?

- [x] Manually


src/modules/token/fungible/components/SendForm/SendForm.js
```diff
@@ -127,7 +132,7 @@ const SendForm = (props) => {
         address: recipient.value,
         title: recipient.title,
       },
-      token,
+      token: recipientChain,
```
eniolam1000752: this seems weird @ikem-legend

src/modules/token/fungible/components/SendForm/SendForm.js
```diff
@@ -199,22 +213,25 @@ const SendForm = (props) => {
               <span className={styles.balance}>
                 Balance:&nbsp;&nbsp;
                 <span>
-                  <TokenAmount val={amount} />
+                  <TokenAmount val={balance} />
                   {token?.symbol}
```
eniolam1000752: balance should only show when a token has been selected

eniolam1000752 (changes requested):

eniolam1000752:
unit test needs to be fixed too

src/modules/token/fungible/components/SendForm/SendForm.js
ikem-legend: @eniolam1000752 One of the utils for converting desktop transactions to element transactions breaks if token isn't part of the raw transaction params, hence this change

src/modules/token/fungible/components/SendForm/SendForm.js
ikem-legend: @eniolam1000752 Balance is currently blank until a token is selected

src/modules/token/fungible/components/SendForm/SendForm.js
eniolam1000752: hmmm so what I mean is that the key `token` and the value `receipientChain` don't refer to the same thing

eniolam1000752 (dismissed):

sridharmeganathan (dismissed):

eniolam1000752 (approved):

PR #4562, opened 2022-11-04
Migrate deprecated http function to client (api call)

### LINKED ISSUE #4524: Integrate api call with client.js, http.js handler is deprecated
### Description
Lisk wallet now support multiple application, we introduce Client Singleton class to handle api call in rest or rpc depend of the config, we used it most the time in react query, but some of the utils calling the api directly with http.js  they need to integrate with new client

### Acceptance Criteria
- Remove the http.js file
- Update utiles with client
- Adopt Unit Tests

soroushm:
### What was the problem?

This PR resolves #4524 

### How was it solved?
its in progress

### How was it tested?
Unit test


src/modules/blockchainApplication/manage/components/AddApplicationList/index.js
```diff
@@ -1,39 +1,10 @@
...
-      data: mockBlockchainApplications,
-    }),
-    getApiParams: (state) => ({
-      network: state.network,
-    }),
-    transformResponse: response => response.data,
-    autoload: false,
-  },
-};
-
```
ikem-legend, line 35: @soroushm The API util here was removed but the child component is still expecting the props removed. The child component can either be completely refactored too or a TODO should be added here for that

src/modules/transaction/api/index.test.js
```diff
@@ -11,10 +11,8 @@ import accounts from '@tests/constants/wallets';
 import { genKey, blsKey, pop } from '@tests/constants/keys';
 import { mockAppTokens } from '@tests/fixtures/token';
 import {
-  getTransaction,
```
ikem-legend, line 14: @soroushm `getTransaction` was removed from the test but the util in the index file wasn't removed

src/modules/transaction/manager/blockDetailsTransactionManger.js
```diff
@@ -1,21 +1,8 @@
 /* istanbul ignore file */
 import { compose } from 'redux';
 import { withTranslation } from 'react-i18next';
-import withData from 'src/utils/withData';
 import BlockDetailsTransactions from '../components/BlockDetailsTransactions/BlockDetailsTransactions';
-import { getTransactions } from '../api';
 
 export default compose(
```
ikem-legend: @soroushm If `withTranslation` is the only HOC used, then there's no need for `compose`

ikem-legend (commented):
Just a few minor fixes and it's good to go

src/modules/transaction/api/index.test.js
soroushm: Done

src/modules/blockchainApplication/manage/components/AddApplicationList/index.js
soroushm: oh, missed that its need to use react-query

src/modules/transaction/manager/blockDetailsTransactionManger.js
eniolam1000752: We don't even need translation here

src/modules/transaction/manager/recentTransactionManager.js
```diff
@@ -1,22 +1,5 @@
...
-      return {
-        address,
-        network: state.network,
-      };
-    },
-    defaultData: [],
-    transformResponse: response => response.data.splice(0, 5),
-  },
-})(withTranslation()(RecentTransaction));
+export default withTranslation()(RecentTransaction);
```
eniolam1000752, line 5: We don't need translation here we can use 
`useTranslation` in the RecentTransaction component

eniolam1000752 (changes requested):

src/modules/transaction/manager/recentTransactionManager.js
soroushm: yeah, but its just out of scope of this pr, we have a different task to fix this transaction recent

eniolam1000752 (approved):

PR #4584, opened 2022-11-15
Refactored delegate name check with react-query

### LINKED ISSUE #4561: Adopt useDelegateName with react-query
### Description
In register delegate form, we need to validate uniqueness of delegate name, so that the application uses `useDelegateName` hooks in utils and checkUsername function to check the delegate name. We have an API call `getDelegate` which is using old methodology and calling account endpoint which does not exist any more so we need to adopt useDelegateName with react-query

### Acceptance Criteria
- Adopt useDelegateName with react-query
- Remove unused utiles
- Adopt unit test

ikem-legend:
### What was the problem?

This PR resolves #4561

### How was it solved?

Using react-query for delegate name check

### How was it tested?

- Manually


src/modules/dpos/validator/hooks/useDelegateName.js
```diff
@@ -30,39 +29,44 @@ const useDelegateName = (value) => {
...
-    clearTimeout(timeout.current);
+  const [params, setParams] = useState({});
+  const { data, isLoading, error } = useDelegates({
+    config: { params },
+    options: { enabled: name.value?.length >= 3, retry: false },
+  });
 
-    timeout.current = setTimeout(() => {
+  const checkUsername = async () => {
+    if (name.value.length >= 3 && name.value.length <= 20 && isLoading) {
```
reyraa: We are hard-coding `3` and `20` in several places. Would be better if we stored them in constants.  Just an optional recommendation.

src/modules/dpos/validator/hooks/useDelegateName.js
```diff
@@ -30,39 +29,44 @@ const useDelegateName = (value) => {
...
   };
 
   useEffect(() => {
+    clearTimeout(timeout.current);
     const formatError = validate(name.value, t);
     if (!formatError) {
-      checkUsername();
+      timeout.current = setTimeout(() => {
+        setParams({ name: name.value });
+      }, 500);
```
reyraa: Seems as though we have decreased the debounce time from 1sec to 500 ms. Was it intentional?

src/modules/dpos/validator/hooks/useDelegateName.test.js
```diff
@@ -46,13 +52,15 @@ describe('useDelegateName', () => {
   });
 
   it('should mark as error if the value is not unique', async () => {
-    delegateAPI.getDelegate.mockResolvedValue({
-      data: [{}],
-      meta: {},
+    // delegateAPI.getDelegate.mockResolvedValue({
+    //   data: [{}],
+    //   meta: {},
+    // });
```
reyraa: Pleas remove this if not needed.

reyraa (dismissed):
Looks good. Although I left a few comments. You can merge it as is, but I appreciate if you fixed them first.

src/modules/dpos/validator/hooks/useDelegateName.js
ikem-legend: @reyraa Fixed

src/modules/dpos/validator/hooks/useDelegateName.js
ikem-legend: @reyraa It wasn't. Fixed

src/modules/dpos/validator/hooks/useDelegateName.test.js
ikem-legend: @reyraa Removed

oskarleonard:
Im getting some strange behaviour when testing this locally. I get 3 different error states for the same string

image
image
image

Also,
It always flashes (for a second) an error toast that the username is taken even thought it isn't.

oskarleonard (approved):

reyraa (approved):

PR #4589, opened 2022-11-21
Fix remove account flow

### LINKED ISSUE #4535: Adjust remove account flow
### Expected behavior
1- On remove an account if the current account is same account, user needs to be redirect to account management and current account should be erased
2- Cancel button on remove account modal doesn't work

### Actual behavior
1- account will be remove but user still using the same account,
2- modal should be close

### Steps to reproduce
- Add an account
- Select that account
- Remove the same account you're in

v3.0.0

oskarleonard:
### What was the problem?

This PR resolves #4535

### How was it solved?

- By removing current account from the redux store when deleting an account. 
- Merging two different UI for performing the same thing

### How was it tested?

Manually


src/modules/account/components/RemoveAccount/RemoveAccount.test.js
```diff
@@ -19,12 +19,12 @@ jest.mock('@account/hooks', () => ({
   )),
 }));
 
-describe('Remove account', () => {
+describe('Remove account ', () => {
```
ikem-legend: @oskarleonard Please remove empty space

src/modules/account/components/AccountRow/AccountRow.test.js
```diff
@@ -8,7 +8,7 @@ jest.mock('react-i18next');
 
 const props = {
   account: mockSavedAccounts[0],
-  showRemove: true,
+  onRemove: jest.fn(),
```
ikem-legend: @oskarleonard Use a constant for the onRemove mock.
For example, `const mockOnRemove = jest.fn() and then reuse it here and while rendering the component below
```suggestion
  onRemove: mockOnRemove,
```

src/modules/account/components/AccountRow/AccountRow.test.js
```diff
@@ -27,9 +27,9 @@ describe('Select Account Row', () => {
     expect(container.getByText(truncateAddress(mockSavedAccounts[0].metadata.address))).toBeTruthy();
   });
 
-  it('Should  list', async () => {
+  it('Should show delete button', async () => {
     const container = render(<AccountRow {...props} />);
-    container.rerender(<AccountRow {...props} showRemove />);
+    container.rerender(<AccountRow {...props} onRemove={jest.fn()} />);
```
ikem-legend: @oskarleonard 
```suggestion
    container.rerender(<AccountRow {...props} onRemove={mockOnRemove} />);
```

ikem-legend (approved):
Looks good to me

soroushm (approved):
LGTM

PR #4611, opened 2022-11-25
Updated fee logic

### LINKED ISSUE #4529: Adopt transaction fees based on native token
### Description
Each application have a native token and transaction fee should calculate based on the current application native token. Also it seems the calculation of transaction fees is wrong especially for a `register Multi-signature` transaction.

### Acceptance Criteria
- Get the native token id on constant api `/api/v3/dpos/constants`
- Validate fee balance based on the native token
- Fix fee calculation across all types of transactions.
- Adopt Unit test

ikem-legend:
### What was the problem?

This PR resolves #4529

### How was it solved?

By updating the fee calculation logic

### How was it tested?

- Manually


src/modules/transaction/api/index.js
```diff
@@ -210,23 +209,27 @@ export const getTransactionFee = async ({
...
-    baseFees,
-    numberOfSignatures,
-    numberOfEmptySignatures,
-  });
+  const minFee = transactions.computeMinFee(
+    {
+      ...transactionObject,
+      params: {
+        ...transactionObject.params,
+        signatures: new Array(numberOfSignatures).fill(0).map(() => Buffer.alloc(64)),
```
eniolam1000752: Also, I think this should not be done this way only empty signatures should be replaced with an empty buffer, this is because when members are signing, there could be empty signatures(meaning a member is yet to sign) and there could be actual signatures(meaning a member has signed). so in this case I think it can be done like this @ikem-legend 

```suggestion
        ...(numberOfSignatures && !transactionObject.params. signatures.length && { signatures: Array(numberOfSignatures).fill(Buffer.alloc(64)) }) ,
```

eniolam1000752 (changes requested):

src/modules/transaction/api/index.js
ikem-legend: @eniolam1000752 Fixed

eniolam1000752 (dismissed):

eniolam1000752 (approved):

soroushm (approved):
LGTM

Issue #4605, opened 2022-11-23
Sign message flow does not work


eniolam1000752:
### Expected behavior
A user should be able to sign a clear text string by providing a text and getting a signed result.

### Actual behavior
Presently the flow does not present any signed result after providing the text

### Steps to reproduce
- Click on Sign message on the side menu
- provide a text in the input field and click on continue
- A distorted UI is presented

### Which version(s) does this affect? (Environment, OS, etc...)
All versions

PR #4614, opened 2022-11-28
Resolve issue with register multi-signature and sign message

### LINKED ISSUE #4606: Only the sender of a register multi-sig transaction can sign the transaction
### Expected behavior
To sign a multi-signature transaction, all members of the multi-signature group needs to sign the transaction only then would the initiator of the transaction be able sign and send the transaction to the network.

### Actual behavior
A member that has finally signed a transaction whom isn't the initiator of the transaction is able to send the transaction to the network.

### Steps to reproduce
- Create a multi-signature transaction object as a sender
- Members should sign the transaction
- The last member sees a `Send` button (which is wrong)

eniolam1000752:
### What was the problem?

This PR resolves #4606 and #4605

### How was it solved?

- [x] Refactor the boolean logic that generates multi-signature status
- [x] Refactor flow on register group account to accommodate the sender as a member of the multi sig group
- [x] Fix sign message flow
- [x] Fix verify message flow
- [x] Fix failing unit tests

### How was it tested?

- jest
- visually


src/modules/auth/components/EnterPasswordForm/EnterPasswordForm.js
```diff
@@ -27,10 +24,9 @@ const EnterPasswordForm = ({ onEnterPasswordSuccess, title, encryptedAccount, is
     formState: { errors },
   } = useForm();
   const apiError = errors[API_ERROR_NAME];
-  const activeAccount = useSelector(selectActiveTokenAccount);
   const { getAccountByAddress } = useAccounts();
   const [currentAccount] = useCurrentAccount();
-  const requestedAccount = getAccountByAddress(activeAccount.summary?.address);
+  const requestedAccount = getAccountByAddress(currentAccount.metadata.address);
```
oskarleonard: Why this change?  This component is used when adding an account by recovery file. Meaning that currentAccount can be empty for that usecase. Removing the `?` will make it crash, needs to use `currentAccount.metadata?.address` or more refactor needed

soroushm: so we have the currentAccount in the next line i'm guess this logic was for the day we still had an account in redux its should removed now its just duplicated

soroushm:
I've got error during the sign messages
```Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'filter')
    at getNonEmptySignatures (helpers.js:17:33)
    at getTransactionSignatureStatus (helpers.js:61:25)
    at _callee4$ (actions.js:200:49)
    at tryCatch (actions.js:2:1)
    at Generator.<anonymous> (actions.js:2:1)
    at Generator.next (actions.js:2:1)
    at asyncGeneratorStep (actions.js:2:1)
    at _next (actions.js:2:1)
    at actions.js:2:1
    at new Promise (<anonymous>)```

**Review by soroushm — CHANGES_REQUESTED** [review]
I wasn't able to sign multisignature while I add the public key from current account to the list

File: src/modules/auth/components/EnterPasswordForm/EnterPasswordForm.js (hunk 1 above)
Comment (oskarleonard) [reply src/modules/auth/components/EnterPasswordForm/EnterPasswordForm.js]: I see so just a removal of "dead" code. Then we still need optional chaining on the `currentAccount` to not break the "add an account by recovery file" function.

In the add an account by recovery file we pass in the `encryptedAccount` prop to this component and the whole `currentAccount` logic is dead in that case. 

It would be better if this component always took in the correct account and avoided this logic ->
```
  const account = useMemo(
    () => encryptedAccount || requestedAccount || currentAccount,
    [currentAccount]
  );
```
to make it more understandable. But it might be too big of a refactor right now?

eniolam1000752: yhea so the idea is that currentAccount would always have metadata so far the user is logged and since for any use case that a password is to be provided the user is logged in, `currentAccount.metadata` would never be empty.

oskarleonard: @eniolam1000752 
This `EnterPasswordForm` can currently be used when a user is not completely loged in, for example the "Add an account by recovery file" flow and will in that case be empty. 

The "Add an account by recovery file" flow looks like this:

1. Open a new private window
2. Go to http://0.0.0.0:8080/#/account/add/by-file
3. Upload a file and continue
4. Now `EnterPasswordForm` is used even though you are not loged in. And to prevent a crash here we need to have the optional chaining.

eniolam1000752: hmmmm 🤔 true

oskarleonard (approved):

ManuGowda:
@eniolam1000752 two issue needs to be resolved

- Downloading multisignature file has bug the file name should contain a transaction ID, its currently appears as `tx-.json`
- Once the account becomes multisignature I am unable to view the information, here is missing implementation [icon](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=38%3A13849&t=CpuFKV5SeBIImhbI-4) and [details](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=38%3A13923&t=CpuFKV5SeBIImhbI-4)

ManuGowda (changes requested):
@eniolam1000752 issue needs to be resolved

- Unable to transfer token after upgrading to multisignature
- Downloading multisignature file has bug the file name should contain a transaction ID, its currently appears as `tx-.json`
- Once the account becomes multisignature I am unable to view the information, here is missing implementation [icon](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=38%3A13849&t=CpuFKV5SeBIImhbI-4) and [details](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=38%3A13923&t=CpuFKV5SeBIImhbI-4)

ManuGowda (approved):
Created issues for other reported issues.

PR #4657, opened 2022-12-08
Discrete mode is not respected everywhere

### LINKED ISSUE #4620: Discrete mode is not respected in the wallet view
### Expected behavior
All the balances should be blurred out when discrete mode is enabled

### Actual behavior
image
image

### Steps to reproduce
- http://0.0.0.0:8080/#/wallet
- http://0.0.0.0:8080/#/wallet/tokens/all

v3

oskarleonard:
### What was the problem?

This PR resolves #4620

### How was it solved?

- Make sensitive components discrete by default

### How was it tested?

Manually

### DIFF
--- a/src/modules/blockchainApplication/explore/components/BlockchainApplicationStatistics/index.js
+++ b/src/modules/blockchainApplication/explore/components/BlockchainApplicationStatistics/index.js
@@ -79,7 +79,7 @@ const BlockchainApplicationStatistics = () => {
               </Tooltip>
             </div>
             <p className={`${styles.statsInfo} stats-info-value`}>
-              <TokenAmount token="LSK" val={amount} />
+              <TokenAmount Wrapper={React.Fragment} token="LSK" val={amount} />
             </p>
           </div>
           <div>
--- a/src/modules/common/components/div/Div.js
+++ b/src/modules/common/components/div/Div.js
@@ -1,0 +1,7 @@
+import React from 'react';
+
+function Div({ className, children }) {
+  return <div className={className}>{children}</div>;
+}
+
+export default Div;
--- a/src/modules/blockchainApplication/explore/components/BlockchainApplicationRow/BlockchainApplicationRow.test.js
+++ b/src/modules/blockchainApplication/explore/components/BlockchainApplicationRow/BlockchainApplicationRow.test.js
@@ -20,6 +20,11 @@ usePinBlockchainApplication.mockReturnValue({
   checkPinByChainId: jest.fn().mockReturnValue(true),
 });
 
+jest.mock('src/utils/searchParams', () => ({
+  selectSearchParamValue: jest.fn(),
+  addSearchParamsToUrl: jest.fn(),
+}));
+
 describe('BlockchainApplicationRow', () => {
   let wrapper;
   const props = {
--- a/src/modules/common/components/converter/converter.js
+++ b/src/modules/common/components/converter/converter.js
@@ -2,6 +2,7 @@ import React from 'react';
 import numeral from 'numeral';
 import 'numeral/locales';
 import i18n from 'src/utils/i18n/i18n';
+import DiscreetMode from '@common/components/discreetMode';
 import styles from './converter.css';
 
 const Converter = ({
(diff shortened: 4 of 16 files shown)


src/modules/blockchainApplication/explore/components/BlockchainApplicationStatistics/index.js
```diff
@@ -79,7 +79,7 @@ const BlockchainApplicationStatistics = () => {
               </Tooltip>
             </div>
             <p className={`${styles.statsInfo} stats-info-value`}>
-              <TokenAmount token="LSK" val={amount} />
+              <TokenAmount Wrapper={React.Fragment} token="LSK" val={amount} />
```
ManuGowda, line 82:
```suggestion
              <TokenAmount wrapper={React.Fragment} token="LSK" val={amount} />
```
lets ensure the casing is followed, if you had specific intention to change maybe please propose 🙏🏼

src/modules/common/components/div/Div.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
ManuGowda, line 1: same here with file naming `Div.js` the source code might get polluted with different naming patterns, please change it to current format and if you have suggestions please propose them as improvements 🙏🏼

oskarleonard: You mean folder (should be uppercase) and not the component, right? 

Btw, which is the correct pattern? I see very different patterns used, like in the files i work in now. 
image

Sometimes we expose a "private" component (uppercase) with index, sometimes folders only have index (lowercase) and sometimes components in folders are lowercase. The consistent thing are that component folders start with uppercase.

Or am i missing some pattern here 🤔

src/modules/blockchainApplication/explore/components/BlockchainApplicationStatistics/index.js
oskarleonard: It is uppercase to indicate that the prop should be a component/dom node. React components need to be uppercase if used in this way. I havent found a pattern for this in our code, but if we want it lowercase we would have to do this in the component
image

Which is a bit more messy imo. Or what do you think?

src/modules/common/components/div/Div.js
ManuGowda: 😅 thats a fair point, we do not have a standard at the moment, everyone using their own style i think.

Here is my inputs:
- camelCase: hooks
- Pascal: Components, Directory

Probably we should first define and address it for the whole project.

src/modules/blockchainApplication/explore/components/BlockchainApplicationStatistics/index.js
ManuGowda: make sense, I wasn't aware of this, lets keep it as it is.

src/modules/common/components/div/Div.js
oskarleonard: Yes, this would be great to define. I think it has become mixed up when introducing the index pattern, or reversed (moving away from index pattern).

Some notes

**What everybody probably will agree on is that React Components should be Pascal**  
And my input would be that almost everything else should be camelCase. (And to me, the .css module that is part of the component should also be Pascal, while global css should be camelCase or hyphen).

**The index pattern** 
`/IndexComponent/index.js` can be imported like this:
`import IndexComponent from '@x/IndexComponent';`
To make the directory Pascal here make sense, With an index file the directory "is" _the_ component because of the way it is imported. 

**The none index pattern** 
`/indexComponent/IndexComponent.js` can be imported like this:
`import IndexComponent from '@x/indexComponent/IndexComponent';`
To make the directory Pascal here does not make as much sense in my opinion because it can no longer be seen as the component.
_But i wouldn't mind to have every folder which is exporting a component to be Pascal. As long as we are consistent :)._ So it would also be:
`/normalDirectory/IndexComponent/IndexComponent.js` having grouping directories as camelCase though

**Other files**
hyphen-case: some-example-images.png, 1-circleci.yml

src/modules/blockchainApplication/explore/components/BlockchainApplicationStatistics/index.js
oskarleonard: Great, it is nice to have this specific prop in uppercase to indicate this

ManuGowda (approved):

Issue #4616, opened 2022-11-30
Not making a change to the custom derivation when creating an account give you an error


oskarleonard:
### Expected behavior
When user adds a new account and make no change to the pre filled derivation path the default should be used.

### Actual behavior
Making no change to the pre filled value will try to create the account with an undefined derivation path and lead to error
![image](https://user-images.githubusercontent.com/8784876/204768084-1274051f-f6e6-4fd2-8221-164a4249ba68.png)

### Steps to reproduce
- Enable custom derivation path in settings
- Add an account
- Dont change custom derivation path
- Try do continue adding the account
- After trying to setup a password you will see the toast "Failed to setup password"

### Which version(s) does this affect? (Environment, OS, etc...)
v3

soroushm:
@oskarleonard then why you think its about drivition path? if you turn off the custom drivition path is passing?

oskarleonard:
> @oskarleonard then why you think its about drivition path? if you turn off the custom drivition path is passing?

Yes, works if i turn off derivation path. We are using two sources of truth for this and never initialize redux with the default value when setting it to true. Which means that 
![image](https://user-images.githubusercontent.com/8784876/204770068-b5fcd377-cfa8-4f38-83f6-cc6ba2d78385.png)
`customDerivationPath` will be undefined. And why it works for Manu in [issue](https://github.com/LiskHQ/lisk-desktop/issues/4591) is because he is triggering the onChange here 
![image](https://user-images.githubusercontent.com/8784876/204770391-4bc6002c-e5e4-4ca9-9214-c2be1d63d7e0.png)

So this issue is very related to [this](https://github.com/LiskHQ/lisk-desktop/issues/4591), and both need to be kept in mind when fixing this.

soroushm:
Similar issues #4591

PR #4658, opened 2022-12-08
Fix custom derivation path

### LINKED ISSUE #4591: Default custom derivation path setting is overridden
### Expected behavior

When a user adds new account the default custom derivation path should be `m/44'/134'/0'`

### Actual behavior

Default custom derivation path setting is overridden
Screenshot 2022-11-23 at 11 59 04 AM

### Steps to reproduce

- Enable custom derivation path in `settings`
- Add an account
- Change default path to `m/44'/134'/5'`
- After the successful account addition, add another account
- Observe the default derivation path is set to `m/44'/134'/5'` instead of `m/44'/134'/0'`

v3

oskarleonard:
### What was the problem?

This PR resolves #4591 #4616

### How was it solved?

- Remove custom derivation path from redux 
- Add validation to derivation path, typing an incorrect derivation path shall prevent a user to go to SetPasswordForm
- Fix alignment bug of error icon when input size = m

### How was it tested?

Manually


src/modules/account/components/AddAccountForm/AddAccountForm.js
```diff
@@ -22,18 +23,42 @@ const AddAccountForm = ({ settings, onAddAccount }) => {
...
+  children,
+}) => {
+  const { t } = useTranslation();
+
   const onFormSubmit = (e) => {
     e.preventDefault();
     // istanbul ignore else
     if (passphrase.value && passphrase.isValid) {
-      onAddAccount(passphrase);
+      onAddAccount(passphrase, derivationPath);
```
ManuGowda, line 49: Now that you separated the logic, I am not sure why do we need to pass this property? an account can be created with or without derivationPath.
- With derivationPath: either default or specific
- Without derivationPath: nothing to pass

src/modules/wallet/utils/account.js
```diff
@@ -303,3 +304,48 @@ export const validate2ndPass = async (account, passphrase, error) => {
   }
   return messages;
 };
+
+/**
+ * Validate a derivation path
+ * TODO: Replace this function when @liskhq (lisk-sdk) exposes the one they use
```
ManuGowda: could you please GitHub issue in SDK repo and link the issue here?

src/modules/account/components/AddAccountForm/AddAccountForm.js
oskarleonard: The `AddAccountForm` handle both those states so for one state `derivationPath` is return and for the other state `derivationPath` is not returned.

src/modules/wallet/utils/account.js
oskarleonard: Do you mean that we shall wait to merge this until response from SDK?

https://github.com/LiskHQ/lisk-sdk/issues/7877

src/theme/buttons/css/base.css
```diff
@@ -65,6 +65,7 @@
 
   &:disabled {
     opacity: 0.58;
+    cursor: default;
```
clemente-xyz, line 68:
```suggestion
    cursor: not-allowed;
```

clemente-xyz (changes requested):
small obs

src/theme/buttons/css/base.css
oskarleonard: I thought of that one but `cursor: default;` seems to be the current state for disabled buttons when i searched through the codebase, I didn't want to break any current design patterns (I have worked with designers that think `not-allowed` just adds confusion). Many "go to" design systems dont use it either and it is not even used here on github. 

Personally, as a user it would just add confusion to me because barely no other site is using it for disabled states.

src/modules/wallet/utils/account.js
ManuGowda: No we dont need to wait, in line 310 TODO we can refer the above SDK issue and we can clean this up once SDK exposes the function

ManuGowda (approved):

src/modules/wallet/utils/account.js
oskarleonard: Done

src/theme/buttons/css/base.css
clemente-xyz: Hmm i don't agree with you at all :)

I think we should make use as much as possible of cursor state-changes resources — is a very important accessibility aid for users, specially the ones using keyboard controls.

That Github or other websites don't use them does not mean they are right :)

[MDN](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/CSS_and_JavaScript) explains good the point I'm addressing on this comment.

oskarleonard: I am fine with using it as long as it is consistent throughout the application. I have created an issue regarding this https://github.com/LiskHQ/lisk-desktop/issues/4664

clemente-xyz (approved):
Approving considering that it will be addressed on https://github.com/LiskHQ/lisk-desktop/issues/4664

PR #4661, opened 2022-12-08
Resolved issue with wallet transaction table

### LINKED ISSUE #4602: Transactions table and filter design is not matching the proposal
### Expected behavior

The table column, font size and colour should match according to design

### Actual behavior

Following table column alignment, font size and colour is not matching according to design
Screenshot 2022-11-23 at 1 34 55 PM

### Steps to reproduce
- Navigate to transactions monitoring tab

v3

eniolam1000752:
### What was the problem?

This PR resolves #4602

### How was it solved?

- [x] Fix status column
- [x] Fix table column alignment on smaller screen dimensions
- [x] Fix failing unit tests

### How was it tested?

Visually


ikem-legend (dismissed):
Looks good to me

ManuGowda (approved):

PR #4666, opened 2022-12-13
Refactor useSearch to eliminate duplications - Closes #4646

### LINKED ISSUE #4646: Adopt useSearch Hooks api call with existing one
### Description

Use search internally make api call with usecustomQuery which is duplicated on code and cache its need to be adjust with rest of the system

### Acceptance Criteria

- Adopt addressSearch to useDelegate
- Adopt delegateSearch to useDelegate
- Adopt transactionsSearch to useTransactions
- Adopt blockHeightSearch to useBlocks
- Make sure search functionality working fine
- Adopt unit test

### Additional Information
Still You can pass options to the existing hooks in order to avoid unnecessary api call

soroushm:
### What was the problem?

This PR resolves #4646 

### How was it solved?
by refractor and reuse exciting hooks

### How was it tested?
unit test


oskarleonard (dismissed):

reyraa (dismissed):
LGTM

oskarleonard (dismissed):

src/modules/search/hooks/useSearch.js
```diff
@@ -1,65 +1,48 @@
...
+  const blocks = useBlocks({
+    config: { params: { height: search } },
+    options: { enabled: isBlockHeight },
+  });
+
+  const isLoading =
+    delegates.isLoading || transactions.isLoading || addresses.isLoading || blocks.isLoading;
+
+  const isFetched =
+    delegates.isFetched && transactions.isFetched && addresses.isFetched && blocks.isFetched;
```
ikem-legend: @soroushm This was always be false because all the searches can't run at the same time. So I think it should be removed

oskarleonard: Isnt this more related to this https://github.com/LiskHQ/lisk-desktop/issues/4617 . Either way, we could fix it in that issue

src/modules/search/hooks/useSearch.js
```diff
@@ -34,15 +34,11 @@ export const useSearch = (search = '') => {
   const isLoading =
     delegates.isLoading || transactions.isLoading || addresses.isLoading || blocks.isLoading;
 
-  const isFetched =
-    delegates.isFetched && transactions.isFetched && addresses.isFetched && blocks.isFetched;
-
```
reyraa: So `isFetched` is not needed?
How about if we change it to
```
const isFetched =
     (isDelegate  && delegates.isFetched) || (isTxId  && transactions.isFetched) || (isAddress  && addresses.isFetched) || (isBlockHeight  && blocks.isFetched);
```

ikem-legend: @reyraa It might be needed but this is another way to handle it

reyraa (dismissed):

reyraa (approved):

ikem-legend (approved):

oskarleonard (approved):

Issue #4600, opened 2022-11-23
Update account menu texts


ManuGowda:
### Expected behavior

- Update the texts according to [design](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=38%3A13012&t=prTQCIyJ1X49773R-4) 

### Actual behavior

- The menu texts are not up to date

### Steps to reproduce

- Add an account
- Select account dropdown and compare with proposed design

### Which version(s) does this affect? (Environment, OS, etc...)
v3

PR #4667, opened 2022-12-13
Use the same derivation path validation as the SDK

### LINKED ISSUE #4665: Use the same derivation path validation as the SDK
### Description

Use the same derivation path validation as the SDK.

### Motivation

To not manage validation in two places. It is better with one source of truth. 

### Additional Information

Modify `getDerivationPathErrorMessage` to use cryptography.utils.parseKeyDerivationPath();

https://github.com/LiskHQ/lisk-sdk/issues/7877

ManuGowda:
### What was the problem?

This PR resolves #4665 and #4600

### How was it solved?

- Use `cryptography.utils.parseKeyDerivationPath`
- Updated menu items as designed

### How was it tested?

- Updated unit tests
- Test the menu items visual and compare with design


oskarleonard (dismissed):

src/modules/wallet/utils/account.js
```diff
@@ -312,37 +312,12 @@ export const validate2ndPass = async (account, passphrase, error) => {
  * @returns {string} - Error message
  */
 export const getDerivationPathErrorMessage = (derivationPath) => {
-  if ((!derivationPath || !derivationPath.startsWith('m')) || !derivationPath.includes('/')) {
```
soroushm, line 315: validation is missing now, i think we should keep the validation

oskarleonard: We still have the validation, `getDerivationPathErrorMessage` now makes use of the SDK function `parseKeyDerivationPath` to validate. To make sure we are using the same validation as the SDK

src/modules/account/components/EditAccount/EditAccount.js
```diff
@@ -17,7 +17,7 @@ const EditAccount = ({ history }) => {
       <EditAccountForm onBack={history.goBack} />
       <SetPasswordSuccess
         encryptedPhrase={currentAccount}
-        headerText={t('Edit name successful')}
+        headerText={t('Edit account name successful')}
```
oskarleonard, line 20: Shouldnt we commit this change in common.json file as well?

src/modules/wallet/utils/account.js
soroushm: in this case this error is message is wrong maybe we can have it as `"derivationPath not found"`

oskarleonard (dismissed):

soroushm (dismissed):

oskarleonard (approved):

PR #4692, opened 2022-12-30
Reward constants hook

### LINKED ISSUE #4639: Create Reward Constants Query Hooks
### Description
Fetch the reward module constants

### Acceptance Criteria

- Create UseRewardConstants hooks
- Add MSW mocks
- Add unit test

### Additional Information

- HTTP GET /api/v3/reward/constants
- RPC  get.reward.constants

oskarleonard:
### What was the problem?

This PR resolves #4639

### How was it solved?

- Create reward constants hook
- Move rewards hook to the reward module

### How was it tested?

Manually
Test


src/modules/reward/hooks/queries/useRewardConstants.js
```diff
@@ -0,0 +1,30 @@
...
+    event: 'get.reward.constants',
+    ...customConfig,
+  };
+
+  return useCustomQuery({
+    keys: [REWARD_CONSTANTS],
+    config,
+    options,
+  });
+};
```
eniolam1000752: because this would be a hook to get rewards constants from service, @soroushm @oskarleonard I think we had a discussion during grooming that the query hook should have a large stale/cache time, would this still be the case ?

oskarleonard: Hmm, that makes sense. Maybe we should set it to `Infinity` if those values almost never change. Do you remember how often these values change?

src/modules/reward/mocks/index.js
```diff
@@ -0,0 +1,7 @@
+import { rest } from 'msw';
+import { API_VERSION } from 'src/const/config';
+import { mockRewardConstants, mockRewardInflation } from 'src/modules/reward/__fixtures__';
+
+export const rewardConstants = rest.get(`*/api/${API_VERSION}/reward/constants`, async (req, res, ctx) => res(ctx.delay(20), ctx.json(mockRewardConstants)));
+
```
eniolam1000752, line 6: space can be removed

src/modules/transaction/utils/encoding.js
```diff
@@ -13,8 +13,7 @@
  *
  */
 
-import { codec } from '@liskhq/lisk-codec';
-import { utils } from '@liskhq/lisk-cryptography';
+import { codec as liskCode, cryptography  } from '@liskhq/lisk-client';
```
eniolam1000752: This should be `liskCodec` not `liskCode` I guess @oskarleonard 
```suggestion
import { codec as liskCodec, cryptography  } from '@liskhq/lisk-client';
```

eniolam1000752 (changes requested):

src/modules/reward/mocks/index.js
oskarleonard: Why? Isnt it nicer with a space in the end of files.Better diffs in the future when adding to this file

src/modules/transaction/utils/encoding.js
oskarleonard: Fixed

src/modules/reward/mocks/index.js
oskarleonard: ah, or between the functions you mean? I think it is more readable to have a space between, other mock endpoints have a space as well

eniolam1000752 (approved):

src/const/queries.js
```diff
@@ -31,4 +31,5 @@ export const BLOCKCHAIN_APPS_META_TOKENS = 'GET_BLOCKCHAIN_APPS_META_TOKENS';
 export const POS_CONSTANTS = 'GET_POS_CONSTANTS';
 export const POS_REWARDS_LOCKED = 'GET_POS_REWARDS_LOCKED';
 export const POS_REWARDS_CLAIMABLE = 'GET_POS_REWARDS_CLAIMABLE';
-export const POS_REWARDS_INFLATION = 'GET_POS_REWARDS_INFLATION';
```
soroushm, line 34: POS_REWARDS_INFLATION removed by mistake?

src/modules/transaction/utils/encoding.js
```diff
@@ -124,17 +123,17 @@ export const fromTransactionJSON = (transactionJSON, paramsSchema) => {
 export const toTransactionJSON = (transaction, paramsSchema) => {
   if (Buffer.isBuffer(transaction.params)) {
     return {
-      ...codec.toJSON(baseTransactionSchema, transaction),
-      params: paramsSchema ? codec.decodeJSON(paramsSchema, transaction.params) : {},
+      ...liskCodec.codec.toJSON(baseTransactionSchema, transaction),
```
soroushm, line 126: do we have ticket for liskCodec changes? if not please create one and assign to the pr

soroushm (changes requested):

Issue #4593, opened 2022-11-23
Recent transactions widget not rendered


ManuGowda:
### Expected behavior

Should render recent transactions widget
- Should show recent transactions when an account has transactions
- Should show empty widget when  an account has no transactions
<img width="1491" alt="Screenshot 2022-11-23 at 12 05 11 PM" src="https://user-images.githubusercontent.com/6449871/203485415-33b7a36a-4b56-4a6c-b5b0-30541ec0163b.png">
<img width="1465" alt="Screenshot 2022-11-23 at 12 05 30 PM" src="https://user-images.githubusercontent.com/6449871/203485421-0b5b43a6-6442-4503-b12f-cb730b0e9e09.png">

### Actual behavior

Recent transactions widget not rendered

### Steps to reproduce

- Add an account 
- Navigate to dashboard

### Which version(s) does this affect? (Environment, OS, etc...)
v3

PR #4694, opened 2023-01-02
Update recent transactions widget 

### LINKED ISSUE #4534: Update recent transactions widget
### Description
Recent transaction in dashboard also transaction list in wallet is still using withdata to fetch data its need to be adopt with useQuery api call

### Acceptance Criteria
- Integrate Api call with useTransactions hook
- Make sure [design](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=124%3A77502) requirement is matched
- Integrate with queryTable in order to show loading state
- Adopt unit test

reyraa:
### What was the problem?

This PR resolves #4534

### How was it solved?

This seems to be a follow up ticket to #4593. I've already updated the design and used the `useTransactions` hook. The only thing left is the use of QueryTable which I've done it here.

### How was it tested?

Tested manually.

### DIFF
--- a/src/modules/transaction/components/RecentTransactions/RecentTransactions.js
+++ b/src/modules/transaction/components/RecentTransactions/RecentTransactions.js
@@ -13,7 +13,7 @@ import BoxHeader from 'src/theme/box/header';
 import BoxContent from 'src/theme/box/content';
 import BoxEmptyState from 'src/theme/box/emptyState';
 import Icon from 'src/theme/Icon';
-import Table from 'src/theme/table';
+import { QueryTable } from '@theme/QueryTable';
 import TransactionRow from '../TransactionRow';
 import styles from './RecentTransactions.css';
 import header from './RecentTransactionsHeaderMap';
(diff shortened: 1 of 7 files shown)


src/modules/transaction/components/RecentTransactions/RecentTransactions.js
```diff
@@ -63,14 +63,14 @@ const RecentTransactions = ({ className, t }) => {
...
       </BoxHeader>
       <BoxContent className={styles.content}>
-        <Table
-          data={transactions?.data || []}
-          isLoading={transactions?.isLoading}
+        <QueryTable
+          queryHook={useTransactions}
+          queryConfig={{ config: {
+            params: { limit: 5, address: host },
+            options: { enabled: !!host }
```
eniolam1000752: I think for readability, instead of making the variable name `host`, it an be `currentAddress` or `hostAddress`

eniolam1000752 (changes requested):

src/theme/QueryTable/QueryTable.js
```diff
@@ -40,6 +41,12 @@ export const QueryTable = ({
     </LoadNewButton>
   ) : null;
 
+  useEffect(() => {
+    if (isFetched && typeof props.onFetched === 'function') {
```
oskarleonard: `if (isFetched && typeof onFetched === 'function') {`

`onFetched` seems to only be part of the `QueryTable` api, `Table` will never need it?

Maybe better to destruct it in `QueryTable` props to prevent passing dead props to Table.
```export const QueryTable = ({ queryHook, xxx, transformResponse,  onFetched, ...props```

reyraa: Total agree with you, but this would be refactoring QueryTable. I have no idea how many places I'll break. This deserves a dedicated ticket.

src/modules/transaction/components/RecentTransactions/RecentTransactions.js
reyraa: Renamed it in the recent transactions component.

src/theme/QueryTable/QueryTable.js
oskarleonard: I think you misunderstood me but then i see your last commit and it seems like you understood me after all :D

oskarleonard (approved):

eniolam1000752 (approved):

Issue #4632, opened 2022-12-04
Adapt fee computation for all transactions


sridharmeganathan:
### Description

Adapt the Lisk protocol fee calculation.

##### Compute minimum fee 

All transactions in Lisk has minimum fee, calculate the [minFee](https://github.com/LiskHQ/lisk-sdk/blob/f6641aca002a563b8cd8c19729b3f3479a36c0d8/elements/lisk-transactions/src/fee.ts#L28) using following function by passing the valid input parameter based transaction (e.g for multisignature registration/transaction from multsig account requires particular options to be passed to work properly)

##### Token transfer

- call token_getInitializationFees to get the set fees
e.g 
```
/api/v3/invoke

{
 "endpoint": "token_getInitializationFees",
  "params": {}
}
```
- check if recipientAddress exists on the chain by token_userAccountExists
- if recipientAddress does not exist, add userAccount fee to the fee
- Use `Fees` subtitle [`Account initialization`](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=111%3A81733&t=3px49TNHq2uSvQZf-4)
  - Once an account has been initialised do not show Initialisation fee

##### Transfer CrossChain

- call token_getInitializationFees to get the set fees
- check if receivingChainID exists on the chain by token_escrowAccountExists
- if receivingChainID does not exist, add escrowAccount fee to the fee
- Use `Fees` subtitle [`Account initialization`](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=111%3A81733&t=3px49TNHq2uSvQZf-4)
 - Once an account has been initialised do not show Initialisation fee

##### Register Validator

- call pos_getRegistrationFee to get the fee
- when calling registerValidator command, add the fee
- Use `Fees` subtitle `Registration`

### Additional Information

- Use `invoke` endpoint with `endpoint` name and `params` to call below mentioned endpoints
- Show additional fee where required (e.g do not show initialisation fee on register validator flow, rather show registration fee)
  - Use correct title for following transactions:
    -  Token transfer, Transfer CrossChain, Reclaim, Register Multisignature: `Account initialization`
    - Register Validator: `Registration`
    - Transactions that do not have extra command fees: `Stake, Unstake, Unlock`
  - Ensure the UI and logic are updated to handle the above cases
- Consider caching the fee related endpoints as they wont change
[...]

PR #4695, opened 2023-01-02
Adapt the transaction creation and summary views to the new design guidelines

### LINKED ISSUE #4391: Adapt the transaction creation and summary views to the new design guidelines
### Description
Following the new design changes, we now have new styles for labels and values in all transaction forms. The transaction priority and fee values are also slightly different. We should make sure all forms and summery screens look similar to the designs.

### Motivation
- Adapt to the design guide
- Create a consistent user experience 
- The transaction form for all types for transactions design should be updated
  - We have only updated token transfer and few other transactions
  - We need to cross check following transaction design
    - Reclaim
    - Multi-signature registration
    - Delegate registration
    - Voting
    - Unlocking
    
### More Information
The differences are limited to:
- The back button has a different design and location.
- The "Continue" buttons have different
   - Title
   - Width
   - Border radius
- The screen title seems to have a different typography
- Input border color, and border radius, and value font size and weight are different.
- The form field title has a different typography. Same goes for the titles in the summary step, and the value of each title in the summary step.
- The password input step has different
   - Typography
   - Button title
   - Input spacing and border colors
   - Also the account information, unlike in the design, is not displayed in the app.

eniolam1000752:
### What was the problem?

This PR resolves #4391 

### How was it solved?
Resolve UI defects on the following transaction types;

- [x] `send token` flow
- [x] `register validator` flow
- [x] `Multi signature` tx flows
- [x] `Reclaim` transaction flow 
- [x] `Staking` flow
- [x] fix unit tests

### How was it tested?

- Visually
- jest


oskarleonard:
I dont see any code in this PR that requires us to update our package-lock.json files

src/modules/legacy/components/AccountsComparison/index.js
```diff
@@ -16,7 +16,12 @@ const AccountsComparison = ({ t }) => {
   useDeprecatedAccount();
   useSchemas();
   const wallet = useSelector(selectActiveTokenAccount);
+  const nonce = wallet.sequence?.nonce;
   const hasEnoughBalance = Number(wallet.token?.[0]?.availableBalance) >= dustThreshold;
+  const hasDepositedTokens = useMemo(
```
oskarleonard: Feels unjustified to use useMemo for this calculation. Using useMemo also has its performance implications, for example, it does a `shouldThisUpdate` operation, which could have worse performance than checking `hasEnoughBalance && nonce >= 0`... 

I comment this because i see that we have a very generous useMemo/useCallback pattern i this codebase. useMemo and React.memo should not be thought of as guarantees of freezing a value, it is better to first not use them to make sure code works without them (very difficult bugs can appear if thinking of them as "freezing a value") and later optimize.

src/modules/legacy/components/Status/Status.js
```diff
@@ -5,7 +5,7 @@ import statusMessages from './statusMessages';
 import styles from './status.css';
 
 const Status = ({
-  account, transactions, t,
+  account, transactions, t, prevStep
```
oskarleonard: I cannot see that we pass in `prevStep` as a prop to `<Status`, or is it from the MultiStep component? The name prevStep seems like a very specific name for an unspecific component, maybe `onRetry` would be better?

libs/wcm/constants/permissions.js
```diff
@@ -1,5 +1,5 @@
 export const SIGNING_METHODS = {
```
oskarleonard, line 1: NRTTF (not related to this file)
I dont see any code in this PR that requires us to update our package-lock.json files

src/modules/pos/validator/components/RegisterValidatorView/index.js
```diff
@@ -7,19 +7,27 @@ import Summary from '../RegisterValidatorSummary';
...
-      <Form t={t} />
-      <Summary t={t} />
-      <TxSignatureCollector t={t} />
-      <Status t={t} />
-    </MultiStep>
-  </Dialog>
-);
+const RegisterValidator = ({ history, t }) => {
+  const [isStepTxSignatureCollector, setIsStepTxSignatureCollector] = useState(false);
+  const onMultiStepChange = useCallback(({ step: { current } }) => {
```
oskarleonard, line 14: Similar to `useMemo` in the case that our MultiStep component is not optimised enough to make this a performance optimisation it will instead lead to wore performance and more complicated code.

src/theme/dialog/dialog.js
```diff
@@ -8,22 +8,21 @@ import Description from './description';
...
-const Dialog = ({
-  children, hasClose, hasBack, className, history,
-}) => {
+const Dialog = ({ children, hasClose, hasBack, className, history, size }) => {
   const onCloseClick = () => removeSearchParamsFromUrl(history, ['modal'], true);
   const onBackClick = () => history.goBack();
 
   return (
-    <div className={`${styles.wrapper} ${className}`}>
+    <div className={`${styles.wrapper} ${className} ${size ? styles[size] : ''}`}>
```
oskarleonard, line 16: if className and size are undefined it will look like this in the dom:
`class="wrapper  ">`

Excellent case for classNames lib to clean the dom and simplify the condition
`className={classNames(styles.wrapper, className, size && styles[size])}`

Now if they are empty, dom will look like:
`class="wrapper">`

src/modules/legacy/components/Status/Status.js
eniolam1000752: so prevStep is a function exposed by the `<MultiStep />` component to its children to navigate between child steps

libs/wcm/constants/permissions.js
eniolam1000752: I don't understand this comment

src/modules/pos/validator/components/RegisterValidatorView/index.js
eniolam1000752: The reason for this is because navigating between steps would cause a re-render every time since we want to track the step the `multistep` component has taken but all we need is a function that wouldn't get re-instantiated every time this occurs the reason, why I strongly advise we have a useCallback wrapping this function. 

Assuming the number of renders done by the host component isn't affected by stepping through the children then just  having a function would have been preferred. @oskarleonard @soroushm

src/theme/dialog/dialog.js
eniolam1000752: yhea true we could easily write one or use a package like `classNames` that simply performs this but I think that would be going out of scope for now on this PR

libs/wcm/constants/permissions.js
oskarleonard: Just wondering why the lock files are pushed for this PR? What new packaged do we need to update to make the changes in this PR run?

ManuGowda (changes requested):
@eniolam1000752 success and error screen needs some UI changes, please address these comments  

Screenshot 2023-01-09 at 10 43 49 AM
Screenshot 2023-01-09 at 10 26 55 AM

src/modules/pos/validator/components/RegisterValidatorView/index.js
oskarleonard: I understand that this is what we wanted to achieve. However, adding a callback here will not prevent that and It would still re-render as many times as if we dont have a callback here.

src/theme/dialog/dialog.js
oskarleonard: Just wanted to share that we actually have classnames installed now :D

src/modules/legacy/components/Status/Status.js
oskarleonard: Where are we using this legacy status with our forms? The statuses in the multistep seems to be specific statuses components.

src/modules/pos/validator/components/RegisterValidatorView/index.js
soroushm: @eniolam1000752 if the function in children is not a dependencies is not helping any how, unlike useMemo useCallback will memoize by each render that means its increase the memory usage

src/modules/token/fungible/components/SendView/index.js
```diff
@@ -12,22 +13,27 @@ import Status from '../SendStatus';
...
       <MultiStep
         key="send"
         finalCallback={backToWallet}
+        onChange={onMultiStepChange}
         className={styles.wrapper}
       >
         <Form initialValue={initialValue} />
         <Summary />
-        <TxSignatureCollector />
+        <TxSignatureCollector confirmText="Confirm and Sign" />
```
soroushm: missing translation
```suggestion
        <TxSignatureCollector confirmText={t("Confirm and Sign" )}/>
```

src/modules/pos/validator/components/RegisterValidatorView/index.js
soroushm: by the way i think it can be simplify by 
```js
const dialogSize = useMemo(
() => [2, 3].includes(current) ? 'sm' : undefiend
, [current]);
```

```
<Dialog hasClose size={dialogSize}>
```

ManuGowda (dismissed):

ManuGowda (dismissed):

src/modules/common/components/customRoute/index.js
```diff
@@ -46,13 +46,22 @@ const CustomRoute = ({ path, exact, isPrivate, forbiddenTokens, component, t, hi
     );
   }
 
+  // Redirect back to actual path when an account is not reclaimable
+  if (isMigrated && path === routes.reclaim.path) {
+    history.goBack();
+    return <div/>
```
soroushm: please redirect it to the dashboard 
```suggestion
     return <Redirect to={`${routes.dashboard.path}`} />;
```

src/modules/legacy/components/AccountsComparison/index.js
```diff
@@ -16,7 +16,9 @@ const AccountsComparison = ({ t }) => {
   useDeprecatedAccount();
   useSchemas();
   const wallet = useSelector(selectActiveTokenAccount);
-  const hasEnoughBalance = Number(wallet.token?.[0]?.availableBalance) >= dustThreshold;
+  const nonce = wallet.sequence?.nonce;
+  const hasEnoughBalance = +wallet.token?.[0]?.availableBalance >= dustThreshold;
```
soroushm: +string is not readable plese use number function

src/modules/legacy/components/AccountsComparison/index.js
```diff
@@ -77,47 +75,45 @@ const AccountsComparison = ({ t }) => {
                 </>
               </Tooltip>
               <br />
-              {!hasEnoughBalance && (
```
soroushm, line 80: why this condition is removed?

src/modules/legacy/components/AccountsComparison/index.js
```diff
@@ -77,47 +75,45 @@ const AccountsComparison = ({ t }) => {
...
                 )}
               </span>
             </div>
           </li>
         </ul>
       </section>
-      <DialogLink component="reclaimBalance">
-        <PrimaryButton className={styles.button} disabled={!hasEnoughBalance}>
+      <DialogLink component="reclaimBalance" data={{ tokenID: wallet.token?.[0]?.tokenID }}>
+        <PrimaryButton className={styles.button} disabled={!hasDepositedTokens}>
```
soroushm: user may have blance with nonce 0 they should be allowed to be claim their amount 
its looks wrong to me
@ManuGowda what's your openion here?

src/modules/legacy/components/Status/status.test.js
```diff
@@ -1,15 +1,20 @@
...
 import TxBroadcaster from '@transaction/components/TxBroadcaster';
 import accounts from '@tests/constants/wallets';
+import { useCommandSchema } from '@network/hooks';
+import { mockCommandParametersSchemas } from 'src/modules/common/__fixtures__';
 import Status from './Status';
 
 jest.mock('@libs/wcm/hooks/useSession', () => ({
   respond: jest.fn(),
 }));
+jest.mock('@network/hooks');
```
soroushm: you should not ever mock entier directory 😭😭😭😭😭😭
its coused an issue all of the hooks on the directory will break on the test and in some how may effected on the other testes use jest.spyOn insted of mock

```js
import * as networkhook from @network/hooks

jest.spyOn(networkhook, 'useCommandSchema')

useCommandSchema.mockImplementation(() => ({
```

cc @eniolam1000752 @ikem-legend @reyraa @oskarleonard @ManuGowda

src/modules/legacy/components/Summary/Summary.js
```diff
@@ -19,35 +19,37 @@ const Summary = ({ history, balanceReclaimed, nextStep, wallet, t, fees }) => {
...
-  };
+  const { minFee } = useTransactionFeeCalculation({
+    selectedPriority,
+    priorityOptions,
+    token,
+    wallet,
+    transactionJSON,
+  });
+
+  const accountInitializationFee = 0.04; // @TODO: https://github.com/LiskHQ/lisk-sdk/blob/f732135e90cd2f1ba1fd46dad045041e289f8763/framework/src/modules/token/constants.ts#L35
```
soroushm: please always add issue number to the todo #4632

src/modules/legacy/components/Summary/summary.test.js
```diff
@@ -10,17 +11,24 @@ import flushPromises from '@tests/unit-test-utils/flushPromises';
...
+const mockTxFeeCalculationFn = jest.fn();
+
 jest.mock('@auth/hooks/queries');
 jest.mock('@account/hooks', () => ({
   useCurrentAccount: jest.fn(() => [mockedCurrentAccount, jest.fn()]),
 }));
 
 jest.mock('@transaction/hooks/useTransactionFeeCalculation');
 jest.mock('@transaction/api');
+jest.mock('@network/hooks');
```
soroushm: same mock issue here

src/modules/legacy/components/Summary/summary.test.js
```diff
@@ -128,4 +140,23 @@ describe('Reclaim balance Summary', () => {
...
+    useTransactionFeeCalculation.mockImplementation((params) => {
+      mockTxFeeCalculationFn(params);
+      return {
+        minFee: { value: 0.001 },
+      };
+    });
+    mountWithCustomRouterAndStore(Summary, props, state);
+
+    act(() => {
+      expect(mockTxFeeCalculationFn).toHaveBeenCalledWith(
```
soroushm, line 155: it should be like
```
 expect(useTransactionFeeCalculation).toHaveBeenCalledWith(smt)
 
 ```

soroushm (changes requested):

src/modules/common/components/customRoute/index.js
ManuGowda: @soroushm  I think the scenario is different here, imagine you have already reclaimed the tokens and for some reason you navigate to explicitly to `/reclaim` then you should be redirected to where you were last e.g if i was in `/validators` tab and typed `/reclaim` then i should be navigated back to `/validator` and not `/dashboard`?

src/modules/legacy/components/AccountsComparison/index.js
ManuGowda: I think that logic was invalid, now we are checking for both balance and nonce, try to reclaim you see the user journey changes

ManuGowda (approved):

soroushm (approved):

Issue #4597, opened 2022-11-23
Remove unlock feature from wallet


ManuGowda:
### Expected behavior

Remove the ability to unlock from wallet. Refer the design and update the table accordingly https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=38%3A13142&t=prTQCIyJ1X49773R-4

Unlock should only exists under pos domain
### Actual behavior

<img width="1658" alt="Screenshot 2022-11-23 at 12 38 03 PM" src="https://user-images.githubusercontent.com/6449871/203489388-6d5b84fa-974a-4b4c-841b-a9c6807f4d4a.png">

### Steps to reproduce

- Navigate to Wallets tab
- Click on `all tokens`

### Which version(s) does this affect? (Environment, OS, etc...)
v3

PR #4711, opened 2023-01-11
Retain form values and fix unlock feature bug

### LINKED ISSUE #4699: Form input values are not retained while navigating back
### Expected behavior

All types of transaction should hold the input values when a user navigates `back` and `continue` during transaction submission.

### Actual behavior

Input values are cleared while navigates `back` and `continue` during transaction submission.

### Steps to reproduce

- Try to send token transfer/multisignature registration  and navigate back after transaction summary

3.0.0

ManuGowda:
### What was the problem?

This PR resolves #4699, #4597  

### How was it solved?

- Retain form values when navigated back to previous step

### How was it tested?

- Updated tests
- Tested manually

### DIFF
--- a/src/modules/pos/validator/components/SentStakes/SentStakes.js
+++ b/src/modules/pos/validator/components/SentStakes/SentStakes.js
@@ -5,7 +5,7 @@ import DialogLink from 'src/theme/dialog/link';
 import Box from 'src/theme/box';
 import { PrimaryButton } from 'src/theme/buttons';
 import BoxContent from 'src/theme/box/content';
-import Table from 'src/theme/table';
+import { QueryTable } from 'src/theme/QueryTable';
 import BoxHeader from 'src/theme/box/header';
 import { selectSearchParamValue } from 'src/utils/searchParams';
 import { useCurrentAccount } from '@account/hooks';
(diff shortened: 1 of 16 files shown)


src/modules/pos/validator/components/SentStakes/SentStakes.js
```diff
@@ -60,10 +60,10 @@ const SentStakes = ({ history }) => {
         </Heading>
       </BoxHeader>
       <BoxContent>
-        <QueryTable
+        <Table
```
soroushm: whats the reason to change query table to the table ?

ManuGowda: At max one can have 10 stakes, not beyond that, so we just need table for this scenario.

oskarleonard: I guess it is the loadmore button you want to fix with this? 
The root cause of that is that `useSentStakes` uses `useCustomInfiniteQuery` instead of `useCustomQuery`, i have a refactored that in my branch.

oskarleonard (dismissed):

oskarleonard (dismissed):

eniolam1000752 (dismissed):

src/modules/pos/validator/components/SentStakes/SentStakes.js
soroushm: please revert the changes, the hook should useCustomQuery its alrady taking care by oscar, if that one is changed the update button will removed at the same queryTable lets app stay consistent

soroushm (approved):

oskarleonard (approved):

PR #4712, opened 2023-01-11
Create a hook to calculate the tx fee using given parameters

### LINKED ISSUE #4632: Adapt fee computation for all transactions

### LINKED ISSUE #4716: Create a hook to calculate the tx fee using given parameters
### Description

We should be able to calculate transaction fees using on the following formula:
```
minimum fee + command fee + tie breaker
```
The fee also includes values such as
 - Account initialization fee
 - Cross chain initialization fee

A desired response should include the total fee and the components creating it:
```
[
 { value: 523000, type: 'total' },
 { value: 123000, type: 'minFee' },
 { value:  500000, type: 'commandFee'}
]
```

**Minimum fee**
We calculate the minimum fee based on the transaction size. 

**Command fee**
We fetch this value from Lisk Service. This value is defined by the blockchain application developer.

**Tie breaker**
This is also a value to fetched. This value takes the transaction pool status into account.

### Acceptance Criteria
- Create a hook to calculate the minimum fee using `computeMinFee` function from Lisk Elements.
- Use correct transaction schema for calculation of minimum fee.
- The response should be in the explained form.
- Secure the logic using unit tests.

reyraa:
### What was the problem?

Partially resolves #4716

### How was it solved?

- [x] Create a utility function to calculate fee for a given transaction, account and priority options.
- [x] Create a utility to extract corresponding schema.
- [x] Create a useTransactionFee hook to calculate fees and return total + components

### What is not included
- Retrieve and include command fee in useTransactionFee
- Enhance the hook to calculate values for
   - Minimum fee
   - Fee corresponding the selected priority
   - Fee corresponding maximum transferrable amount
- Create a hook to identify the status of recipient account (initialized?)
- Use above hook to calculate initialization fee and include it in the response (incorporate in total fee and add it to fee components)

### How was it tested?

Unit tests

### DIFF
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
@@ -1,0 +1,38 @@
+import { useEffect, useState } from 'react';
+import { useCommandSchema } from '@network/hooks';
+import { useAuth } from '@auth/hooks/queries';
+import { computeFee, getParamsSchema } from './utils';
+/**
+ * 
+ * @param {object} data
+ * @param {boolean} data.isValid Whether the transaction is valid or not. TxComposer defines this
+ * @param {object} data.wallet The sender account info as returned by useDeprecatedAccount
+ * @param {object[]} data.priorities An array of object as { value: number, title: string, selected: boolean }
+ * @param {object} data.transaction Transaction object as Lisk Element expects without fee
+ * @returns {object} The fee object with a total value, and a component value as an array of fees
+ * that contribute in the total value
+ */
+export const useTransactionFee = ({
+  isValid,
+  wallet,
+  priorities,
+  transaction,
+}) => {
+  const [fee, setFee] = useState(0);
+  const { data: auth } = useAuth({ config: { params: { address: wallet?.summary?.address } } });
+  const { moduleCommandSchemas } = useCommandSchema();
+  const schema = getParamsSchema(transaction, moduleCommandSchemas);
+
+  useEffect(() => {
+    setFee(computeFee(transaction, schema, auth, priorities, isValid));
+  }, [
+    priorities,
+  ]);
+
+  return {
+    total: fee,
+    components: [
+      { current: fee },
+    ],
+  };
+};
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
@@ -31,7 +31,9 @@ export const useTransactionFee = ({ isValid, senderAddress, priorities, transact
   );
 
   return {
+    /* istanbul ignore next */
     isLoading: isSchemaLoading || isLoading,
+    /* istanbul ignore next */
     isFetched: isSchemaFetched && isFetched,
     total: fee,
     components: [{ value: fee, type: 'bytesFee' }],
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
@@ -1,18 +1,17 @@
 import { useMemo } from 'react';
 import { useCommandSchema } from '@network/hooks/useCommandsSchema';
 import { useAuth } from '@auth/hooks/queries/useAuth';
-import { computeTransactionFee, getParamsSchema } from './utils';
+import { computeTransactionMinFee, getParamsSchema } from './utils';
 /**
  *
  * @param {object} data
  * @param {boolean} data.isValid Whether the transaction is valid or not. TxComposer defines this
  * @param {object} data.wallet The sender account info as returned by useDeprecatedAccount
- * @param {object[]} data.priorities An array of object as { value: number, title: string, selected: boolean }
  * @param {object} data.transaction Transaction object as Lisk Element expects without fee
  * @returns {object} The fee object with a total value, and a component value as an array of fees
  * that contribute in the total value
  */
-export const useTransactionFee = ({ isValid, senderAddress, priorities, transaction }) => {
+export const useTransactionFee = ({ isValid, senderAddress, transaction }) => {
   const {
     data: auth,
     isLoading,
@@ -26,15 +25,13 @@ export const useTransactionFee = ({ isValid, senderAddress, priorities, transact
   const schema = getParamsSchema(transaction, moduleCommandSchemas);
 
   const fee = useMemo(
-    () => computeTransactionFee(transaction, schema, auth, priorities, isValid && !isSchemaLoading),
-    [transaction, schema, auth, priorities, isValid, isSchemaLoading]
+    () => computeTransactionMinFee(transaction, schema, auth, isValid && !isSchemaLoading),
+    [transaction, schema, auth, isValid, isSchemaLoading]
   );
 
   return {
-    /* istanbul ignore next */
-    isLoading: isSchemaLoading || isLoading,
-    /* istanbul ignore next */
-    isFetched: isSchemaFetched && isFetched,
+    isLoading: isSchemaLoading || /* istanbul ignore next */ isLoading,
+    isFetched: isSchemaFetched && /* istanbul ignore next */ isFetched,
     total: fee,
     components: [{ value: fee, type: 'bytesFee' }],
   };
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
@@ -6,7 +6,7 @@ import { computeTransactionMinFee, getParamsSchema } from './utils';
  *
  * @param {object} data
  * @param {boolean} data.isValid Whether the transaction is valid or not. TxComposer defines this
- * @param {object} data.wallet The sender account info as returned by useDeprecatedAccount
+ * @param {string} data.senderAddress The sender address in Lisk 32 format
  * @param {object} data.transaction Transaction object as Lisk Element expects without fee
  * @returns {object} The fee object with a total value, and a component value as an array of fees
  * that contribute in the total value
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
@@ -32,7 +32,7 @@ export const useTransactionFee = ({
   return {
     total: fee,
     components: [
-      { current: fee },
+      { value: fee, type: 'bytesFee' },
     ],
   };
 };
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
@@ -1,9 +1,9 @@
-import { useEffect, useState } from 'react';
-import { useCommandSchema } from '@network/hooks';
-import { useAuth } from '@auth/hooks/queries';
-import { computeFee, getParamsSchema } from './utils';
+import { useMemo } from 'react';
+import { useCommandSchema } from '@network/hooks/useCommandsSchema';
+import { useAuth } from '@auth/hooks/queries/useAuth';
+import { computeTransactionFee, getParamsSchema } from './utils';
 /**
- * 
+ *
  * @param {object} data
  * @param {boolean} data.isValid Whether the transaction is valid or not. TxComposer defines this
  * @param {object} data.wallet The sender account info as returned by useDeprecatedAccount
@@ -12,27 +12,28 @@ import { computeFee, getParamsSchema } from './utils';
  * @returns {object} The fee object with a total value, and a component value as an array of fees
  * that contribute in the total value
  */
-export const useTransactionFee = ({
-  isValid,
-  wallet,
-  priorities,
-  transaction,
-}) => {
-  const [fee, setFee] = useState(0);
-  const { data: auth } = useAuth({ config: { params: { address: wallet?.summary?.address } } });
-  const { moduleCommandSchemas } = useCommandSchema();
+export const useTransactionFee = ({ isValid, senderAddress, priorities, transaction }) => {
+  const {
+    data: auth,
+    isLoading,
+    isFetched,
+  } = useAuth({ config: { params: { address: senderAddress } } });
+  const {
+    moduleCommandSchemas,
+    isLoading: isSchemaLoading,
+    isFetched: isSchemaFetched,
+  } = useCommandSchema();
   const schema = getParamsSchema(transaction, moduleCommandSchemas);
 
-  useEffect(() => {
-    setFee(computeFee(transaction, schema, auth, priorities, isValid));
-  }, [
-    priorities,
-  ]);
+  const fee = useMemo(
+    () => computeTransactionFee(transaction, schema, auth, priorities, isValid && !isSchemaLoading),
+    [transaction, schema, auth, priorities, isValid, isSchemaLoading]
+  );
 
   return {
+    isLoading: isSchemaLoading || isLoading,
+    isFetched: isSchemaFetched && isFetched,
     total: fee,
-    components: [
-      { value: fee, type: 'bytesFee' },
-    ],
+    components: [{ value: fee, type: 'bytesFee' }],
   };
 };
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
@@ -29,6 +29,7 @@ export const useTransactionFee = ({ isValid, senderAddress, transaction }) => {
     [transaction, schema, auth, isValid, isSchemaLoading]
   );
 
+  // @todo Assert the returned value when all fee components are defined
   return {
     isLoading: isSchemaLoading || /* istanbul ignore next */ isLoading,
     isFetched: isSchemaFetched && /* istanbul ignore next */ isFetched,
--- a/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.test.js
+++ b/src/modules/transaction/hooks/useTransactionFee/useTransactionFee.test.js
@@ -1,8 +1,18 @@
 import { renderHook } from '@testing-library/react-hooks';
 import wallets from '@tests/constants/wallets';
-import { mockAuth } from '@auth/__fixtures__';
-import { useAuth } from '@auth/hooks/queries';
+import { queryWrapper as wrapper } from 'src/utils/test/queryWrapper';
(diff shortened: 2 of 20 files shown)


src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -0,0 +1,38 @@
...
+  transaction,
+}) => {
+  const [fee, setFee] = useState(0);
+  const { data: auth } = useAuth({ config: { params: { address: wallet?.summary?.address } } });
+  const { moduleCommandSchemas } = useCommandSchema();
+  const schema = getParamsSchema(transaction, moduleCommandSchemas);
+
+  useEffect(() => {
+    setFee(computeFee(transaction, schema, auth, priorities, isValid));
+  }, [
```
oskarleonard: I dont think this is prettified?

oskarleonard (dismissed):

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
reyraa: I most likely touch this line in my next PR. I'll make sure it's well prettified then.

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param {object} data.transaction Transaction object as Lisk Element expects without fee
+ * @returns {object} The fee object with a total value, and a component value as an array of fees
+ * that contribute in the total value
+ */
+export const useTransactionFee = ({
+  isValid,
+  wallet,
+  priorities,
+  transaction,
+}) => {
```
ManuGowda:
```
export const useCalculateTransactionFee = ({
  isValid, // Not required? as the calling function can avoid calling this hook and populate `0` fee?
  wallet, // No need to pass the whole object?
  priorities,
  transaction,
}) => {
```

Suggestion
```
export const useCalculateTransactionFee = ({
  walletAddress: string,
  priorities,
  transaction,
}) => {
```

src/modules/transaction/hooks/useTransactionFee/utils.js
```diff
@@ -0,0 +1,59 @@
...
+ * @param {object} paramsSchema transaction schema as retrieved from Service
+ * @param {object} auth the auth WS response data
+ * @param {array} priorities list of objects with value (number) and title (string), and selected (boolean)
+ * @param {boolean} isTxValid defines if the form was validated successfully
+ * @param {bigint} extraFee
+ *
+ * @returns {bigint} the transaction fee in Beddows
+ */
+export const computeFee = (transaction, paramsSchema, auth, priorities, isTxValid, extraFee = BigInt(0)) => {
+  if (!isTxValid) return ZERO_FEE;
```
ManuGowda:
```suggestion
export const computeTransactionFee = (transaction, paramsSchema, auth, priorities, isTxValid, extraCommandFee = BigInt(0)) => {
  if (!isTxValid) return ZERO_FEE;
```
- `computeTransactionFee` more verbose name to avoid confusion with `computeMinFee`
- `if (!isTxValid) return ZERO_FEE;` we can remove this logic and reduce the complexity of function to be more generic, the calling function can handle this logic, more the responsibilities are isolated the function will be more atomic and less scope to change in future.

src/modules/transaction/hooks/useTransactionFee/utils.js
```diff
@@ -0,0 +1,59 @@
...
+    {
+      numberOfSignatures,
+      numberOfEmptySignatures,
+    }
+  );
+
+  return minFee + BigInt(selectedPriority.value * transactionSize) + extraFee;
+};
+
+export const getParamsSchema = (transaction, schemas) => {
```
ManuGowda, line 43: maybe throw an error when schema is not found? this way the error can be handle properly?
e.g for a given modulecommand schema might not exists then it will fail at this step itself?

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -0,0 +1,38 @@
...
+    priorities,
+  ]);
+
+  return {
+    total: fee,
+    components: [
+      { value: fee, type: 'bytesFee' },
+    ],
+  };
+};
```
soroushm: so useEffect dependcies is missing evrey single value inside the useEffect is a dependence other than var from global scope, there is no need for an setState witch is coused an extra re-render
```suggestion
const fees = usememo( () => {
const fee = computeFee(transaction, schema, auth, priorities, isValid);
return {
  totalFee: fee,
  fees: [{ fee, type: 'bytesFee' }],
}
, [transaction, schema, auth, priorities, isValid])
 return fees
```

src/modules/transaction/hooks/useTransactionFee/utils.js
```diff
@@ -0,0 +1,59 @@
...
+ * @returns {bigint} the transaction fee in Beddows
+ */
+export const computeFee = (transaction, paramsSchema, auth, priorities, isTxValid, extraFee = BigInt(0)) => {
+  if (!isTxValid) return ZERO_FEE;
+
+  const selectedPriority = priorities.find(item => item.selected);
+  const transactionSize = transactions.getBytes(transaction, paramsSchema).length;
+
+  const allocateEmptySignaturesWithEmptyBuffer = (signatureCount) =>
+    new Array(signatureCount).fill(Buffer.alloc(64));
```
ManuGowda: maybe define a constant for buffer length? e.g `const EMPTY_SIGNATURE_BUFFER_SIZE = 64`

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.test.js
```diff
@@ -0,0 +1,83 @@
...
+    tokenId: bufferify('00000000'),
+    amount: BigInt('1000000'),
+    recipientAddress: bufferify(wallets.genesis.summary.address),
+    data: '',
+  },
+  signatures: [],
+};
+
+jest.mock('@auth/hooks/queries');
+jest.mock('@network/hooks', () => ({
```
soroushm: useAuth is already mocked with msw please avoid moc it on this way by the way if you need to please use Jest.spyOn then mock it, mock entier directory it will efect other test using this hook

ManuGowda: import moduleCommandSchemas from "@tests/constants/schemas";
const schema = moduleCommandSchemas["token:transfer"];
useCommandSchema: jest.fn(() => ({ schema})?

soroushm (changes requested):

src/modules/transaction/hooks/useTransactionFee/utils.test.js
```diff
@@ -0,0 +1,98 @@
...
+      signatures: [],
+    };
+    const schema = moduleCommandSchemas["auth:registerMultisignature"];
+    const auth = {
+      numberOfSignatures: 1,
+    };
+
+    it('Returns the calculated fee given transaction is valid', () => {
+      const priorities = defaultPriorities.map((item) => ({ ...item, selected: item.title === 'Low' }));
+      expect(computeFee(registerMultisignature, schema, auth, priorities, true)).toEqual(BigInt(275000));
```
ManuGowda: instead of hardcoding, we could snapshot these for all types of transaction so that in future when someone accidentally changes the logic the snapshot can capture?

also can you compute for all transactions? so that we can ensure the fee calculation is done correctly for all types of transaction?

ManuGowda: maybe this will help to compose transactions https://github.com/LiskHQ/lisk-sdk/blob/transaction_fixture/framework/tx/tx_samples.json

some idea:

create `unsigned_transaction.json` under fixture
read the JSON file and 

```
it('compute transaction fee', () => {
  const unsignedTransactions = `unsigned_transaction.json`
  unsignedTransactions.forEach(transaction => {
    compute
    expect with snapshot
  });
})
```

src/modules/transaction/hooks/useTransactionFee/utils.js
reyraa: Renamed it, but I rather keep the invalidity check to reduce unnecessary calculations.

src/modules/transaction/hooks/useTransactionFee/utils.js
reyraa: If a schema doesn't exist, the entire transaction signature will fail. I think this should be handled in the form Leven, not fee calculation. case the fee calculation is not the only consumer of it.

src/modules/transaction/hooks/useTransactionFee/utils.js
reyraa: Done

src/modules/transaction/hooks/useTransactionFee/utils.test.js
reyraa: TBH I never liked the snapshots. The fact that you can run with `-u` and not care how and why the snapshots were updated worries me. I'd rather bring the attention of the developer down to the test code line to at least see what they're about to change.
Though if you insist I can change it to snapshots

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.test.js
reyraa: I went a whole different direction.

src/modules/transaction/hooks/useTransactionFee/utils.js
```diff
@@ -0,0 +1,67 @@
...
+ * @param {object} transaction transaction object in JSON format
+ * @param {object} paramsSchema transaction schema as retrieved from Service
+ * @param {object} auth the auth WS response data
+ * @param {array} priorities list of objects with value (number) and title (string), and selected (boolean)
+ * @param {boolean} isTxValid defines if the form was validated successfully
+ * @param {bigint} extraFee
+ *
+ * @returns {bigint} the transaction fee in Beddows
+ */
+export const computeTransactionFee = (
```
ManuGowda: @reyraa as discussed lets split this into three atomic hooks, which will be easy to understand and maintain

```
useComputeMinFee
useCommandFee
usePriorityAndTieBreakerFee
```

ManuGowda (changes requested):
one minor comment, please consider 🙏🏼

src/modules/transaction/hooks/useTransactionFee/utils.js
reyraa: I've removed the priority specific calculations and renamed the utility function not computeTransactionMinFee to represent the purpose. but I kept the hook name because I'm using it as the hook that populate all other fee types.

soroushm (dismissed):

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -0,0 +1,38 @@
...
+import { useAuth } from '@auth/hooks/queries/useAuth';
+import { computeTransactionMinFee, getParamsSchema } from './utils';
+/**
+ *
+ * @param {object} data
+ * @param {boolean} data.isValid Whether the transaction is valid or not. TxComposer defines this
+ * @param {object} data.wallet The sender account info as returned by useDeprecatedAccount
+ * @param {object} data.transaction Transaction object as Lisk Element expects without fee
+ * @returns {object} The fee object with a total value, and a component value as an array of fees
+ * that contribute in the total value
```
ManuGowda: this is outdated 😅 the params are changed now

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.test.js
```diff
@@ -0,0 +1,44 @@
...
+  it('Returns the calculated fee given transaction is valid', async () => {
+    renderHook(() => useTransactionFee({
+      isValid: true,
+      senderAddress: wallets.genesis.summary.address,
+      transaction,
+    }), { wrapper });
+
+    expect(useCommandSchema).toHaveBeenCalled();
+    expect(useAuth).toBeCalledWith({ config: { params: { address: wallets.genesis.summary.address } } });
+    expect(computeTransactionMinFee).toHaveBeenCalledTimes(1);
```
ManuGowda, line 42: The `useTransactionFee` returned values are not asserted?

src/modules/transaction/hooks/useTransactionFee/utils.js
```diff
@@ -0,0 +1,63 @@
...
+  transaction,
+  paramsSchema,
+  auth,
+  priorities,
+  isTxValid,
+) => {
+  if (!isTxValid || !paramsSchema || !auth || !priorities?.length) return ZERO_FEE;
+
+  const allocateEmptySignaturesWithEmptyBuffer = (signatureCount) =>
+    new Array(signatureCount).fill(Buffer.alloc(DEFAULT_SIGNATURE_BYTE_SIZE));
```
ManuGowda: We can eliminate this logic by passing the `options`? https://github.com/LiskHQ/lisk-sdk/blob/723817d2998270a6ea06d545d72adfca46a9d129/elements/lisk-transactions/src/fee.ts#L62

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
reyraa: Yes, wallet is changed to address, updated it.

ManuGowda (dismissed):

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.test.js
reyraa: Added a todo to handle this when the rest of the logic is implemented.

src/modules/transaction/hooks/useTransactionFee/utils.js
reyraa: We decided to delegate this to the consumer component

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -0,0 +1,39 @@
...
+    data: auth,
+    isLoading,
+    isFetched,
+  } = useAuth({ config: { params: { address: senderAddress } } });
+  const {
+    moduleCommandSchemas,
+    isLoading: isSchemaLoading,
+    isFetched: isSchemaFetched,
+  } = useCommandSchema();
+  const schema = getParamsSchema(transaction, moduleCommandSchemas);
```
oskarleonard, line 25: Would be nice if we could use the same variable name in using functions/components. Makes easier when reading the code and following it to the function args of `computeTransactionMinFee ` 
`const paramsSchema = getParamsSchema(transaction, moduleCommandSchemas);`

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {object} data.transaction Transaction object as Lisk Element expects without fee
+ * @returns {object} The fee object with a total value, and a component value as an array of fees
+ * that contribute in the total value
+ */
+export const useTransactionFee = ({ isValid, senderAddress, transaction }) => {
+  const {
+    data: auth,
+    isLoading,
+    isFetched,
+  } = useAuth({ config: { params: { address: senderAddress } } });
```
oskarleonard, line 19: Maybe we should add an enabled option for this hook and useCommandSchema, which checks for `isValid` ?

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -0,0 +1,39 @@
...
+    () => computeTransactionMinFee(transaction, schema, auth, isValid && !isSchemaLoading),
+    [transaction, schema, auth, isValid, isSchemaLoading]
+  );
+
+  // @todo Assert the returned value when all fee components are defined
+  return {
+    isLoading: isSchemaLoading || /* istanbul ignore next */ isLoading,
+    isFetched: isSchemaFetched && /* istanbul ignore next */ isFetched,
+    total: fee,
+    components: [{ value: fee, type: 'bytesFee' }],
```
oskarleonard, line 37: How will this work?

```
const {components} = useTransactionFee(...);

return (<div>{components.map()}...
```
What i am wondering, do we really want this hook tightly coupled with components, feels like we are over abstracting it somehow 🤔

soroushm (approved):

oskarleonard (approved):

Issue #4644, opened 2022-12-04
Allow user to visit locked reward in locked balance list


sridharmeganathan:
### Description

Allow user to see locked reward in locked balances

### Acceptance Criteria

- Allow user to find information about locked reward
- Make sure [design] is implemented correctly
- Add unit test

PR #4713, opened 2023-01-12
Implement unlock reward, view locked reward

### LINKED ISSUE #4645: Allow user to unlocked reward
### Description

Allow user to unlocked reward

### Acceptance Criteria
- Allow user to unlock rewards from stakes
  - Users should view only pending unlocks details with nothing to unlock [design](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=1450%3A253455&t=vXrkG6dSM9eCeU7O-4)
  - Users should be able to submit unlock when one/more unlocks available to claim [design](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=1450%3A195057&t=vXrkG6dSM9eCeU7O-4)
    - User should enter the password once reviewed the tokens to unlock 
    - User should view the success/fail modal 
- Add unit test

oskarleonard:
### What was the problem?

This PR resolves #4645 #4644 #4744

### How was it solved?

- Refactoring transaction logic to now being able to have an empty schema, ie passing not passing in params to TxComposer. 
- Update Sent Stakes table with more data
- Fix Sent Stakes table bugs
  - Dont load sent stakes when opening a modal to edit/delete a stake
  - Table should not have a load more button. 
- Enable unlock of stakes with new endpoint
- Add a button "Claim Rewards" that initiates a claim rewards flow.
- A bunch of renames from votes to stakes

### How was it tested?

Unit tests and manually


src/modules/legacy/store/action.test.js
```diff
@@ -51,7 +51,7 @@ describe('actions: legacy', () => {
       fee: 100000,
       module: 'legacy',
       command: 'reclaim',
-      moduleCommand: 'legacy:reclaimLSK',
+      moduleCommand: 'legacy:reclaimLSKLSK',
```
ManuGowda:
```suggestion
      moduleCommand: 'legacy:reclaimLSK',
```

src/modules/legacy/store/action.test.js
```diff
@@ -19,7 +19,7 @@ describe('actions: legacy', () => {
   };
   const formProps = {
     isValid: true,
-    moduleCommand: 'legacy:reclaimLSK',
+    moduleCommand: 'legacy:reclaimLSKLSK',
```
ManuGowda:
```suggestion
    moduleCommand: 'legacy:reclaimLSK',
```

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -0,0 +1,97 @@
...
+      formProps,
+      transactionJSON,
+      selectedPriority,
+      fees,
+    });
+  };
+
+  const unlockBalanceFormProps = {
+    moduleCommand: MODULE_COMMANDS_NAME_MAP.claimRewards,
+    isValid: 12 > 0,
```
ManuGowda: what is this logic 🤔

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -0,0 +1,97 @@
...
+  const amountInFiat = reward;
+
+  return (
+    <div className={classNames(styles.rewardsClaimableRow)}>
+      <div className={classNames(styles.logoContainer, rewardsClaimableHeader()[0].classList)}>
+        <img className={styles.logo} src={logo?.png} />
+        <span className={styles.tokenName}>{tokenName}</span>
+      </div>
+      <div className={classNames(rewardsClaimableHeader()[1].classList)}>
+        {`${reward} ${symbol}`}
```
ManuGowda:
```suggestion
        {`${reward} ${symbol}`}
```
this could be trickly, we need to convert reward from specific decimal , in this case beddows to lsk

1. use tokens meta hook response http://165.227.246.146:9901/api/v3/blockchain/apps/meta/tokens
2. match the token and use `decimals` to convert

src/modules/token/fungible/__fixtures__/mockTokens.js
```diff
@@ -62,7 +62,7 @@ export const mockAppsTokens = {
       displayDenom: 'lisk',
       symbol: 'LSK',
       logo: {
-        png: 'https://downloads.lisk.com/lisk/images/tokens/lsk.png',
+        png: 'https://lisk-qa.ams3.digitaloceanspaces.com/Artboard%201%20copy%2019.png',
```
ManuGowda, line 65: probably not a good idea to refer to this png? this will be destroyed once development is completed.

ManuGowda: either fully mock or refer to a dummy png?

src/modules/pos/reward/__fixtures__/mockRewardsClaimable.js
```diff
@@ -1,13 +1,41 @@
...
+
+    return {
+      ...rewardsClaimable,
+      ...token,
+    };
+  }),
+  meta: {
+    count: 3,
+    offset: 0,
+    total: 3,
```
ManuGowda: This can become buggy 😅 maybe use the above variables for actual value?

src/modules/token/fungible/__fixtures__/mockTokens.js
oskarleonard: It would be good to have this fully mocked, Clemente asked service about this last week but it was not a priority. So for now, to be able to test this feature in mock mode, it feels better to have a working correctly displaying image than a dummy one which would not show lisk logo

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
```diff
@@ -20,21 +21,43 @@ import { useCustomQuery } from '@common/hooks';
  * @returns the query object
  */
 
-export const useRewardsClaimable = ({ config: customConfig = {}, options } = { }) => {
-  const hasRequiredParams = customConfig.params?.address || customConfig.params?.name || customConfig.params?.publicKey;
+export const useRewardsClaimable = ({ config: customConfig = {}, options } = {}) => {
+  const hasRequiredParams =
+    customConfig.params?.address || customConfig.params?.name || customConfig.params?.publicKey;
+
+  const createMetaConfig = useAppsMetaTokensConfig();
```
ManuGowda, line 28: the variable should be `appMetaTokensConfig`?

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
```diff
@@ -20,21 +21,43 @@ import { useCustomQuery } from '@common/hooks';
...
+  const transformToken = addTokensMetaData({ createMetaConfig, client: defaultClient });
```
ManuGowda, line 29: probably the transformation logic is generic? and `useRewardsClaimable` is specific logic?

if we implement `useRewardsClaimable` atomically then we can write a generic logic for adding the token information?

src/modules/pos/validator/components/ClaimRewardsForm/ClaimRewardsForm.test.js
```diff
@@ -0,0 +1,69 @@
...
+    minFee: 0.001,
+  }));
+
+  const nextStep = jest.fn();
+
+  const props = {
+    nextStep,
+  };
+
+  it('Should display properly', async () => {
```
ManuGowda, line 50:
```suggestion
  it('Should display claimable rewards', async () => {
```
😅

src/modules/pos/validator/components/ClaimRewardsForm/ClaimRewardsForm.test.js
```diff
@@ -0,0 +1,69 @@
...
+    renderWithQueryClient(ClaimRewardsForm, props);
+    expect(screen.getByRole('heading', { name: 'Claim rewards' })).toBeTruthy();
+    expect(
+      screen.getByText(
+        'Below are the details of your reward balances, you can continue to claim your rewards and they will be transferred to your wallet balance.'
+      )
+    ).toBeTruthy();
+    mockRewardsClaimableWithToken.data.forEach(({ tokenName }) => {
+      expect(screen.getAllByText(tokenName)[0]).toBeTruthy();
```
ManuGowda, line 59: we could also assert for the rewards amount?

ManuGowda (changes requested):
- claim rewards [screen](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=2993%3A242469&t=7RTGj9DY7DTManaD-4) is not implemented?
- Rewards to claim implementation is not working 
- [unlock summary](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=2993%3A242341&t=7RTGj9DY7DTManaD-4) is not implemented as specified

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
oskarleonard: Ah, yes that is a better name

src/modules/pos/validator/components/ClaimRewardsForm/ClaimRewardsForm.test.js
oskarleonard: Current pattern for our test files look like this.  
`describe('ClaimRewardsForm'`
` it('Should display properly', async () => {` // Which is just a "smoke" test for the describe above it :/

src/modules/pos/validator/components/ClaimRewardsForm/ClaimRewardsForm.test.js
oskarleonard: Done

src/modules/pos/reward/__fixtures__/mockRewardsClaimable.js
oskarleonard: Done

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
oskarleonard: Hmm, not sure i completely understand what you mean here. This is how it works now
`fetchRewardsClaimable.then(fetchNeededTokensMetaData).then(mergeTokenDataWithRewards).then(addMergedResultToQueryCache)`

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
oskarleonard: Looks like i coudnt rename this without also refactor other code not related to this PR

ManuGowda (dismissed):

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -0,0 +1,99 @@
...
+        <img className={styles.logo} src={logo?.png} />
+        <span className={styles.tokenName}>{tokenName}</span>
+      </div>
+      <div className={classNames(rewardsClaimableHeader()[1].classList)}>
+        {`${parseInt(reward, denom)} ${symbol}`}
+      </div>
+      <div className={classNames(rewardsClaimableHeader()[2].classList)}>{amountInFiat}</div>
+    </div>
+  );
+};
```
eniolam1000752: Let's not have the row component here. 

To make this component cleaner and well defined, I would suggest abstracting this to a separate file just as we did for the rest of the other table types

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -0,0 +1,99 @@
...
+  },
+  {
+    title: t && t('Reward amount'),
+    classList: `${grid['col-xs-4']}`,
+  },
+  {
+    title: t && t('Fiat'),
+    classList: `${grid['col-xs-1']}`,
+  },
+];
```
eniolam1000752: Also I would suggest that this is also abstracted to a different file just as we did for the rest of our tables we can all this new file lets say `rewardsClaimableHeaderMap`.

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -0,0 +1,99 @@
...
+      </div>
+      <div className={classNames(rewardsClaimableHeader()[2].classList)}>{amountInFiat}</div>
+    </div>
+  );
+};
+
+const ClaimRewardsForm = ({ nextStep }) => {
+  const { t } = useTranslation();
+  const [currentAccount] = useCurrentAccount();
+  const address = currentAccount?.metadata?.address;
```
eniolam1000752: This can be done instead since the `currentAccount` variable isn't going to be used anywhere else and also currentAccount would also exist since the route to the `ClaimRewardsForm` is a private one

```suggestion
  const [{ metadata: { address } }] = useCurrentAccount();
```

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+const RewardsClaimableRow = ({ data }) => {
+  const { tokenName, logo, reward, symbol, denomUnits, displayDenom } = data;
+  const amountInFiat = reward;
+  const denom = denomUnits?.find((denomUnit) => denomUnit.denom === displayDenom);
+
+  return (
+    <div className={classNames(styles.rewardsClaimableRow)}>
+      <div className={classNames(styles.logoContainer, rewardsClaimableHeader()[0].classList)}>
+        <img className={styles.logo} src={logo?.png} />
```
eniolam1000752: I would advise to use the global helper function called `getLogo` to select the logo value for the token. You can find use cases of this function in the `SendForm` component

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
```diff
@@ -20,21 +21,43 @@ import { useCustomQuery } from '@common/hooks';
...
+      return {
+        ...rewardsClaimable,
+        ...token,
+      };
+    });
+    return {
+      ...res,
+      data: rewardsWithToken,
+    };
+  };
```
eniolam1000752, line 45: This would result to a Big O of n^2 (Quadratic time)

Instead can't we convert tokens to an object whose key it the tokenID and value is the token then within the `.map` on rewards response, we can just reference the token by accessing the token from the token map object. I think something like this

```suggestion
    const mapIDToToken = tokens.reduce((result, token) => ({...result, [token.tokenID] : token}) , {})
    const rewardsWithToken = res.data?.map((rewardsClaimable) => {
      const token = mapIDToToken[rewardsClaimable.tokenID]

      return {
        ...rewardsClaimable,
        ...token,
      };
    });
    return {
      ...res,
      data: rewardsWithToken,
    };
  };
```

This way we can reduce this logic's time complexity from Quadratic time to linear time

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -0,0 +1,99 @@
...
+    moduleCommand: MODULE_COMMANDS_NAME_MAP.claimRewards,
+    isValid: rewardsClaimable?.data?.length > 0,
+  };
+
+  return (
+    <section className={classNames(styles.ClaimRewardsForm)}>
+      <TxComposer
+        onConfirm={onConfirm}
+        formProps={unlockBalanceFormProps}
+        buttonTitle="Claim rewards"
```
eniolam1000752: text should be wrapped with translation

```suggestion
        buttonTitle={t("Claim rewards")}
```

src/modules/pos/validator/components/ClaimRewardsView/ClaimRewardsView.test.js
```diff
@@ -0,0 +1,18 @@
...
+describe('ClaimRewardsView', () => {
+  it('should render properly', () => {
+    const wrapper = shallow(<ClaimRewardsView history={{}} />);
+
+    expect(wrapper).toContainMatchingElement(Dialog);
+    expect(wrapper).toContainMatchingElement(MultiStep);
+    expect(wrapper).toContainMatchingElement(ClaimRewardsForm);
+    expect(wrapper).toContainMatchingElement(TxSignatureCollector);
+  });
+});
```
eniolam1000752: I think we should adopt less of `enzyme` and more of `testing-library` in creating our test suits

src/modules/transaction/utils/transaction.js
```diff
@@ -93,7 +92,7 @@ const convertObjectToHex = (data) => {
   // eslint-disable-next-line no-restricted-syntax, no-unused-vars, guard-for-in
   for (const key in data) {
     const value = data[key];
-    if (key === 'votes' || key === 'unlockObjects') {
+    if (key === 'stakes') {
```
eniolam1000752, line 95: was there a reason excluding ` key === 'unlockObjects'`

tests/constants/validators.js
```diff
@@ -4,7 +4,7 @@ const validator = index => ({
   selfStake: `10060000000${index}`,
   commission: 10000,
   validatorWeight: `100600000${index}000`,
-  address: 'lskhbxua8tpdckcewntcttfqfo4rbatampo2dgrno',
+  address: `lskhbxua8tpdckcewntcttfqfo4rbatampo2dgrno${index}`,
```
eniolam1000752, line 7: I think there has be a better way of dynamically generating these addresses because the above would result in invalid addresses

eniolam1000752 (changes requested):
It seems presently @oskarleonard,  you are not board-casting the the unlock claimable rewards transaction because

1. nextStep in the `ClaimRewardForm` is missing the `actionFunction`  property which is suppose to be the action responsible for signing the transaction
2. You don't have the `TxBroadcaster`(which is actually responsible for running a dry run and submitting the tx to the network) in the Multi step in the `ClaimableRewardView` component.

I am guessing there would be another ticket to address this right?



- It seems the height of the skeleton loader in the claimableRewardView is more than the actual height of a rendered row
image.  
image



-

src/modules/pos/validator/components/ClaimRewardsForm/index.js
oskarleonard: I agree, it is now in a separate file

src/modules/pos/validator/components/ClaimRewardsForm/index.js
oskarleonard: Also separate file now

src/modules/pos/validator/components/ClaimRewardsForm/index.js
oskarleonard: Fixed

src/modules/pos/validator/components/ClaimRewardsForm/index.js
oskarleonard: Yep, its cleaner like this if no need for optional chaining here. fixed

src/modules/pos/validator/components/ClaimRewardsForm/index.js
oskarleonard: Nice catch, fixed!

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
oskarleonard: Doing tokens.reduce and than merge has its own performance implications. Was curious about this so i ran it in a perf bench tool, seems to be 86% slower doing it like that. 
Below are some fun tools one can experiment with to test the performance of different operations 
https://jsben.ch/1M9pv
https://jsperf.app/voloyi

src/modules/transaction/utils/transaction.js
oskarleonard: Previously we needed to send the stakes to unlock (unlockObject) when unlocking stakes. In the new SDK version we should not send back the schema in the transaction hex. As can be seen in the schema documentation for `pos:unlock`
image

Good to know, before the new SDK update all module commands needed a schema. Now we have 2 module commands which shouldn't pass along a schema to the transaction (`pos:unlock` and `pos:claimRewards`). One cannot unlock specific stakes and rewards, one can only unlock **all** unlocks for ones address.

src/modules/pos/reward/hooks/queries/useRewardsClaimable.js
eniolam1000752: instead of a reduce we can just use a simple for operation to generate the map. eg

```suggestion
    const mapIDToToken = {};
    tokens.forEach((token) => { 
      mapIDToToken[token.tokenID] = token 
    })
    const rewardsWithToken = res.data?.map((rewardsClaimable) => {
      const token = mapIDToToken[rewardsClaimable.tokenID]
      return {
        ...rewardsClaimable,
        ...token,
      };
    });
    return {
      ...res,
      data: rewardsWithToken,
    };
  };
```

my suggestion actually is that, I think we can do linear time instead of a quadratic time

oskarleonard: If you paste it in the above tools you will find that this is still slower,  ~~now 80% slower~~. [Edit, 2% slower ](https://jsperf.app/voloyi/5), but the more tokens we get the slower it becomes, 2% slower with 100 tokens, 40% slower with 1000 tokens

tests/constants/validators.js
oskarleonard: Unsure on how to do this and still making one endpoint work together another endpoint which needs to have same address. Best would be to have a test database of all our response types to make sure our mocks are in sync. But that would require a lot more code. 

This is how the whole flow is https://github.com/LiskHQ/lisk-desktop/pull/4713/commits/7307b698df55c918bd27bde7f02e34238b9d8d5d

src/modules/pos/validator/components/ClaimRewardsForm/RewardsClaimableRow/index.js
```diff
@@ -0,0 +1,26 @@
...
+      </div>
+      <div className={classNames(rewardsClaimableHeader[1].classList)}>
+        {`${parseInt(reward, denom)} ${symbol}`}
+      </div>
+      <div className={classNames(rewardsClaimableHeader[2].classList)}>{amountInFiat}</div>
+    </div>
+  );
+};
+
+export default RewardsClaimableRow;
```
ManuGowda: new line

src/modules/pos/validator/components/ClaimRewardsForm/utils/getRewardsClaimableHeader.js
```diff
@@ -0,0 +1,18 @@
...
+    title: t && t('Reward amount'),
+    classList: `${grid['col-xs-4']}`,
+  },
+  {
+    title: t && t('Fiat'),
+    classList: `${grid['col-xs-1']}`,
+  },
+];
+
+export default getRewardsClaimableHeader;
```
ManuGowda: new line

src/modules/transaction/components/TxBroadcasterWithStatus/index.js
```diff
@@ -0,0 +1,15 @@
+import { connect } from 'react-redux';
```
ManuGowda, line 1: why was this component necessary? is this the generic component?

ManuGowda (dismissed):

src/modules/transaction/components/TxBroadcasterWithStatus/index.js
oskarleonard: Yes, it is a generic component. It is similar to the abstraction made in store/status -> signAndDispatchTransaction. 
This component could replace UnlockBalanceStatus and RegisterValidatorStatus, but this PR just grows and grows so i decided not to refactor those components

ManuGowda (approved):

eniolam1000752 (approved):

PR #4722, opened 2023-01-16
Fetch command fee for token transfer and validator registration

### LINKED ISSUE #4717: Fetch command fee for token transfer and validator registration
### Description

For each command (transaction) implement command fee logic to fetch the fees required using generic endpoint `invoke`.

The following commands require command fee:

- Token transfer and Cross chain token transfer require initialization fee
- Register validator command require command fee

As a result the fee value should look like:

```
[
  { value: 10523000, type: 'total' }, //  The value of commandFee is added to the total value
  { value: 23000, type: 'minFee' },
  { value: 500000, type: 'commandFee' },
]
```

If an account has a zero balance and its nonce is zero it needs to be initialized 
You can retrieve commandFee (e.g token transfer initialization fee) using the `invoke` method from Lisk Service. Simply send a WS request as
```
{
  "endpoint" : "token_getInitializationFees",
  "params": {} // Empty
}
```

### Acceptance criteria 

- This hook should be generic so that we can reuse this hook for other command fee in the future
  - Accept command name, endpointName, endpointParams as input for this hook, and return the value from `invoke` proxy endpoint
- The command fee should be retrieved from Lisk Service
- The fee should be incorporated in the total value and as a dedicated component
- Sufficient unit test to cover the additional logic is desired
- Adapt token transfer, cross chain token transfer and register validator form to use the above created hook to compute command fee
- Fix number of empty signature calculation for multisignature groups.

### Additional information

Enhancing the hook created by #4716 , we should use fetch the command fee and incorporate it in the fee values.
As mentioned, the value of command fee is defined by the blockchain developer, and no calculation is involved in it. We should merely fetch it from Lisk Service and add it:
- To the total fee value
- As a dedicated fee component

reyraa:
### What was the problem?

This PR resolves #4717

### How was it solved?
- Add `useInvokeQuery`
- Use `useInvokeQuery` to create `useGetInitializationFees`
- Create `useCommandFee`
  - Fetch account initialization fee
  - Fetch command fees
  - Add above fees in the components
  - Incorporate the value in the total fee

### How was it tested?

I've added unit tests for both hooks. For some reason I can't get msw to change the response of `/token` for one of my tests to return `[]`. I need to test the response when an account is not initialized, the that means mws should return no balance and nonce == '0'.


reyraa:
The team has decided to take a different approach. Therefore this PR is now considered obsolete.

PR #4725, opened 2023-01-17
Fix rpc api call

### LINKED ISSUE #4724: API call over RPC does not work as intended
### Expected behavior
While in the config `METHOD` is set to`rpc` Client should use rpc over the rest

### Actual behavior
application will get exception during the response is not as expected

### Steps to reproduce
change `METHOD` in config.js  to rest

v3 development

soroushm:
### What was the problem?

This PR resolves #4724

### How was it solved?
by changing the response to result and using socketRPC

### How was it tested?
Unit test


eniolam1000752 (dismissed):
LGTM

oskarleonard (approved):

eniolam1000752 (approved):

Issue #4746, opened 2023-01-19
Visual issues were found in the transaction summary view of multisig accounts


ikem-legend:
### Expected behavior
The transaction summary view for multisig accounts should display properly

### Actual behavior
Member of multisig account are not found in the transaction summary view of multisig accounts
<img width="748" alt="Screenshot 2023-01-18 at 04 22 13" src="https://user-images.githubusercontent.com/31376203/213345841-8e4dbba9-fef9-4cc9-a729-62893e7a67e0.png">

### Steps to reproduce
- Create a transaction using a multisig account
- All necessary members of the multisig account should sign the transaction
- Observe the transaction summary page

### Which version(s) does this affect? (Environment, OS, etc...)
All

PR #4752, opened 2023-01-19
Fix account multi signature status and and members

### LINKED ISSUE #4729: A group account should indicate its a multi sig account in its popover menu list
### Expected behavior
A multi signature account should not display `Register Multisignature` in its popover list if it has already been registered as a group account.

### Actual behavior
A multi signature account still displays `Register Multisignature` in its popover list if it has already been registered as a group account.
image

### Steps to reproduce
- switch to a multi sig account
- click on the account menu at the top left conner of the application
- would would still see that popover has the menu item called `Register Multisignature`

all

eniolam1000752:
### What was the problem?

This PR resolves #4729 #4746 

### How was it solved?

- [x] Add conditional to AccountMenuListing
- [x] make members props value in txSummarizer an array of object
- [x] fix unit tests

### How was it tested?
jest
visually

### DIFF
--- a/src/modules/account/components/AccountManagementDropdown/AccountManagementDropdown.test.js
+++ b/src/modules/account/components/AccountManagementDropdown/AccountManagementDropdown.test.js
@@ -2,15 +2,20 @@ import { screen, fireEvent } from '@testing-library/react';
 import { renderWithRouter } from 'src/utils/testHelpers';
 import mockSavedAccounts from '@tests/fixtures/accounts';
 import { truncateAddress } from '@wallet/utils/account';
+import { mockAuth } from 'src/modules/auth/__fixtures__';
+import { useAuth } from 'src/modules/auth/hooks/queries';
 import AccountManagementDropdown from './AccountManagementDropdown';
 
 const mockCurrentAccount = mockSavedAccounts[0];
 const mockOnMenuClick = jest.fn();
+jest.mock('@auth/hooks/queries');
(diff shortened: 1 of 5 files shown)


oskarleonard (dismissed):

src/modules/account/components/AccountManagementDropdown/AccountManagementDropdown.test.js
```diff
@@ -2,15 +2,20 @@ import { screen, fireEvent } from '@testing-library/react';
...
 
 const mockCurrentAccount = mockSavedAccounts[0];
 const mockOnMenuClick = jest.fn();
+jest.mock('@auth/hooks/queries');
 jest.mock('../../../account/hooks/useCurrentAccount.js', () => ({
   useCurrentAccount: jest.fn(() => [mockCurrentAccount]),
 }));
 
 describe('AccountManagementDropdown', () => {
+  useAuth.mockReturnValue({ data: mockAuth });
```
soroushm: why?
auth it should return the mock from msw why we mock it like this?

eniolam1000752: reason is that it messes up and it's hard to control. We've tried it previously and we had to go with just mocking it. @soroushm

soroushm (approved):

oskarleonard (approved):

PR #4756, opened 2023-01-23
Fix issues with request token

### LINKED ISSUE #4737: Issues with Request token
### Expected behavior
- Recipient chain selection should render the chain logo and name
- token selection should render the right token values and logo
- Request token link should open the send form when visited

### Actual behavior

- Recipient chain selection don’t render the chain logo and name
Recipient Application

- Token selection renders wrong token values (this should be gotten from the useTransferableTokens hook)
+ Add message (Optional)

- On redirecting to the request link generated , the send token modal doesn’t open

### Steps to reproduce
- Visit the wallet page
- Click on the request button
- above issues raised should be observable

all

eniolam1000752:
### What was the problem?

This PR resolves #4737 

### How was it solved?

- [x] Refactor `Request`` component from class to function
- [x] Consume `useTransferableTokens` hook to get list of tokens
- [x] Consume `useApplicationExplore` hook to get list of chains
- [x] Fix composed shareable link
- [x] Fix SendForm to work with generated request link
- [x] Update unit tests

### How was it tested?

- jest
- visually

### DIFF
--- a/src/modules/wallet/components/request/request.js
+++ b/src/modules/wallet/components/request/request.js
@@ -1,21 +1,50 @@
-import React from 'react';
+/* eslint-disable max-statements */
+import React, { useEffect, useMemo, useReducer, useRef } from 'react';
 import { regex } from 'src/const/regex';
 import { maxMessageLength } from '@transaction/configuration/transactions';
-import { mockAppTokens } from '@tests/fixtures/token';
+import { useApplicationExploreAndMetaData } from '@blockchainApplication/manage/hooks';
+import { useTransferableTokens } from '@token/fungible/hooks';
 import { validateAmountFormat } from 'src/utils/validators';
+import { getLogo } from 'src/redux/actions';
 import { sizeOfString } from 'src/utils/helpers';
 import { Input } from 'src/theme';
-import blockchainApplicationsExplore from '@tests/fixtures/blockchainApplicationsExplore';
 import Converter from 'src/modules/common/components/converter';
 import { useCurrentAccount } from 'src/modules/account/hooks';
 import i18n from 'src/utils/i18n/i18n';
 import Div from '@common/components/div/Div';
+import { useTranslation } from 'react-i18next';
 import MessageField from '../../../token/fungible/components/MessageField';
 import MenuSelect, { MenuItem } from '../MenuSelect';
 import RequestWrapper from './requestWrapper';
 import styles from './request.css';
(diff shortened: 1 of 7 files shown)


src/modules/wallet/components/request/request.js
```diff
@@ -32,61 +61,41 @@ const Account = () => {
...
+      metadata: { address },
+    },
+  ] = useCurrentAccount();
+
+  const { t } = useTranslation();
+  const [state, dispatch] = useReducer(
+    (stateData, value) => ({ ...stateData, ...value }),
+    requestInitState
+  );
+  const timeout = useRef({
```
oskarleonard: This seems to be dead code right now? What do we do with it?

src/modules/wallet/components/request/request.js
```diff
@@ -1,21 +1,50 @@
...
+  },
+  token: {
+    error: false,
+    value: '',
+    loading: false,
+    feedback: '',
+  },
+  recipientChain: {
+    error: false,
+    value: '',
```
oskarleonard: Would be nice if type of value wasnt mixed in the requestInitState. From the glance of it it looks like 
recipientChain.value represents a string holder but is in fact an object. So either 
`value: {},`
or
`value: undefined,`
would make more sense imo

src/modules/wallet/components/request/request.js
```diff
@@ -32,61 +61,41 @@ const Account = () => {
...
+    (stateData, value) => ({ ...stateData, ...value }),
+    requestInitState
+  );
+  const timeout = useRef({
+    amount: null,
+    reference: null,
+  });
+
+  const { applications } = useApplicationExploreAndMetaData();
+  const { data: tokens } = useTransferableTokens(state.recipientChain.value);
```
oskarleonard, line 77: Now when we use this as a hook this will produce RQ cache side effect when typing amount before selecting an application. RQ will cache a wrong client in `useTransferableTokens`... So we need to extend our `useTokensSupported` hook to include an enabled option and use like this 

```
  } = useTokensSupported({
    client: client.current,
    options: { enabled: !!application?.serviceURLs?.[0] },
  });
  ```

oskarleonard: image

To prevent this race condition with RQ

ikem-legend (approved):

oskarleonard (approved):

Issue #4719, opened 2023-01-13
Use transaction fee hook to calculate and display fees in transaction forms 


reyraa:
### Description

Assuming #4718 completes the requirements of the fee calculation, now it's time to adapt our React components to the new data structure.
We should update data wiring to match the changes applied by #4716 to #4718. The form component should display the total feee, alongside each fee component. It should also display human readable titles (with i18n) for each component.
 
### Acceptance criteria 
- Compose titles for all fee components
- Only display none-zero fee components
- Emphasis on the display of the total fee (Larger font, prominent color)
- Adapt unit tests
- Assert the returned value of `useTransactionFee`.

PR #4761, opened 2023-01-26
Integrate useTransaction fee hook on all transactions  

### LINKED ISSUE #4717: Fetch command fee for token transfer and validator registration

eniolam1000752:
### What was the problem?

This PR resolves #4717, #4719, #4781

### How was it solved?

- [x] Implement `usePriority` fee hook
- [x] refactor `useTransactionFee` hook to only consume `useByteFee` hook and `usePriority` hook
- [x] Abstract byte fee calculation to a separate hook
- [x] Integrate `useTransactionFee` hook into TxComposer
- [x] Integrate extra command hook on SendToken and register delegate form
- [x] Create a helper utils that converts a token to a particular denom
- [x] update unit tests

### How was it tested?

- jest
- visually


src/modules/pos/validator/components/UnlockBalanceForm/index.js
```diff
@@ -56,7 +64,7 @@ const UnlockBalanceForm = ({
           <BoxContent className={styles.container}>
             <p>
               {t(
-                'Below are the details of your locked balances and the unlock waiting periods. From here you can submit an unlock transaction when waiting periods are over.',
+                'Below are the details of your locked balances and the unlock waiting periods. From here you can submit an unlock transaction when waiting periods are over.'
```
reyraa: It appears that he prettier config is deviating from our Eslint config.

src/modules/token/fungible/utils/lsk.js
```diff
@@ -10,9 +10,22 @@ BigNumber.config({ ERRORS: false });
...
-);
+export const fromRawLsk = (value) =>
+  new BigNumber(value || 0).dividedBy(new BigNumber(10).pow(8)).toFixed();
+
+/**
+ * Convert Beddow to LSK
+ *
+ * @param {Strong|Number} amount - Amount value to be converted
+ * @param {Strong|Number} token - Token value merged with its equivalent metadata
+ * @returns {BigNumber} Amount value converted to the token symbol's denom
```
reyraa: A few things, Number is the documentation defines the interface, and number defines an instance of the Number interface. I think you need the later. 
The name is confusing. Since you are converting to a different denomination of the same token. I think `toOtherDenom`, or `changeDemnom`  could be a name.
The documentation doesn't clarify what should be inside the token. The default value is an empty object, while the documentation suggests a Number.

reyraa: Also, we already have two other methods. toRawLsk and fromRawLsk. Either refactor the components that use them or event better, create a token to do so.

src/modules/transaction/components/TxComposer/index.js
```diff
@@ -40,47 +50,61 @@ const TxComposer = ({
...
-  const [module, command] = splitModuleAndCommand(formProps.moduleCommand);
+
   const transactionJSON = {
     module,
     command,
-    nonce: wallet.sequence?.nonce,
-    fee: 0,
-    senderPublicKey: wallet.summary?.publicKey,
+    nonce: auth?.data?.nonce,
+    fee: formProps.extraCommandFee || 0,
```
reyraa: I've notice this in multiple places that you use a number as a default value for a bigNumber. I think changing the type arbitrarily is not a good idea.

src/modules/transaction/hooks/usePriorityFee.js
```diff
@@ -0,0 +1,20 @@
+import { transactions } from '@liskhq/lisk-client';
+import { fromTransactionJSON } from '../utils/encoding';
+import { FEE_TYPES } from '../constants';
+
+const usePriorityFee = ({ selectedPriority, transactionJSON, paramsSchema }) => {
+  const feePerByte = selectedPriority.value;
+
+  try {
```
reyraa: I'm not a fan of using `try-catch` for an error that we can prevent. This function may fail if the transaction is invalid, or the schema or not right. You have all the means to prevent both cases. 
Using a try catch causes the function to run, consume the time, and the fail in an oblivion fashion. You don't know why it fails, you just know it failed. Even worse, we are not even using the exception message to had the error gracefully.

src/modules/transaction/hooks/useTransactionFee/useTransactionFee.js
```diff
@@ -2,52 +2,89 @@ import { useMemo } from 'react';
...
 import { useAuth } from '@auth/hooks/queries/useAuth';
 import { computeTransactionMinFee, getParamsSchema } from './utils';
-import { useCommandFee } from './useCommandFee';
-import { FEE_TYPES } from './constants';
+import { FEE_TYPES } from '../../constants';
+import usePriorityFee from '../usePriorityFee';
 
-const getTotalFee = (components) =>
-  components.reduce((acc, { value }) => acc + value, 0);
+export const useByteFee = ({ isFormValid, senderAddress, transactionJSON }) => {
```
reyraa: Can't we have a dedicated file for this hook as we do for all the others?

src/modules/transaction/hooks/useTransactionFee/utils.js
```diff
@@ -1,40 +1,29 @@
...
  *
  * @returns {bigint} the transaction fee in Beddows
  */
-export const computeTransactionMinFee = (
-  transaction,
-  paramsSchema,
-  auth,
-  isTxValid,
-) => {
-  if (!isTxValid || !paramsSchema || !auth) return ZERO_FEE;
```
reyraa, line 25: This is why you need to `try-catch` at the top

reyraa (commented):
I left a few questions and concerns. Feel free to apply or respond.

src/modules/token/fungible/utils/lsk.js
eniolam1000752: So we already plan of having a ticket that would integrate all other usage of `toRawLsk` and `fromRawLsk` with the new util function. Though I think having a naming called `changeDemnom` is better

src/modules/transaction/hooks/usePriorityFee.js
eniolam1000752: true I am also not a fan of try-catch. I did this as a temporary solution though I have found a solution that would make me not use this

src/modules/auth/hooks/queries/useGetInitializationFees.js
```diff
@@ -0,0 +1,30 @@
...
+
+  const queryConfig = {
+    options: { enabled: !!address },
+    config: { params: { address } },
+  };
+
+  const { data: auth, isLoading: isAuthLoading } = useAuth(queryConfig);
+  const { data: token, isLoading: isTokenLoading } = useTokensBalance(queryConfig);
+
+  const isAccountInitialized = auth?.data?.nonce !== '0' && +token?.data?.[0]?.availableBalance;
```
ManuGowda: explicitly check if balance is greater than 0 `token?.data?.[0]?.availableBalance`

src/modules/auth/hooks/queries/useGetInitializationFees.js
```diff
@@ -0,0 +1,30 @@
...
+  const shouldReturnInitalizationFee = !isAccountInitialized && !isAuthLoading && !isTokenLoading;
```
ManuGowda, line 22: I think this flag is used to decide if we make api call or not? so `isEqueryEnabled` or something similar is better?

src/modules/auth/hooks/queries/useGetInitializationFees.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+jest.useRealTimers();
+
+jest.mock('@auth/hooks/queries/useAuth');
+jest.mock('@token/fungible/hooks/queries');
+
+beforeEach(() => jest.clearAllMocks());
+
+describe('useGetInitializationFees hook', () => {
+  it('does not call the useInoke query if token is still loading', () => {
```
ManuGowda:
```suggestion
  it('should not call the useInvoke query if token is still loading', () => {
```

ManuGowda: Please use `should` for below cases as well and fix the typo `Invoke`

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -37,6 +50,7 @@ const ClaimRewardsForm = ({ nextStep }) => {
       <TxComposer
         onConfirm={onConfirm}
         formProps={unlockBalanceFormProps}
+        commandParams={{}}
```
ManuGowda: why not composer commandParams default be set to empty object? this way we can avoid passing it at the first place

src/modules/pos/validator/components/UnlockBalanceForm/index.js
```diff
@@ -27,15 +37,19 @@ const UnlockBalanceForm = ({ nextStep }) => {
...
+      token: dposToken,
+    },
   };
 
   return (
     <section className={styles.wrapper}>
       <TxComposer
         onConfirm={onConfirm}
         formProps={unlockBalanceFormProps}
+        commandParams={{}}
```
ManuGowda: same here

src/modules/token/fungible/components/SendForm/SendForm.js
```diff
@@ -89,23 +90,27 @@ const SendForm = (props) => {
...
 
         return result;
-      }, true),
-    [amount, recipient, reference, recipientChain, sendingChain]
-  );
+      },
+      true
+    );
+    const isTokenValid = !!token && Object.keys(token).length;
+    return isFieldsValid && isTokenValid;
```
ManuGowda:
```suggestion
    return areFieldsValid && isTokenValid;
```

`is` for singular and `are` for plural?

src/modules/token/fungible/components/SendForm/SendForm.js
```diff
@@ -164,6 +170,7 @@ const SendForm = (props) => {
         onConfirm={onConfirm}
         formProps={sendFormProps}
         commandParams={commandParams}
+        commandParamsWithMaxBalance={{ ...commandParams, amount: token?.availableBalance }}
```
ManuGowda: dead code?

src/modules/token/fungible/utils/lsk.js
```diff
@@ -10,9 +10,22 @@ BigNumber.config({ ERRORS: false });
...
+  new BigNumber(value || 0).dividedBy(new BigNumber(10).pow(8)).toFixed();
+
+/**
+ * Converts a given token amount to its token symbol denom
+ *
+ * @param {BigNumber|number} amount - Amount value to be converted
+ * @param {BigNumber|number} token - Token value merged with its equivalent metadata
+ * @returns {BigNumber} Amount value converted to the token symbol's denom
+ */
+export const changeDemnom = (amount, token = {}) => {
```
ManuGowda:
```suggestion
export const convertToDenom = (amount, token = {}) => {
```

src/modules/token/fungible/utils/lsk.js
```diff
@@ -10,9 +10,22 @@ BigNumber.config({ ERRORS: false });
...
+  const { decimals = 8 } =
+    token.denomUnits?.find?.(({ denom }) => denom === token.symbol.toLowerCase()) || {};
+
+  return new BigNumber(amount || 0).dividedBy(new BigNumber(10).pow(decimals)).toFixed();
```
ManuGowda: define the constant BASE = 10;

```suggestion
  return new BigNumber(amount || 0).dividedBy(new BigNumber(BASE).pow(decimals)).toFixed();
```

src/modules/token/fungible/utils/lsk.js
```diff
@@ -10,9 +10,22 @@ BigNumber.config({ ERRORS: false });
...
+
+/**
+ * Converts a given token amount to its token symbol denom
+ *
+ * @param {BigNumber|number} amount - Amount value to be converted
+ * @param {BigNumber|number} token - Token value merged with its equivalent metadata
+ * @returns {BigNumber} Amount value converted to the token symbol's denom
+ */
+export const changeDemnom = (amount, token = {}) => {
+  const { decimals = 8 } =
```
ManuGowda: The default `8` can lead to buggy value, this is only true for `BEDDOWS->LSK`

src/modules/token/fungible/utils/lsk.test.js
```diff
@@ -21,4 +26,19 @@ describe('lsk', () => {
       expect(toRawLsk(0)).to.be.equal(0);
     });
   });
+
+  describe('changeDemnom', () => {
```
ManuGowda: their is typo, maybe once you update the function name change here too 🙏🏼

src/modules/token/fungible/utils/lsk.test.js
```diff
@@ -21,4 +26,19 @@ describe('lsk', () => {
...
+    it('should convert 100000000 to 1', () => {
+      expect(changeDemnom(100000000, mockToken)).to.be.equal('1');
+    });
+
+    it('should use default decimal value', () => {
```
ManuGowda, line 35: probably use a different decimal value to test different token? e.g `1000 SENV -> 1 ENV`

src/modules/transaction/components/TxComposer/index.js
ManuGowda: Thats a valid point.. but in this case its a JSON and not object 🤔 ? so it should be "0"?

src/modules/transaction/components/TxComposer/index.js
```diff
@@ -25,11 +27,19 @@ const TxComposer = ({
   formProps = {},
   commandParams,
 }) => {
+  const [module, command] = splitModuleAndCommand(formProps.moduleCommand);
   const { t } = useTranslation();
+
   // @todo Once the transactions are refactored and working, we should
   // use the schema returned by this hook instead of reading from the Redux store.
```
ManuGowda: I think this todo can be removed? as we are not going to remove the redux anytime soon. and if do plan to remove it we have to clean up the whole code base, so please remove this TODO if you agree.

src/modules/transaction/components/TxComposer/index.js
```diff
@@ -40,47 +50,61 @@ const TxComposer = ({
     prioritiesLoadError,
     loadingPriorities,
   ] = useTransactionPriority();
-  const [module, command] = splitModuleAndCommand(formProps.moduleCommand);
+
   const transactionJSON = {
     module,
     command,
-    nonce: wallet.sequence?.nonce,
```
ManuGowda, line 47: why are we removing this and adding `auth`? if its because its outdate information then we have to remove the usage of `wallet.token?.balance`? and `const wallet = useSelector(selectActiveTokenAccount);` in overall?

src/modules/transaction/components/TxComposer/index.js
```diff
@@ -40,47 +50,61 @@ const TxComposer = ({
...
+    {
+      title: 'Message',
+      value: getFeeStatus({
+        fee: Number(fromRawLsk(transactionJSON.params.messageFee || 0)),
+        token,
+        customFee,
+      }),
+      isHidden: !transactionJSON.params.messageFee,
+      components: [],
+    },
```
ManuGowda, line 99: Except cross chain token transfer other types of transactions do not need this logic? so its better to always avoid the logic execution where its not required?

src/modules/transaction/constants.js
```diff
@@ -0,0 +1,7 @@
+export const FEE_TYPES = {
+  BYTES_FEE: 'bytesFee',
+  PRIORITY_FEE: 'priorityFee',
+  INITIALIZE: 'initialize',
+  INITIALIZE_ESCROW: 'initializeEscrow',
+  REGISTER_VALIDATOR: 'registerValidator',
```
ManuGowda:
```suggestion
  USER_ACCOUNT_INITIALIZATION: 'userAccountInitialization',
  ESCROW_ACCOUNT_INITIALIZATION: 'escrowAccountInitialization',
  VALIDATOR_REGISTRATION: 'validatorRegistration',
```
lets try to express as much closely possible to protocol https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/token/endpoint.ts#L184

https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/pos/endpoint.ts#L335

src/modules/transaction/hooks/useTransactionFee/useByteFee.js
```diff
@@ -0,0 +1,47 @@
+import { useMemo } from 'react';
+import { useCommandSchema } from '@network/hooks/useCommandsSchema';
+import { useAuth } from '@auth/hooks/queries/useAuth';
+import { computeTransactionMinFee, getParamsSchema } from './utils';
+import { FEE_TYPES } from '../../constants';
+
+/**
+ *
```
ManuGowda: Missing documentation.

`Calculates minimum transaction fee based on bytes as per Lisk protocol`

ManuGowda (changes requested):
Nice job 🎉

src/modules/auth/hooks/queries/useGetInitializationFees.js
eniolam1000752: and its also used to determine if we should return a fee value from the query's cache because `react-query` would still return a cached value even though the `enable` flag on the query's option has been set to `false`

src/modules/transaction/components/TxComposer/index.js
eniolam1000752: yhea true, though that would result to code refactoring that would spill out side the scope of this PR

src/modules/transaction/components/TxComposer/index.js
eniolam1000752: Since composed fee value is done in the generic component `<TxComposer />`, that was the reason I added the extra property `isHidden` which dependent on the fact if a transaction has a messageFee then it would be enabled. Meaning for now only send token for a cross chain transfer would show this value since the value would exist in that scenario.

ManuGowda: my comment was more towards the code execution. Imagine we are doing validator registration, then the above logic would still be executed which it shouldn't? i.e `MessageFee` is only be executed for cross chain token transfer?

ManuGowda (changes requested):
Replied to few comments. I think test will still fail due to 

```
export const FEE_TYPES = {
  BYTES_FEE: 'bytesFee',
  PRIORITY_FEE: 'priorityFee',
  USER_ACCOUNT_INITIALIZATION: 'userAccountInitialization',
  ESCROW_ACCOUNT_INITIALIZATION: 'escrowAccountInitialization',
  VALIDATOR_REGISTRATION: 'validatorRegistration',
};
```

src/modules/transaction/components/TxComposer/index.js
eniolam1000752: Yhea actually that's what is happening the message fee is provided by the send form not from within the TxComposer

ManuGowda (approved):

ManuGowda:
Merging this PR as Ali reviewed and approved it offline, since he his on holiday until monday we are merging with one approval.

PR #4765, opened 2023-01-30
Fix visual issues with multisig registration

### LINKED ISSUE #4738: Issue with multi signature registration
### Expected behavior
- Dark mode on the register multi-signature form should match the figma [designs](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=635%3A169790&t=gt7kUFWjEZRDi7yN-4)
- I should be able to set a member of the multi-signature group as an optional member

- Dark mode on the register multi-signature summary modal should match the figma [designs](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=635%3A172674&t=gt7kUFWjEZRDi7yN-4)

### Actual behavior

- Dark mode on register multi-signature form has some UI defects as seen below
Pasted Graphic 14

-  I am unable to set a member as optional and I also get this error
Transaction

- Dark mode on the summary step of register multi-signature has a wrong header text color
Pasted Graphic 16

- On downloading the transaction file to be shared between members of a multi-signature account the file name is wrongly created as what is seen is this `tx-.json`

### Steps to reproduce
- Initiate a register multi signature transaction 
- The above issues should be observed on the specified transaction steps

all

ikem-legend:
### What was the problem?

This PR resolves #4738

### How was it solved?

- Updated the views according to the [design](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=635%3A169790&t=gt7kUFWjEZRDi7yN-4)

### How was it tested?

- Visually

### DIFF
--- a/src/modules/transaction/components/Multisignature/Multisignature.js
+++ b/src/modules/transaction/components/Multisignature/Multisignature.js
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { PrimaryButton, SecondaryButton } from 'src/theme/buttons';
-import { codec } from '@liskhq/lisk-client';
+import { codec, cryptography } from '@liskhq/lisk-client';
 import Illustration from 'src/modules/common/components/illustration';
 import routes from 'src/routes/routes';
 import { txStatusTypes } from '@transaction/configuration/txStatus';
@@ -87,7 +87,12 @@ const Multisignature = ({
 
   const onDownload = () => {
     const transaction = JSON.parse(transactionToJSON(transactions.signedTransaction));
-    downloadJSON(transaction, `tx-${transaction.id}`);
+    const fileSuffix =
+      transaction?.id ??
+      `registerMultisig-${cryptography.address.getLisk32AddressFromPublicKey(
+        transactions.signedTransaction.senderPublicKey
+      )}`;
+    downloadJSON(transaction, `tx-${fileSuffix}`);
   };
 
   const onSend = () => {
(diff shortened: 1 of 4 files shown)


src/modules/transaction/components/Multisignature/Multisignature.js
```diff
@@ -87,7 +87,12 @@ const Multisignature = ({
 
   const onDownload = () => {
     const transaction = JSON.parse(transactionToJSON(transactions.signedTransaction));
-    downloadJSON(transaction, `tx-${transaction.id}`);
+    const fileSuffix =
+      transaction?.id ??
```
oskarleonard, line 91: since we have this condition here wouldnt this allow us to download a broken file with `tx-undefined` as the name?

So doing something like this might make sense?
```
if(transaction?.id) {
  const fileSuffix = `registerMultisig-${cryptography.address.getLisk32AddressFromPublicKey(
        transactions.signedTransaction.senderPublicKey
      )}`;
  downloadJSON(transaction, `tx-${fileSuffix}`);
} else {
  // show error message, "Transaction id is missing" or something?
}
```

ikem-legend: @oskarleonard Transaction ID would exist if the transaction has been signed by any of the parties involved. However, the ID would not exist when a multisig registration is carried out but none of the transaction members has signed. So based on the condition, `tx-undefined` can not be a possible name because the second condition is used if ID is undefined

oskarleonard (approved):

reyraa (approved):

PR #4782, opened 2023-02-02
Add HW client and Manager

### LINKED ISSUE #4776: Create HWClient class for communicating with HW
### Description
This class isolates functionalities required to subscribe for events and execute commands on HW. This class is meant to be used by HWManager defined in #4766 as a wrapper over IPC.

### Motivation

- [ ] Establish connection through IPC
- [ ] Expose a method to subscribe to any given event on HW
- [ ] Expose a method to execute commands on HW (async)

soroushm:
### What was the problem?

This PR resolves #4776

### How was it solved?
by having a class to communicate with HW


libs/hwServer/manufacturers/ledger/index.js
```diff
@@ -25,6 +26,7 @@ const addDevice = (device, path, { add }) => {
     model: device.productName,
     path,
     manufacturer: LEDGER.name,
+    status: DEVICE_STATUS.STAND_BY,
   };
```
oskarleonard, line 27: @soroushm Do you know why we do Math.random() for `deviceId` here?
It will look pretty strange for the user if the status in the TopBar gets a new id each time they connect or disconnect their device. 

Also, to be able to complete https://github.com/LiskHQ/lisk-desktop/issues/4769,  we should add a `modelName: device.product`.

This probably need to change to: 
```
const addDevice = (deviceModel, device { add }) => {
  const newDevice = {
    productId: device.productId, // '4113' Which is the ID we show in the top bar
    modelName: device.product, // 'Nano S' Which is the model we show in the top bar
    path: device.path,
    ... (keeping the old values as well)
  };
```

oskarleonard: For reference, this is what the TransportNodeHid.listener returns
```
  TransportNodeHid.listen({
    next: (props) => {
      console.log(props);
      if (deviceModel && descriptor) {
        if (type === ADD_DEVICE) addDevice(deviceModel, descriptor, actions); // Here we need to update
        removeDevice(transport, actions);
      }
    },
  });
```
image

soroushm: @oskarleonard we can review it difrent ticket its out of the scope of current ticket

oskarleonard: Create this https://github.com/LiskHQ/lisk-desktop/issues/4799

ikem-legend (approved):

oskarleonard (approved):

PR #4786, opened 2023-02-02
Fix token transfer validation

### LINKED ISSUE #4603: Token transfer amount validation is missing
### Expected behavior

- When user enters amount greater than available balance an error message feedback should be given
- Disable `Go to confirmation` button when form has validation errors.

### Actual behavior

- User is able to enter amount value greater than available balance and not receiving error message feedback
Screenshot 2022-11-23 at 5 47 14 PM

- Amount validation is hardcoded to LSK via utils like toRawLsk. For example, INSUFFICIENT_FUNDS: { message: i18n.t('Provided amount is higher than your current balance.'), fn: () => funds < toRawLsk(numeral(value).value()), }

### Steps to reproduce

- Initiate token transfer and enter amount greater than available balance

v3

eniolam1000752:
### What was the problem?

This PR resolves #4603 

### How was it solved?

- [x] Resolve issue with token validation
- [x] Fix feedback amount and token symbol being displayed
- [x] Fix breaking unit tests

### How was it tested?

- Jest
- Visually


src/modules/transaction/hooks/useTransactionFee/utils.js
```diff
@@ -16,19 +16,26 @@ import { fromTransactionJSON } from '../../utils/encoding';
...
-    paramsSchema,
-    options
-  );
 
-  return minFee;
+  let convertedTx = {};
+  try {
+    convertedTx = fromTransactionJSON(transactionJSON, paramsSchema);
+  } catch (exp) {
+    convertedTx = {};
```
oskarleonard, line 35: Isnt it already an empty object?
Maybe initialise it with
`let convertedTx;` 

Or not mutating it at all
```
  try {
    const convertedTx = fromTransactionJSON(transactionJSON, paramsSchema);
    
    return transactions.computeMinFee(convertedTx, paramsSchema, options);
  } catch (exp) {
    return transactions.computeMinFee({}, paramsSchema, options);
  }
```

oskarleonard (dismissed):

oskarleonard (approved):

ikem-legend (approved):

PR #4795, opened 2023-02-08
Added test for HWClient

### LINKED ISSUE #4776: Create HWClient class for communicating with HW

oskarleonard:
### What was the problem?

This PR resolves #4776

### How was it solved?

By adding a test file for HWClient that test each of the functions in that class

### How was it tested?

Unit tests


oskarleonard:
Closed because these test are implemented in the source branch

PR #4796, opened 2023-02-08
Memoize hardware wallet accounts

### LINKED ISSUE #4767: Memoize HW Accounts in HWManager
### Description
This allows for easy access to the accounts stored on a hardware wallet (HW) from within the application. It works in conjunction with the HW Manager, which is a singleton class that communicates with the HW.
The HW Manager has to memoize the account list in it, HW manager will fetch list of public address from HW then it will call the utility function `utilityFunction(publicKey, accountIndex)` in order to check account is initialised and generate the account object. 
This utility function receives a `publicKey` and account Index, and uses HW info and returns the account information in the format defined below. Said utility should return account object if the account is initialized, otherwise return `null`.

### Acceptance Criteria

- Reading the accounts from the HW Manager and checking with utilityFunction for initialized 
- Storing the accounts in HWManager
- By disconnecting the HW, account list should be erased
- Merging the account names with existing names from redux, to keep user data compatible. It shouldn't have to change in redux structure
- Create and merge account object
- Add unit test

### Additional information
The HW Provider should be used in combination with the HW Manager, which should be placed at the top level of the application as a bootstrap child. The HW Provider allows for easy access to the accounts stored on the HW, by storing the accounts in the context and providing them via the useContext hook. This allows the account manager to have continuous access to the list of HW accounts 

```js

{
  "hw": {
    deviceId: "",
    model: "",
    brand: "",
  },
  "metadata": {
    "name": "masoud123",
    "pubkey": "a07e06d3d21bfcbd6e6cb158a0adc18b6ccc92db66dc8a323c7462af9c539fab",
    "path": ",
    “isHW”: true,	
    "address": "lskdgtenb76rf93bzd56cqn6ova46wfvoesbk4hnd",
    "creationTime": "2022-12-08T15:30:39.979Z"
  },
  "version": 1
}
```

ikem-legend:
### What was the problem?

This PR resolves #4767

### How was it solved?

- Implementing the memoization based on the specification

### How was it tested?

- Unit tests


src/modules/hardwareWallet/store/reducers/reducers.js
```diff
@@ -0,0 +1,28 @@
+import actionTypes from '../actions/actionTypes';
+
+const initAccounts = [];
+const initState = {
+  deviceId: 0,
+  status: 'disconnected',
```
soroushm, line 6: please remove device id and status, we will have like below and they arent part of this ticket
store.hardwarewallet = { accounts, currentDevice, devices }

src/modules/token/fungible/hooks/queries/useTokenAccountExists.test.js
```diff
@@ -0,0 +1,29 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { mockTokensAccountExists } from '@token/fungible/__fixtures__';
+import { queryWrapper as wrapper } from 'src/utils/test/queryWrapper';
+import { useTokenAccountExists } from './useTokenAccountExists';
```
soroushm, line 4: look like we don't have usage for it  and it has overlab with useCheckInitializedAccount

soroushm (changes requested):
first look

src/modules/token/fungible/hooks/queries/useTokenAccountExists.test.js
ikem-legend: @soroushm I still kept it for future use. I can remove it if you feel it should be removed

src/modules/hardwareWallet/store/reducers/reducers.js
ikem-legend: @soroushm If I did that, it would affect line 12 (`const { deviceId, status } = useSelector(selectHW);`) in `src/modules/hardwareWallet/hooks/useManageHWAccounts.js` and its related tests. So I believe it should be refactored in the new ticket

eniolam1000752 (approved):

soroushm (approved):

PR #4798, opened 2023-02-09
Transaction page should not crash

### LINKED ISSUE #4760: Transaction page should not crash
### Expected behavior
 http://0.0.0.0:8080/#/transactions should not crash

### Actual behavior
It crashes because claim rewards is not in the labels here
image

### Steps to reproduce
Go to http://0.0.0.0:8080/#/transactions

### More details
Adding this will fix the issue, but we might need to do some more logics
image

N/A

ManuGowda:
### What was the problem?

This PR resolves #4760

### How was it solved?

:bug: Add missing key value to avoid crashing

### How was it tested?

- Updated unit test

### DIFF
--- a/src/modules/transaction/utils/moduleCommand.test.js
+++ b/src/modules/transaction/utils/moduleCommand.test.js
@@ -10,24 +10,26 @@ describe('Utils: moduleCommand', () => {
   describe('getModuleCommandSenderLabel', () => {
     it('should return a dictionary of strings', () => {
       const t = jest.fn(str => str);
-      const dict = getModuleCommandSenderLabel(t);
-      const label = dict[MODULE_COMMANDS_NAME_MAP.transfer];
+      Object.values(MODULE_COMMANDS_NAME_MAP).forEach((moduleCommand) => {
+        const label = getModuleCommandSenderLabel(t)[moduleCommand];
 
-      expect(label).toBeDefined();
-      expect(typeof label).toBe('string');
-      expect(t).toHaveBeenCalled();
+        expect(label).toBeDefined();
+        expect(typeof label).toBe('string');
+        expect(t).toHaveBeenCalled();
+      });
     });
   });
 
   describe('getModuleCommandTitle', () => {
     it('should return a dictionary of strings', () => {
       const t = jest.fn(str => str);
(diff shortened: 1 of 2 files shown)


src/modules/transaction/utils/moduleCommand.test.js
```diff
@@ -10,24 +10,26 @@ describe('Utils: moduleCommand', () => {
...
-      const label = dict[MODULE_COMMANDS_NAME_MAP.transfer];
+      Object.values(MODULE_COMMANDS_NAME_MAP).forEach((moduleCommand) => {
+        const label = getModuleCommandSenderLabel(t)[moduleCommand];
 
-      expect(label).toBeDefined();
-      expect(typeof label).toBe('string');
-      expect(t).toHaveBeenCalled();
+        expect(label).toBeDefined();
+        expect(typeof label).toBe('string');
+        expect(t).toHaveBeenCalled();
```
eniolam1000752: I think we can add an extra assertion to test if t was called by the text it should be called by
```suggestion
        expect(t).toHaveBeenCalledWith(label);
```

src/modules/transaction/utils/moduleCommand.test.js
```diff
@@ -10,24 +10,26 @@ describe('Utils: moduleCommand', () => {
...
-      const label = dict[MODULE_COMMANDS_NAME_MAP.transfer];
+      Object.values(MODULE_COMMANDS_NAME_MAP).forEach((moduleCommand) => {
+        const label = getModuleCommandTitle(t)[moduleCommand];
 
-      expect(label).toBeDefined();
-      expect(typeof label).toBe('string');
-      expect(t).toHaveBeenCalled();
+        expect(label).toBeDefined();
+        expect(typeof label).toBe('string');
+        expect(t).toHaveBeenCalled();
```
eniolam1000752: I think we can add an extra assertion to test if t was called by the text it should be called by
```suggestion
        expect(t).toHaveBeenCalledWith(label);
```

eniolam1000752 (changes requested):

oskarleonard (dismissed):

eniolam1000752 (approved):

oskarleonard (approved):

PR #4805, opened 2023-02-13
Create redux state for hardware wallet

### LINKED ISSUE #4777: Create a hook to use HWClient and update Redux accordingly
### Description
We need a React hook which uses HWClient to subscribe to events coming from HW, and maintain relevant information on the Redux store, such as list of accounts, list of connected devices, the status of the connected HW.

### Acceptance criteria
- [ ] Displaying toast messages when the HW is connected or disconnected
- [ ] Display a toast with an option to navigate to the switch device modal if more than one HW is connected.

oskarleonard:
### What was the problem?

This PR resolves #4777

### How was it solved?
- [x] Displaying toast messages when the HW is connected or disconnected
- [x] Display a toast with an option to navigate to the switch device modal if more than one HW is connected.
- [x] Add a React hook listener which populates redux with events coming from hardware server
- [x] Added a skeleton modal for switching hardware wallets. Which will be completed in https://github.com/LiskHQ/lisk-desktop/issues/4770

### How was it tested?
Unit tests and manually


libs/hwServer/index.js
```diff
@@ -111,7 +111,10 @@ export class HwServer {
     const { sender } = this.pubSub;
     this.devices.push(device);
     this.syncDevices();
-    publish(sender, { event: IPC_MESSAGES.HW_CONNECTED, payload: { model: device.model } });
+    publish(sender, {
+      event: IPC_MESSAGES.HW_CONNECTED,
+      payload: { model: device.model, devices: this.devices },
```
soroushm, line 116: we have difrent even  for it called "DEVICE_LIST_CHANGED" please avoid pass hole devices from here

src/modules/hardwareWallet/hooks/useHwListener.js
```diff
@@ -0,0 +1,28 @@
...
+
+function useHwListener() {
+  const dispatch = useDispatch();
+
+  const { ipc } = window;
+
+  if (!ipc) return;
+
+  useEffect(() => {
+    ipc.on(DEVICE_LIST_CHANGED, (action, data) => {
```
soroushm, line 19: please use hwManager instade of direct ipc call

```suggestion
hwManager.subscribe(DEVICE_LIST_CHANGED, (action, data) => {
```

soroushm (changes requested):

src/modules/hardwareWallet/hooks/useHwListener.js
oskarleonard: @soroushm we dont have that method on the hwManager. Do you mean that i should use:
 `HwClient.subscribe(`

or should i create that function on the HwManager?

libs/hwServer/index.js
oskarleonard: I added this here so we can show a toast for switching hardware devices when a new device is **connected** and making it +1 devices. 

Or do you want me to use the DEVICE_LIST_CHANGED for showing that specific toast (switch active hardware device)? The drawback of using this event for that would be that we will in that case also show the toast (switch active hardware device) when **disconnecting** a device. 

Or did you have something else in mind?

soroushm: then we can just send length of the device list is enough, by the way we can even show the toast by just listning to DEVICE_LIST_CHANGED

src/modules/hardwareWallet/hooks/useHwListener.js
soroushm: hwManager extended hwClient so  subscirbe is included in hwmanager 
HWManager is a singlton class witch hwClient is not so we need to use hwManager

oskarleonard: Oh, yes that is true, will change so we are consistent on how we talk to the ledger

soroushm (dismissed):
LGTM

src/modules/hardwareWallet/store/selectors/hwSelectors.js
```diff
@@ -0,0 +1,3 @@
+export const selectHardwareDevices = (state) => state.hardwareWallet.devices;
+export const selectActiveHardwareDeviceId = (state) => state.hardwareWallet.currentDevice.deviceId;
+export const selectActiveHardwareDevice = (state) => state.hardwareWallet.currentDevice;
```
ikem-legend: @oskarleonard There's already a single file of selectors in `src/redux/selectors.js`. I think these can be added there

ikem-legend (approved):

src/modules/hardwareWallet/store/selectors/hwSelectors.js
oskarleonard: Fixed by moving the global hw selectors to the hw module

soroushm (approved):

Issue #4708, opened 2023-01-11
Adapt is default application with metadata service


soroushm:
### Description
Since the service behaviour is changed and isDefault argument is now part of the application meta data, this change needs to update

### Acceptance Criteria
- BaseUrl of meta service should be configured in the env
- Split MetaData Client from application Client in order to always call meta data from the ENV BaseURL
- First call `/api/v3/blockchain/apps` with query string of isDefault: true and network Mainnet 
- Fetch the rest of information from `api/v3/blockchain` and merge it by using `transformResult` from Client
- Select first index default application(LISK) check for application status  and  set the current application with Lisk
- These logic should be part of custom hooks and used in [bootstrap](https://github.com/LiskHQ/lisk-desktop/issues/4707) component 

### Additional Information
#4707

ManuGowda:
With regards to environment configurations, we should consider the following:

Lisk supports following networks, devnet, alphanet, betanet, testnet and mainnet. for all of these Lisk Service provides following URLs:

https://devnet-service.lisk.com
https://alphanet-service.lisk.com
https://betanet-service.lisk.com
https://testnet-service.lisk.com
https://mainnet-service.lisk.com

Based on the above base URL we can fetch application metadata and corresponding application specific service URLs.

So in env config file, we should use appropriate network name e.g `devnet` during development, `alphanet` once we tag alpha and release it for testing, `betanet` once we tag beta and release it for testing, etc..

PR #4815, opened 2023-02-15
Implement application bootstrap component

### LINKED ISSUE #4707: Application Bootstrap Loading State
### Description
At very beginning, mounting application wallet need to have loading state in order to fetch default application and create an instance of client with default application

### Acceptance Criteria
- Create a Bootstrap Component and Wrap entire routing with bootstrap component
- Bootstrap should only render Props.children
- In Loading State, children shouldn't mount till loading is over
- In case apis get failed, it should show retry stage and user are not allowed to pass this stage
- Add Unit Test

eniolam1000752:
### What was the problem?

This PR resolves #4707 and #4708

### How was it solved?

- [x] Implement application bootstrap wrapper
- [x] update order of query hook call (i.e app meta data first then applications on-chain data)
- [x] update unit tests

### How was it tested?

- Jest
- Visually


setup/config/webpack.config.dev.js
```diff
@@ -27,6 +27,7 @@ const config = {
     new webpack.DefinePlugin({
```
oskarleonard: **Comment not related to this file**

Question: This seems to be unrelated to hardware wallet? Shouldnt we merge this into development?

src/modules/blockchainApplication/manage/store/reducer.js
```diff
@@ -36,20 +36,32 @@ export const pins = (state = initialState.pins, { type, chainId }) => {
 /**
  *
  * @param {Object} state
- * @param {type: String, data: Object} action
+ * @param {type: String, app: Object, apps: Object, chainId: String} action
  */
-export const applications = (state = initialState.applications, { type, application, chainId }) => {
+export const applications = (state = initialState.applications, { type, app, apps, chainId }) => {
```
oskarleonard, line 41: It is not advisable [to mutate the redux state ](https://redux.js.org/faq/immutable-data#why-does-a-reducer-mutating-the-state-prevent-react-redux-from-re-rendering-a-wrapped-component) this way. This whole reducer smells a bit to me, dont we have any noticeable bugs on the components utilising this state? 🤔

src/modules/blockchainApplication/manage/store/reducer.js
```diff
@@ -36,20 +36,32 @@ export const pins = (state = initialState.pins, { type, chainId }) => {
...
+      if (app.chainID in state) {
+        state[app.chainID].serviceURLs.push(app.serviceURLs);
       } else {
-        state[application.chainID] = application;
+        state[app.chainID] = app;
       }
       return state;
 
+    case actionTypes.setApplications: {
+      apps.forEach((application) => {
```
oskarleonard: Couldnt this be written in an immutable way with just a `.map(`?

src/modules/blockchainApplication/manage/store/reducer.js
eniolam1000752: hmmm nope we don't so we have refactored reducer handlers and all suitable actions to adjust to this

src/modules/blockchainApplication/manage/store/reducer.js
eniolam1000752: I guess so but I think just iterating through the array list and re-consuming the checks that are done when setting a single application was way easier.

setup/config/webpack.config.dev.js
```diff
@@ -27,6 +27,7 @@ const config = {
...
       PRODUCTION: JSON.stringify(false),
       VERSION: JSON.stringify(version),
+      DEFAULT_NETWORK: 'devnet',
```
soroushm: no iots wront its should be in .env
and need a prefix of REACT_APP like the REACT_APP_MSW you can find in config file
REACT_APP_ DEFAULT_NETWORK

setup/react/app/ApplicationBootstrap.js
```diff
@@ -0,0 +1,43 @@
...
+
+  useEffect(() => {
+    if (defaultApps.length && isFetched) {
+      setCurrentApplication(defaultApps[0]);
+      setApplications(defaultApps);
+    }
+  }, [isFetched]);
+
+  if (error) {
+    // @TODO: this return should be replaced with an actual error message page
```
soroushm, line 31: create a ticket if is not exist and add ticket number to todo

src/modules/blockchainApplication/manage/store/reducer.js
soroushm: @oskarleonard good catch 

@eniolam1000752 redux is imotable we shoulnt manipulate the state in this way we need to return a new object like below

```js
 if (app.chainID in state) {
cons newState = {
 ...state
}
newState[app.chainID].serviceURLs.push(app.serviceURLs)
       return newState
      } else {
       return {
        ...state,
        [app.chainID]: app;
        }
      }
```

soroushm: more info here
https://stackoverflow.com/questions/43759080/what-could-happen-if-modifying-state-directly-inside-a-redux-reducer

soroushm (changes requested):

setup/config/webpack.config.dev.js
soroushm: @eniola its wrong env you can have .env file in your pc for test propose check this out
https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used

soroushm (dismissed):

oskarleonard (dismissed):

soroushm (approved):

oskarleonard (approved):

PR #4823, opened 2023-02-21
Add hardware wallet status to HWServer

### LINKED ISSUE #4803: Store active HW status in HWServer
### Description

At the moment we store the HWServer on a per device basis. But in practice we only consume the status of the active device. Since devices do not emit events when they change to standby status, two issues my occur
- The status of inactive devices can fall our of sync
- The status of devices can not be updated automatically.

Therefore we should resort to maintaining the status of the active device and providing the require facilities to update the status per request, e.g. from the React app.

### Motivation
- Store the status of the active device on HWServer
- Remove the status from all devices
- Provide required methods on HWManager to retrieve the active device status from HWServer.
- Add unit test coverage

soroushm:
### What was the problem?

This PR resolves #4803

### How was it solved?
by adding status in HW server

### How was it tested?
unit test


libs/hwServer/manufacturers/ledger/index.js
```diff
@@ -72,26 +71,28 @@ const getLedgerAccount = (index = 0) => {
...
   try {
-    transport = await transporter.open(device.path);
+    transport = await transporter.open(path);
     const liskLedger = new LiskApp(transport);
-    const ledgerAccount = getLedgerAccount();
-    const account = await liskLedger.getAddressAndPubKey(ledgerAccount.derivePath());
-    device.status = account ? DEVICE_STATUS.CONNECTED : DEVICE_STATUS.DISCONNECTED;
+    const ledgerAccount = getLedgerAccount(0);
+    const hasAccount = await liskLedger.getAddressAndPubKey(ledgerAccount.derivePath());
+    status = hasAccount ? DEVICE_STATUS.CONNECTED : DEVICE_STATUS.STAND_BY;
```
oskarleonard: We probably need to do something like this: 
`status = hasAccount.address ? DEVICE_STATUS.CONNECTED : DEVICE_STATUS.STAND_BY;`
or
`status = hasAccount.error_message === 'No errors' ? DEVICE_STATUS.CONNECTED : DEVICE_STATUS.STAND_BY;`

soroushm: but there is no error, (i think while the device path is not exist its return error nothing to do with firmware is connected or not) we keep the same logic as we had in past

oskarleonard: `await liskLedger.getAddressAndPubKey(ledgerAccount.derivePath());`

If inside lisk app it will return
```
{
  pubKey: '6cea247907980c412ec1b0f51957f7bd50391361a8ab2408b584f182b89e5f32',
  address: 'lsk3hm2t58ze3gs848hsbjca5dc2nc7uqedqdfehm',
  return_code: 36864,
  error_message: 'No errors'
}
```
If  ledger is open but **not** inside lisk app it will return:
```
{
  returnCode: 28161,
  errorMessage: 'Unknown Status Code: 28161',
  return_code: 28161,
  error_message: 'Unknown Status Code: 28161'
}
```

If ledger is disconnected it will throw (i think)

oskarleonard: So since `hasAccount` is an object in both these states: 
1.  Inside lisk app. 
2.  Not inside lisk app  

standby will never be true

libs/hwServer/index.js
```diff
@@ -34,7 +39,8 @@ export class HwServer {
     subscribe(receiver, {
       event: IPC_MESSAGES.HW_COMMAND,
       action: async ({ action, data }) => {
-        const device = this.getDeviceById(this.currentDeviceId);
+        const device = this.getCurrentDevice();
+        console.log('device', action, device)
```
ikem-legend: @soroushm Please remove console log

libs/hwServer/constants.js
```diff
@@ -4,7 +4,7 @@ export const REQUEST = 'request';
 export const PIN = 'pin'; // TODO: Remove when trezor logic is removed
 export const PASSPHRASE = 'passphrase';
 export const IPC_MESSAGES = {
-  CHECK_LEDGER: 'checkLedger',
+  CHECK_LEDGER: 'checkStatus',
```
ikem-legend: @soroushm Shouldn't the object key be renamed to `CHECK_STATUS`?

libs/hwServer/constants.js
```diff
@@ -47,7 +47,7 @@ export const FUNCTION_TYPES = {
 };
 
 export const METHOD_NAMES = {
-  [IPC_MESSAGES.CHECK_LEDGER]: 'checkLedger',
+  [IPC_MESSAGES.CHECK_LEDGER]: 'checkStatus',
```
ikem-legend: @soroushm I believe the object key here should also be renamed to `IPC_MESSAGES.CHECK_STATUS`

src/modules/account/components/ManageAccounts/ManageAccounts.js
```diff
@@ -30,8 +31,7 @@ export const ManageAccountsContent = ({
   const title = customTitle ?? t('Manage accounts');
   const { accounts: hwAccounts } = useHWAccounts();
 
-  // @TODO: actual status should be replaced when the useHWStatus hook is integrated by issue #4768
-  const status = DEVICE_STATUS.STAND_BY;
+  const {status} = useHWStatus();
```
ikem-legend: @soroushm This line should be formatted

libs/hwServer/constants.js
soroushm: i just avoid to change it becouse of the conflict but as must the pr already merged its changed

oskarleonard (approved):

ikem-legend (approved):

PR #4828, opened 2023-02-21
Fix select application's service URL and current app logo

### LINKED ISSUE #4818: Choose application URL shows invalid service URLs
### Expected behavior

- Show valid service URLs provided by service http://165.227.246.146:9901/api/v3/blockchain/apps/meta?limit=20&network=devnet&chainID=04000001
- Render correct logo for the selected application

### Actual behavior

Screenshot 2023-02-15 at 11 45 00 AM

### Steps to reproduce

- Select application dropdown to switch to different application

3.0

eniolam1000752:
### What was the problem?

This PR resolves #4818

### How was it solved?

- [x] Refactor current chain logo to use app metadata logo
- [x] Fix crash on add application(this is so that I can test if adding application doesn't exhibit unnecessary behaviour on the serviceUrls list
- [x] Fix applications reducer so that it sets serviceUrl not append for an existing application in the redux state
- [x] Resolve failing unit tests.

### How was it tested?

- Jest
- Visually


ikem-legend (dismissed):

src/modules/blockchainApplication/manage/store/reducer.js
```diff
@@ -44,11 +44,15 @@ export const applications = (state = initialState.applications, { type, app, app
       // In cases where a new node for an existing application is being added,
       // the new node url should be appended to the apis array of the application
       if (app.chainID in state) {
-        state[app.chainID].serviceURLs.push(app.serviceURLs);
-      } else {
-        state[app.chainID] = app;
+        return {
+          ...state,
+          [app.chainID]: {
+            serviceURLs: app.serviceURLs,
```
oskarleonard: Do we really want to replace the whole application object with just one serviceUrl? Probably this is what we want?
This will keep it.
```
      if (app.chainID in state) {
        return {
          ...state,
          [app.chainID]: {
            ...state[app.chainID], // We still want the rest of the fields of an application, right?
            serviceURLs: app.serviceURLs,
          },
        };
      }
```

But why do we need this if statement at all? couldnt we just replace the whole application object with the new one?
Just having:
`return { ...state, [app.chainID]: app };`

or does `app` in this context only contain `serviceURLs`?

setup/react/app/ApplicationBootstrap.js
```diff
@@ -17,13 +17,15 @@ const ApplicationBootstrap = ({ children }) => {
   } = useGetDefaultApplication();
   const { setApplications } = useApplicationManagement();
   const [, setCurrentApplication] = useCurrentApplication();
+  const [isFirstTimeLoading, setIsFirstTimeLoading] = useState(true);
```
oskarleonard, line 20: why do we need this one?

eniolam1000752: so what happens is, when we change application the UI flickers because isLoading would now be set to true by the `useGetDefaultApplication` hook. So I did this so we can always track first time boastrap of the application.

src/modules/blockchainApplication/manage/store/reducer.js
eniolam1000752: true actually I guess we can just replace the whole application

oskarleonard (dismissed):

oskarleonard (dismissed):

ikem-legend:
@eniolam1000752 The actions, actionTypes and reducers for `setApplicationNode` need to be removed too since the `currentNode` has been removed

soroushm (dismissed):

src/modules/blockchainApplication/manage/store/actionTypes.js
```diff
@@ -3,8 +3,6 @@ const actionTypes = {
   addApplicationByChainId: 'ADD_APPLICATION_BY_CHAIN_ID',
   deleteApplicationByChainId: 'DELETE_APPLICATION_BY_CHAIN_ID',
   setCurrentApplication: 'SET_CURRENT_APPLICATION',
-  setApplicationNode: 'SET_APPLICATION_NODE',
-  setApplications: 'SET_APPLICATIONS',
```
ikem-legend, line 7: @eniolam1000752 This (`setApplications`) seems to have been accidentally removed and should be reverted

soroushm (approved):

soroushm (approved):
LGTM

ikem-legend (approved):

PR #4833, opened 2023-02-27
Setting options.enabled to false shouldnt  be overridden in useTransactions

### LINKED ISSUE #4826: Search dropdown shouldn't always be open
### Expected behavior
Setting `options.enabled` to `false` should override everything. For example:

Using useTransactions hook like this:
```
  const transactions = useTransactions({
    config: { params: { transactionID: search } },
    options: { enabled: isTxId },
  });
```

Should keep `options.enabled` in the `useCustomInfiniteQuery` hook.

### Actual behavior
This code strips away the `options.enabled` completely.
```
  const response = useCustomInfiniteQuery({
    keys,
    config,
    options: { ...options, enabled: !!(client.socket || client.http) },
  });
```

Which results in the search dropdown always being open
image

### Suggested fix
```
  const isEnabled = options?.enabled !== false && !!(client.socket || client.http);
  const response = useCustomInfiniteQuery({
    keys,
    config,
    options: { ...options, enabled: isEnabled },
  });
```

oskarleonard:
### What was the problem?

This PR resolves #4826

### How was it solved?

- Collect enabled logic in a scoped function
- Look for options?.enabled false as the top condition to pass
- getUpdate enabled logic should be scoped. 

### How was it tested?

Manually


ManuGowda (approved):

src/modules/transaction/hooks/queries/useTransactions.js
```diff
@@ -83,10 +83,17 @@ export const useTransactions = ({
     // @todo invalid this transaction by specific unique query with config
   }, [queryClient, setHasUpdate]);
 
+  function getIsEnabled() {
+    const isDisabled = options?.enabled === false;
```
eniolam1000752, line 87: can't we do this

```suggestion
    const isDisabled = !options?.enabled;
```

eniolam1000752 (changes requested):

PR #4836, opened 2023-02-28
Use denomUnits to calculate correct radix 

### LINKED ISSUE #4753: Use denomUnits to calculate correct radix
### Description

- With Lisk supporting different denom for different chains we should create a common util function to calculate this.
- We also need to update all the places where we use hardcoded value for this. 

### Motivation

To make sure we are using the correct denom for amounts.

### Additional Information
The information for the correct radix is found in 
`GET /blockchain/apps/meta/tokens`.
`tokenId` is now provided in these responses and can be use to get the correct token. 

When we have the token we could do something like:
```
  const denom = denomUnits?.find((denomUnit) => denomUnit.denom === displayDenom);
```

### Additional information

https://github.com/LiskHQ/lisk-sdk/blob/fa63da67112d29cc1c42479a2a9be56b49675f05/elements/lisk-transactions/src/format.ts#L38

eniolam1000752:
### What was the problem?

This PR resolves #4753

### How was it solved?

- [x] Introduce `convertFromRawDenom` and `convertToRawDenom` util function
- [x] Replace all occurrences of `fromRawLsk` to `convertFromRawDenom` and `toRawLsk` to `convertToRawDenom`
- [x] Fix unit tests

### How was it tested?

- Jest
- visually


src/modules/pos/validator/hooks/usePosToken.js
```diff
@@ -0,0 +1,21 @@
+import { useMemo } from 'react';
+import { useCurrentAccount } from 'src/modules/account/hooks';
+import { useTokensBalance } from 'src/modules/token/fungible/hooks/queries';
+import { usePosConstants } from './queries';
+
+const usePosToken = (props) => {
+  const [currentAccount] = useCurrentAccount();
+  const { address } = props?.address || currentAccount.metadata || {};
```
ikem-legend: @eniolam1000752 This seems to suggest that there might be a `props.address.address` which doesn't sound likely

src/modules/pos/validator/components/RegisterValidatorForm/index.test.js
```diff
@@ -4,19 +4,17 @@ import { useTokensBalance } from '@token/fungible/hooks/queries';
...
 import RegisterValidatorForm from '.';
+import { usePosConstants } from '../../hooks/queries';
 
 const mockCurrentAccount = mockSavedAccounts[0];
 
-jest.mock('../../hooks/queries', () => ({
-  ...jest.requireActual(),
-  usePosConstants: jest.fn().mockReturnValue({ data: mockPosConstants }),
-}));
+jest.mock('../../hooks/queries');
```
ikem-legend: @eniolam1000752 The specific file could have been mocked rather than the entire hooks directory

src/modules/pos/validator/components/StakeForm/StakeRow.js
```diff
@@ -1,10 +1,10 @@
+/* eslint-disable max-statements */
```
ikem-legend: @eniolam1000752 This should be removed as it is invalid in the file

src/modules/token/fungible/components/BalanceInfo/index.js
```diff
@@ -11,6 +11,7 @@ import ActionBar from './ActionBar';
 import styles from './BalanceInfo.css';
 
 // eslint-disable-next-line complexity
+// @Todo: this component needs to be removed cleaningly because it seems it not used in the application
```
ikem-legend: @eniolam1000752 Typo at `cleaningly`. Please fix

src/modules/token/fungible/utils/lsk.js
```diff
@@ -41,6 +41,45 @@ export const toRawLsk = (value) => {
...
+/**
+ * Converts from denom denoted by the token symobl to a smaller unit
+ *
+ * @param {BigNumber|number} amount - Amount value to be converted
+ * @param {BigNumber|number} token - Token value merged with its equivalent metadata
+ * @returns {BigNumber} Amount value converted from the token symbol's denom
+ */
+export const convertToBaseDenom = (amount, token = {}) => {
+  // eslint-disable-next-line no-restricted-globals
+  if (isNaN(amount)) return '0';
```
ikem-legend, line 74: @eniolam1000752 `Number.isNaN` should be used here

src/modules/transaction/utils/moduleCommand.js
```diff
@@ -11,12 +11,15 @@ const getModuleCommandSenderLabel = (t = str => str) => ({
   [MODULE_COMMANDS_NAME_MAP.reportValidatorMisbehavior]: t('Reporter'),
   [MODULE_COMMANDS_NAME_MAP.claimRewards]: t('Claim rewards'),
   [MODULE_COMMANDS_NAME_MAP.changeCommission]: t('Edit commission'),
+  [MODULE_COMMANDS_NAME_MAP.registerKeys]: t('Register keys'),
+  [MODULE_COMMANDS_NAME_MAP.registerSidechain]: t(''),
```
ikem-legend: @eniolam1000752 Was this intentionally left empty?

src/modules/wallet/components/RegisterMultisigSummary/Summary.js
```diff
@@ -76,7 +76,7 @@ const Summary = ({
       nextStep({ formProps, transactionJSON, sender }, 2);
     }
   }, [transactions.signedTransaction, transactions.txSignatureError]);
-
+
```
ikem-legend: @eniolam1000752 Empty space should be removed

src/modules/blockchainApplication/explore/components/BlockchainApplicationRow/BlockchainApplicationRow.js
```diff
@@ -4,13 +4,14 @@ import { usePinBlockchainApplication } from '@blockchainApplication/manage/hooks
...
-import TokenAmount from 'src/modules/token/fungible/components/tokenAmount';
+import FormattedNumber from 'src/modules/common/components/FormattedNumber';
+import { fromRawLsk } from '@token/fungible/utils/lsk';
 import liskLogo from '../../../../../../setup/react/assets/images/LISK.png';
 import styles from './BlockchainApplicationRow.css';
 
 const DepositAmount = ({ amount }) => (
   <span className={`deposit-amount ${styles.amount} ${grid['col-xs-3']}`}>
-    <TokenAmount val={amount} token="LSK" />
+    <FormattedNumber val={fromRawLsk(amount)} /> LSK
```
reyraa: In some places you are using TokenAmount and in others you use FormattedNumber with LSK next to it. TokenAmount is essentially FormattedNumber + token symbol.
Why do you do such a change?

reyraa: this is done is so many places, I'm not going to commend all of them, you can take a  look yourself

src/modules/blockchainApplication/manage/components/RemoveApplicationDetails/RemoveApplicationDetails.js
```diff
@@ -120,34 +117,28 @@ const RemoveApplicationDetails = ({
             <span>{t('Deposited:')}</span>
             {/* TODO: this is a placeholder value pending when its part of service response */}
             <span>
-              <TokenAmount val={deposit} />
-              {' '}
-              LSK
+              <FormattedNumber val={fromRawLsk(deposit)} /> LSK
```
reyraa: Formatted number accepts token symbol and performs this job.

src/modules/common/components/amountField/index.js
```diff
@@ -76,7 +77,7 @@ const AmountField = ({
   useEffect(() => {
     if (isMaximum) {
       const value = formatAmountBasedOnLocale({
-        value: fromRawLsk(maxAmount.value),
+        value: convertFromBaseDenom(maxAmount.value, token),
```
reyraa, line 80: There's still 19 more files where we use `fromRawLsk ` and 3 more files where we use `toRawLsk`. Should we replace these too?

reyraa (changes requested):

src/modules/token/fungible/utils/lsk.js
eniolam1000752: `Number.isNaN` would not work here because `Number.isNaN` would return `false` even though `abc` is provided which in this use case its wrong (Number.isNaN tests that the provided parameter can be represented by an equivalent numeric value) and what I want to test is to ensure that the string value provided is a number and window.isNaN does that

https://stackoverflow.com/questions/25176459/is-number-isnan-more-broken-than-isnan

src/modules/blockchainApplication/explore/components/BlockchainApplicationRow/BlockchainApplicationRow.js
eniolam1000752: yhea so reason was because in some places , we need to only render LSK amount in-regardless of the chain we are in. but since TokenAmount now takes in token as a token object to render its values, I kind of did this.

But i guess I can just extend TokenAmount to be able to render LSK types of amounts

src/modules/common/components/amountField/index.js
eniolam1000752: hmm i guess those are in the test files okay I would take a look

reyraa (dismissed):
Good job

ikem-legend (dismissed):

ikem-legend (approved):

reyraa (approved):

PR #4841, opened 2023-03-01
Refactor transaction fee calculation

### LINKED ISSUE #4792: Integrate invoke hook to get CCM fees
### Description

Presently, we already have a mock implementation of the `useGetMessage` as address by this [issue](https://github.com/liskhq/lisk-desktop/issues/4521#issue-1416407770), but this is without integrating the invoke hook. We would need to integrate the invoke hook which would be invoking a command to get the CCM fee once its exposed by [sdk](https://github.com/LiskHQ/lisk-sdk/issues/8114)

### Motivation

`useGetMessage` needs to integrate the actual invoke command to get the CCM fee value. Presently it doesn't.

ManuGowda:
### What was the problem?

This PR resolves #4792

### How was it solved?

- Integrate invoke hook to get CCM fees 
- Use token_hasUserAccount to know account initialisation 
- Refactor useGetInitializationFees hook
- Implement useGetMinimumMessageFee hook

### How was it tested?

- Update unit tests
- Added unit tests


src/modules/token/fungible/hooks/queries/useGetHasUserAccount.js
```diff
@@ -0,0 +1,20 @@
+import { useInvokeQuery } from '@common/hooks';
+
+export const useGetHasUserAccount = ({ options = {}, config: customConfig = {}, } = {}) => {
+  const config = {
+    data: {
+      endpoint: 'token_hasUserAccount',
+      params: {
```
oskarleonard, line 7: To not deviate from our a bit strange query hook the pattern, like we do in some places, which have caused bugs (for example the search being open all the time). 

We should do:
```
    data: {
      endpoint: 'token_hasUserAccount',
      ...customConfig
    },
```

src/modules/network/hooks/useCommandsSchema.js
```diff
@@ -15,5 +15,16 @@ export const useCommandSchema = () => {
     );
   }, [data]);
 
-  return { moduleCommandSchemas, ...rest };
+  const messagesSchemas = useMemo(() => {
```
oskarleonard, line 18: The only thing these memos prevent is if `... rest` changes, ie, `isLoading, isFetching`. This is good to be aware of since it could cause unintended behaviour. 

if `data.commands` or actually `data.x` changes, this `useMemo` will still recalculate.

To fix that one would need to do:
```
  const messagesSchemas = React.useMemo(() => {
    // calcualte
  }, [data?.data?.messages]);
```
But even then, if we fetches and `data.data.messages` gets the same array, it would still recalculate. To prevent that one would need to use `JSON.stringify(` or `.length` since react only shallow compare (primitive values):

```
  const messagesSchemas = React.useMemo(() => {
    // calcualte
  }, [JSON.stringify(data.data.messages)]); // No advisable if not a heavy heavy calculation is performed inside
  OR
    }, [data.data.messages.length]); // Which is probably not what we want.
```

As you probably can tell, i am not a big fan of premature optimisation 😆

oskarleonard (approved):

eniolam1000752 (approved):

Issue #4537, opened 2022-10-20
Refactor private route rerouting logic


ikem-legend:
### Description

There is a functionality for handling redirect to private routes for unauthenticated accounts as shown below:
```<Redirect to={`${routes.manageAccounts.path}?referrer=${path.replace(
          /\/(send|vote)/,
          ''
        )}&${search.replace(/^\?/, '')}`}
      />```

The current implementation is rigid as it is tied to the send and vote modals but the implementation can be made more flexible for any other private route, based on the Redirect API

eniolam1000752:
We need to also consider redirection of modals that are to be private (i.e only show when an account is selected).
@soroushm

PR #4848, opened 2023-03-06
Fix referrer on private routing

### LINKED ISSUE #4531: Adapt private route referer with account managment
### Description
In some cases while the route is private (user not selected), user will redirect to the account management page by add/select account other than going to the dashboard user should navigate back to page came from, referer query params have the information

### Acceptance Criteria
- Navigate user to the account management instead of login page in private route with referer query params of current url
- Navigate user to the account management while they try to open modal witch is private referer query params of current url
- On select/add account navigate them back to the page they came from
- Adopt unit test

eniolam1000752:
### What was the problem?

This PR resolves #4531 and #4537

### How was it solved?

- [x] Introduce referrer on customRoute component when wallet has no account 
- [x] refactor customRoute regex logic to be more generic
- [x] Fix failing unit tests

### How was it tested?

- Visually
- Jest


src/modules/common/components/customRoute/index.js
```diff
@@ -26,20 +26,14 @@ const CustomRoute = ({ path, exact, isPrivate, forbiddenTokens, component, t, hi
     return <Redirect to={`${routes.dashboard.path}`} />;
   }
 
-  if (!isAuthenticated && path === routes.manageAccounts.path && !(accounts.length > 0)) {
-    history.replace(routes.addAccountOptions.path);
+  if (!isAuthenticated && path === routes.manageAccounts.path && accounts.length === 0) {
+    history.replace({ pathname: routes.addAccountOptions.path, search });
```
oskarleonard, line 30: How come we are not using `return <Redirect ... />` component here as well?

oskarleonard (approved):

ikem-legend (approved):

PR #4849, opened 2023-03-07
Update missing module command mappings

### LINKED ISSUE #4814: Update missing module command mappings
### Expected behavior

- Add all known `module` `command` mappings
- Handle `unknown` `module` `command` mappings
- Application shouldn't crash
- We could use `/api/v3/network/status` `moduleCommands` for handling the mapping dynamically

### Actual behavior

- Wallet is unable to render dashboard or transactions table when an `unknown module command` response needs to be processed
Screenshot 2023-02-14 at 7 14 08 PM

### Steps to reproduce

- Submit any transactions whos mapping doesn't exists e.g `Legacy RegisterKeys`
- Observe the dashboard crashes

3.0

oskarleonard:
### What was the problem?

This PR resolves #4814

### How was it solved?
- Adding missing module commands

### How was it tested?
Manually


src/modules/common/__fixtures__/mockCommandParametersSchemas.js
```diff
@@ -2389,54 +2389,6 @@ export const mockCommandParametersSchemas = {
           },
         },
       },
-      {
```
ManuGowda, line 2392: why is this removed?

oskarleonard: isnt that renamed to `transferCrossChain` ?

ManuGowda (approved):

PR #4872, opened 2023-03-23
Do not show the isLoading icon when a search has completed

### LINKED ISSUE #4617: Do not show the isLoading icon when a search has completed
### Expected behavior
Do not show the isLoading icon when a search has completed

### Actual behavior
image

### Steps to reproduce
- Search for something in the top menu search bar
- Watch as the isLoading icons spins forever

v3

### More info
Relevant thread with more info and with possible solutions: https://github.com/TanStack/query/issues/3584

oskarleonard:
### What was the problem?

This PR resolves #4617

### How was it solved?

By checking both for queryRes.isLoading && queryRes.isFetching to return a loading spinner state to searchbar

### How was it tested?

Manually

### DIFF
--- a/src/modules/search/hooks/useSearch.js
+++ b/src/modules/search/hooks/useSearch.js
@@ -5,7 +5,6 @@ import { useValidators } from '@pos/validator/hooks/queries';
 import { useBlocks } from '@block/hooks/queries/useBlocks';
 
 function getIsLoading(queryRes) {
-  // TODO: Function to be replaced with isInitialLoading when upgrading react-query to +v4.26
   return queryRes.isLoading && queryRes.isFetching;
 }
 
--- a/src/modules/search/hooks/useSearch.js
+++ b/src/modules/search/hooks/useSearch.js
@@ -4,6 +4,11 @@ import { useTransactions } from '@transaction/hooks/queries';
 import { useValidators } from '@pos/validator/hooks/queries';
 import { useBlocks } from '@block/hooks/queries/useBlocks';
 
+function getIsLoading(queryRes) {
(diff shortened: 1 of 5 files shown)


ManuGowda (approved):

src/modules/search/hooks/useSearch.js
```diff
@@ -4,6 +4,10 @@ import { useTransactions } from '@transaction/hooks/queries';
 import { useValidators } from '@pos/validator/hooks/queries';
 import { useBlocks } from '@block/hooks/queries/useBlocks';
 
+function getIsLoading(queryRes) {
+  return queryRes.isLoading && queryRes.isFetching;
```
eniolam1000752, line 8: shouldn't this boolean logic be an `OR` i.e 

```suggestion
  return queryRes.isLoading || queryRes.isFetching;
```

since if isLoading is true or isFetching is true, it means the entire query is still in a pending state

eniolam1000752 (changes requested):

src/modules/search/hooks/useSearch.js
oskarleonard: What we want to do it to bring over the `isInitialLoading` logic which was introduced in v4.2 (we are still on v4.0) which keeps the isLoading state for spinners which were apparent in v3. 

This comment from the RQ author:
https://github.com/TanStack/query/issues/3584#issuecomment-1262638656

And this is how he implemented `isInitialLoading` 
https://github.com/TanStack/query/pull/4244/files#diff-144cf89d34f29ea1e30da14ec1cd6c323b645d0d446435aa1373b7d0fa00ccbc

eniolam1000752 (approved):

PR #4874, opened 2023-03-23
Fix explorable and managed application

### LINKED ISSUE #4817: Non active blockchain applications are listed in cross chain transfer
### Expected behavior

  - Users should view the actual state i.e inactive, active, terminated in application list
  - Users should not see applications which are inactive or terminated while sending token cross chain
  - Users should able to switch between blockchains on select without issues
 
### Actual behavior

- Currently the blockchain applications do not take state (inactive, active, terminated) into account

### Steps to reproduce

- Switch to `Lisk` mainchian
- Try token transfer cross chain
- Observe `Enevti` is shown even though its not active

### Additional information

- Fetch blockchain application list from `blockchain/apps` endpoint
- Fetch blockchain application meta data list from `blockchain/apps/meta` endpoint
- Filter non active applications
- By default `from` and `to` application should be selected based on the current `chainID`

#### Duplicate applications listed in drop down

- send token
Screenshot 2023-03-21 at 3 32 10 PM

- Application management 
Screenshot 2023-03-21 at 3 32 55 PM

3.0

eniolam1000752:
### What was the problem?

This PR resolves #4817

### How was it solved?

- [x] Allow only active application to be rendered in the to application dropdown on sendForm
- [x] Fix duplication on manage application list 
- [x] Fix bug with Network selector on mount
- [x] update unit test

### How was it tested?

- Visually
- Jest


setup/react/app/ApplicationBootstrap.js
```diff
@@ -22,7 +22,7 @@ const ApplicationBootstrap = ({ children }) => {
   useTransactionUpdate();
```
oskarleonard, line 22: Not related to this file:
Adding Enevti then trying to remove it crashes the application 
image

src/modules/blockchainApplication/manage/store/reducer.js
```diff
@@ -37,20 +37,26 @@ export const pins = (state = initialState.pins, { type, chainId }) => {
...
+          ...result,
+          [network]: { ...result[network], [application.chainID]: application },
+        }),
         { ...state }
       );
     }
 
     case actionTypes.deleteApplicationByChainId: {
-      delete state[chainId];
+      delete state[network][chainId];
```
oskarleonard, line 59: In all places we mutate the redux state we should replace the mutable logic with unmutable logic. For instance, this could be written as:
```
const { [chainId]: removedApplication, ...applications } = state[network];

const updatedState = {...state, [network]: applications};

return updatedState;
```

setup/react/app/ApplicationBootstrap.js
eniolam1000752: I guess I would raise this in another issue because trying to apply fixes here would be out of scope of this ticket @oskarleonard

oskarleonard: I added it to my network/application issue

oskarleonard (approved):

ikem-legend (approved):

PR #4885, opened 2023-03-28
Fix issues with table filter popover

### LINKED ISSUE #4736: Issues with table filter popover
### Expected behavior
- Popover filter button should have the right `Apply filter` button
- Filter popover should be layered properly on the blocks monitor page.
- Filter popover for the `Events` table  on the `wallet page` should contain the right filter parameters
-  Transactions filter popover should contain the correct filter paramters

### Actual behavior

- Transactions filter popover in the `wallets page` has incomplete filter parameters implemented
Date range

- The filter popover for the `Events` table within the `wallet page` doesn't contain the right filter parameters as described on [figma](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=209%3A95478&t=tKiTcNzqzAmJ5qNH-4)
image

- Filter popover isn’t layered properly on the blocks monitor page.
Transactions Status

### Steps to reproduce
- Visit the blocks monitor, wallet page and transaction monitor
- You would observe the above reported issues on the filter pop over 

all

reyraa:
### What was the problem?

This PR resolves #4736

### How was it solved?

- [x] Fix the filter and load new items overlap issue
- [x] Fix the horizontal position of the filter button
- [x] Fix form labels and inputs according to the design
- [x] Add missing inputs. Please note that the API does not support `module` and `name`, therefore I replaced them with the only other two supported parameters: `blockID`, `transactionID`.

### How was it tested?

- Updated unit test
- Tested manually


ManuGowda (approved):

ManuGowda:
@reyraa fyi.. add few commits to fix minor issues.

ManuGowda (changes requested):
Minor bug in wallet events
Screenshot 2023-03-30 at 10 15 18 AM

src/modules/transaction/components/TransactionEvents/TransactionEvents.js
```diff
@@ -32,13 +33,24 @@ const getFilterFields = (t) => [
...
 
+  const params = Object.keys(filters).reduce(
+    (acc, key) => {
+      if (filters[key] && key !== 'address') {
+        acc[key] = filters[key];
+      }
+      return acc;
+    },
+    { senderAddress: address }
+  );
```
eniolam1000752: I guess this is too much and kind of not readable I guess this is more readable

```
  const { address, ...filtersWithoutAddres } = filters;
  const params = { ...filtersWithoutAddres, senderAddress: address };
```

eniolam1000752 (dismissed):

ManuGowda (approved):

eniolam1000752 (approved):

PR #4891, opened 2023-04-03
Fixed blockchain application management issues

### LINKED ISSUE #4438: Issues observed in blockchain application management during the objective review
### Description

- Added blockchain applications gets removed after a refresh of the application ( they don’t get persisted).

- Delete blockchain application does not remove the application from the application list and has a weird behaviour when the application is pinned

- Render terminated applications at the bottom of the manage list.

- Back button on application details should be dark on light mode and light on dark mode (it should have the same color as the dismiss button)
image

- There exists an empty space on the block chain explore table when at most two applications are rendered as seen in the screenshot.
Pasted Graphic

- We need to review the tooltip message that's reveal when an attempt to delete a default app is done as seen below.
The default

### Motivation

These were suggestions made by the team during the objective review on the blockchain application feature.

### Additional Information

`suggestion` : We should have a way where when ever an application is terminated, it should show a termination application notification and this should get automatically deleted form the managed application list.

ikem-legend:
### What was the problem?

This PR resolves #4438

### How was it solved?

- [x] Fixed blockchain application delete
- [x] Render terminated applications at the bottom of the manage list
- [x] Fixed back button display

### How was it tested?

- Manually
- Unit tests


src/modules/blockchainApplication/manage/components/RemoveApplicationDetails/RemoveApplicationDetails.js
```diff
@@ -61,85 +94,131 @@ const RemoveApplicationDetails = ({ location, application, onCancel, nextStep })
...
+    deleteApplicationByChainId(chainId);
     nextStep({ application });
   };
 
+  if (isOnChainDataError || isOffChainDataError)
+    return (
+      <Dialog hasClose className={`${grid.row} ${grid['center-xs']}`}>
+        <div className={`${styles.wrapper} ${styles.errorWrapper}`}>
+          <Illustration name="applicationDetailsError" />
+          <div className={styles.errorText}>{t("Couldn't load application data")}</div>
```
reyraa: "Couldn't load application data" sounds too informal. Could you replace it with "Error loading application data"?

reyraa (changes requested):
Issues are fixed. Good job. I only have a comment regarding a copy. I think we should update it.

src/modules/blockchainApplication/manage/components/RemoveApplicationDetails/RemoveApplicationDetails.js
ikem-legend: @reyraa Fixed

eniolam1000752 (changes requested):
## Review comments

- Since I would be removing an application would there be a need to have the toggle pin application button here 
image

- shouldn't the logo of on the add application flow have a circular mask as we have it on other use cases?
image

- The container's overflow clips the popover at the top
image


- The skeleton loader on the app application modal seems to be off
image


- If Enevti is my current application and I go to the applications page I get an error as seen below
image





-

src/modules/blockchainApplication/manage/store/reducer.js
```diff
@@ -22,6 +22,7 @@ const initialState = {
 export const pins = (state = initialState.pins, { type, chainId }) => {
   switch (type) {
     case actionTypes.toggleApplicationPin:
+    case actionTypes.deleteApplicationByChainId:
```
oskarleonard: The delete logic cannot be placed here, needs a separate case. For example, deleting an application (that is not pinned) will push that application to the pins array.

src/modules/blockchainApplication/manage/components/RemoveApplicationSuccess/RemoveApplicationSuccess.js
```diff
@@ -14,7 +14,7 @@ const RemoveApplicationSuccess = ({ history, sharedData: { application } }) => {
         <Icon name="successCheckMark" />
         <p>{t('Application has now been removed')}</p>
         <p>
-          {t('You can always add')} <a href="#">{application.data.chainName}</a>{' '}
+          {t('You can always add')} <a href="#">{application.chainName}</a>{' '}
```
oskarleonard: This link makes no sense 🤔. Clicking Enevti text and be linked to my accounts :P. Would be much better to be linked to `http://0.0.0.0:8080/#/?modal=addApplicationList`

src/modules/blockchainApplication/manage/store/reducer.js
ikem-legend: @oskarleonard Nice catch. Fixed

src/modules/blockchainApplication/manage/components/RemoveApplicationSuccess/RemoveApplicationSuccess.js
ikem-legend: @oskarleonard Fixed

src/modules/blockchainApplication/manage/store/reducer.js
```diff
@@ -27,6 +27,12 @@ export const pins = (state = initialState.pins, { type, chainId }) => {
       }
       return chainId ? [...state, chainId] : [...state];
 
+    case actionTypes.deleteApplicationByChainId:
+      if (state.includes(chainId) && chainId) {
```
oskarleonard: I dont understand this extra check `&& chainId`

oskarleonard (dismissed):

src/modules/blockchainApplication/manage/store/reducer.js
ikem-legend: @oskarleonard I believe it was implemented to prevent empty string or other nullish values. Such values can exist if the button is clicked while the application is still loading. I'll remove it

eniolam1000752 (approved):

ManuGowda (approved):

Issue #4743, opened 2023-01-19
Filtering a validator's stakers leads to errors


ikem-legend:
### Expected behavior
Filtering a validator's stakers should return matching results or an empty list. Should be able to filter by name or address.

### Actual behavior
Filtering a validator's stakers throws a HTTP 400 error and also hides the filter bar
URL: validators/profile?address=lskaeec6426y8mkoq4oqgf5g4fsau738gb697pj8q
<img width="1327" alt="Screenshot 2023-01-18 at 03 43 33" src="https://user-images.githubusercontent.com/31376203/213344042-4bc28217-89f0-4c48-9cb5-e2f4ace74493.png">

### Steps to reproduce
- Select Validators from the side menu
- Select one validator with stakers
- Filter the list of stakers

### Which version(s) does this affect? (Environment, OS, etc...)
All

PR #4892, opened 2023-04-03
Resolve issue with block monitor and stakers

### LINKED ISSUE #4748: Issue with blocks monitor and details page
### Expected behavior
- On the block details page, empty block properties should be properly aligned
- On the block details page the event’s table header should have correct border bottom
- On the block monitor page, the status column is should be correctly placed when the new blocks button appears
- Sender should be in first column in transaction tables and blocks
### Actual behavior

- On the block details page the event’s table header seems to have a weird border bottom as though there are two overlaping lines
Pasted Graphic 21

- The styling on the transactions table row in the block details page is kind of off from [figma](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=38%3A36160&t=3yCFRSJGb1CxcDTB-4) also there is a blue like underline when a user hovers a row
Pasted Graphic 23

- Dont overlap table header with filter 
image

### Steps to reproduce
- Visit the blocks monitor page
- Clicking on a particular block should navigate you to the block details page
- issues recored above should be observed

all

eniolam1000752:
### What was the problem?

This PR resolves #4748 and #4743

### How was it solved?

- [x] Fix styling issues on validator details page and table
- [x] Fix styling issues on validator monitor page
- [x] Fix styling skeleton loader on staker section

### How was it tested?
- JEST
- Visually


src/modules/block/components/blockDetails/blockDetails.css
```diff
@@ -50,5 +50,21 @@ h1 {
...
+.feedback {
+  position: relative;
+}
+
+.blockDetailsWrapper {
+  & .errorFeedbackWrapper {
+    display: flex;
+    align-items: center;
+
+    & > div:first-child {
```
reyraa, line 66: The use of element name and pseudo selector can easily get affected by the data structure or component changes. Is this not possible using a class name?

src/modules/block/components/blocks/blocks.css
```diff
@@ -42,3 +42,29 @@
...
+
+.header {
+  width: 100%;
+  height: calc(var(--header-height) + 45px);
+
+  & > h1 {
+    margin-bottom: 12px;
+  }
+
+  & ~ div:nth-child(2) {
```
reyraa, line 58: This is also pretty tricky. Is this the only way to address your target element?

src/modules/block/components/blocks/blocks.css
```diff
@@ -42,3 +42,29 @@
...
+    background-color: var(--color-body-bg);
+    z-index: 4;
+    width: fit-content;
+    margin-top: -95px;
+    margin-left: calc(100% - 119px);
+    padding: 22px 0px;
+    padding-bottom: 20px;
+  }
+
+  & ~ div:nth-child(3) > header:first-child {
```
reyraa, line 70: Same here

src/modules/block/components/blocks/blocks.css
```diff
@@ -42,3 +42,29 @@
...
+
+  & > h1 {
+    margin-bottom: 12px;
+  }
+
+  & ~ div:nth-child(2) {
+    position: sticky;
+    top: 0;
+    background-color: var(--color-body-bg);
+    z-index: 4;
```
reyraa: Any way we could use variables for z-index and margin-top and paddings?

src/modules/transaction/components/BlockDetailsTransactions/BlockDetailsTransactions.css
```diff
@@ -1,5 +1,9 @@
 .content {
   padding: 0 !important;
+
+  & > header ~ a:first-of-type::before {
```
reyraa, line 4: Also here. This is a very specific selector and since it doesn't leave a trail on the HTML, it's hard for the next developer to apply changes without breaking your work.

src/modules/transaction/components/TransactionRow/schemas.css
```diff
@@ -1,11 +1,14 @@
...
+  margin-left: 0px;
+
   & .stakes {
     grid-area: stakes;
   }
 
   &::before {
     content: '';
-    min-width: 100%;
+    min-width: calc(100% - 30px);
```
reyraa, line 11: If you use one of the exiting variables instead of the hard coded `30px`, could can be sure this will remain aligned with the other components later on if we decide to change the box spacings.

src/modules/transaction/components/TransactionRow/schemas.css
```diff
@@ -25,6 +28,7 @@
       padding: 16px 8px;
       box-sizing: border-box;
       color: var(--color-content-light);
+      font-weight: var(--font-weight-normal);
     }
 
     & > *:nth-child(1) {
```
reyraa, line 34: Selector too weak and generic. Same goes for the next few rules.

src/theme/QueryTable/QueryTable.js
```diff
@@ -36,18 +36,18 @@ export const QueryTable = ({
     addUpdate();
   };
 
-  const subHeader = hasUpdate ? (
+  const subHeader = !!hasUpdate && (
```
reyraa: the `hasUpdate` name sounds boolean. What is the value? I'm just curious to know why do you need the double negative.

src/theme/QueryTable/QueryTable.js
```diff
@@ -36,18 +36,18 @@ export const QueryTable = ({
...
-  ) : null;
+  );
 
   useEffect(() => {
     if (isFetched && typeof onFetched === 'function') {
       onFetched(response);
     }
   }, [isFetched]);
-
+
```
reyraa: No trailing spaces.

reyraa (changes requested):
Looks good in general. Good job. Just a few concerns.

ikem-legend:
An observation about the blocks table header during scrolling
Screenshot 2023-04-04 at 09 09 43

src/modules/block/components/blockDetails/blockDetails.css
eniolam1000752: I had to do this because the element I was targeting is been abstracted by another component and the component exposed a className prop which did not target my element of interest so in-order to prevent my self from creating a new className prop just for the purpose of targeting this html element I had to do this

src/modules/block/components/blocks/blocks.css
eniolam1000752: I guess yess, because this element is been abstracted to a different component that is being consumed and there is not direct className prop exposed by this element that allows me to target it from the parent css.

src/modules/block/components/blocks/blocks.css
eniolam1000752: the same reason as above also applies here too

ManuGowda (commented):
Steps to reproduce:
- Navigate to transactions tab
- Click on one transaction details
- Click on blockID

Observe the blogs details showing invalid block and transaction and events information

Rendering invalid blocks page

Screenshot 2023-04-04 at 1 28 51 PM

src/modules/transaction/components/TransactionRow/schemas.css
eniolam1000752: I guess the reason this was done was because the table row can take in any element. to start defining styles based on each element would make the stying too specific making it hard for a third person to refactor if there is a change in component structure

src/theme/QueryTable/QueryTable.js
eniolam1000752: I did it to be double safe its a boolean at evaluating the `&&` operation. But guess it would always be a boolean type

ManuGowda (approved):
LGTM, nice job

reyraa (approved):

PR #4893, opened 2023-04-03
Merged network selector with application selector

### LINKED ISSUE #4873: Network Management Issues
### Description

While sanity testing alpha we found these issues regarding network switching

- Switch network in settings doesn’t work
- Changing network doesn’t reflect immediately
- Incorrect network passed while sending tokens https://alphanet-service.liskdev.net/api/v3/blockchain/apps/meta/tokens?limit=1&network=devnet&chainID=03000000&tokenID=0300000000000000
- Remove network switcher from settings
Screenshot 2023-03-31 at 11 56 25 AM

oskarleonard:
### What was the problem?

This PR resolves #4873

### How was it solved?

- Merging network and application selector in the same dropdown
- Removing network from settings modal

### How was it tested?

Manually


src/modules/blockchainApplication/manage/components/NetworkApplicationDropDownButton/NetworkApplicationDropDownButton.js
```diff
@@ -0,0 +1,47 @@
...
+import { getLogo } from '@token/fungible/utils/helpers';
+import { TertiaryButton } from 'src/theme/buttons';
+import Icon from 'src/theme/Icon';
+import { addSearchParamsToUrl } from 'src/utils/searchParams';
+import routes from 'src/routes/routes';
+import Network from '@common/components/bars/topBar/networkName';
+import { useCurrentApplication } from '../../hooks/useCurrentApplication';
+import styles from './NetworkApplicationDropDownButton.css';
+
+const ApplicationManagementList = ({ history, location }) => {
```
ManuGowda: The filename and export component name are different? `NetworkApplicationDropDownButton.js`

src/modules/network/components/DialogAddNetwork/DialogAddNetwork.js
```diff
@@ -0,0 +1,78 @@
...
+            />
+            <Input
+              size="l"
+              label="Service URL"
+              placeholder={t('e.g https://service.lisk.com or localhost:9901')}
+              feedback={errors.serviceUrl?.message}
+              status={errors.serviceUrl?.message ? 'error' : undefined}
+              {...register('serviceUrl', {
+                required: 'Service URL is required',
+              })}
```
ManuGowda: can you add missing validations for `name` and `URL`?

src/modules/network/components/networkSwitcherDropdown/NetworkSwitcherDropdown.js
```diff
@@ -121,15 +82,22 @@ function NetworkSwitcherDropdown({ noLabel, onNetworkSwitchSuccess }) {
...
         </MenuSelect>
       </div>
-      {blockchainAppsMeta.isError && !blockchainAppsMeta.isFetching && (
-        <div>
+      {networkStatus.isError && networkStatus.isFetched && (
+        <div className={styles.connectionFailedBlock}>
           <span>
             {t('Failed to connect to network!  ')}
-            <span onClick={blockchainAppsMeta.refetch}>{t('Try again')}</span>
+            <span onClick={networkStatus.refetch}>{t('Try again')}</span>
```
ManuGowda: maybe the `Try again` can be aligned to right side, currently its sandwiched

src/modules/blockchainApplication/manage/components/NetworkApplicationDropDownButton/NetworkApplicationDropDownButton.js
oskarleonard: Yes hate when they are different, fixed here https://github.com/LiskHQ/lisk-desktop/pull/4893/commits/49f528607e5bb5c216eda6f800dbe3e0a848b1cf

src/modules/network/components/DialogAddNetwork/DialogAddNetwork.js
oskarleonard: Yes, we can add whatever regex we want, what validation do you want? Was hesitant to add a standard one for URL, guess we want to support localhost etc?

ManuGowda: for both name and URL we have regex? https://github.com/LiskHQ/lisk-desktop/blob/9fb42a1a8f7ef621216a21fea7d1116ad6b87882/src/const/regex.js

oskarleonard: Aha, just added regex.url but do we really need regex for the name? Why shouldn't i as a user be able to call my network as i want 🤔

src/modules/network/components/networkSwitcherDropdown/NetworkSwitcherDropdown.js
oskarleonard: Design is next after another. But should probably have an underline if next to eachother:
1. 
image

2. 
image


One or two?

ManuGowda: 1

src/modules/network/components/DialogAddNetwork/DialogAddNetwork.js
ManuGowda: i think allowing only words and numbers is fine? any name would be not nice, so good to keep it under `a-z,0-9,_`?

ManuGowda (commented):
UI issues:

- Dropdown text not visible in dark mode
- Remove network switcher from dropdown

Screenshot 2023-04-06 at 2 58 31 PM
Screenshot 2023-04-06 at 3 00 36 PM

ManuGowda (approved):

ikem-legend (approved):

PR #4895, opened 2023-04-10
Retrieve off chain metadata for signing wallet connect transactions

### LINKED ISSUE #4824: Retrieve off chain metadata for signing wallet connect transactions
### Description

Blockchain applications should send the following set of information to request transaction signature from Lisk Desktop:

- `payload`: Encoded transaction hex string
- `schema`: Param schema for decoding
- `sendingChainID`

Lisk Desktop should be able to retrieve the following information from Lisk Service to correctly render transaction summary page:
- Sending blockchain application Logo in PNG and SVG
- Receiving blockchain application Logo in PNG and SVG
- Sending blockchain application name
- Token Symbol 

**Design**
- [Light mode](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?node-id=3590-252618&t=t1TUgNk0QMuaGC8K-4)

**Endpoints**
- [Metadata endpoint](http://165.227.246.146:9901/api/v3/blockchain/apps/meta?limit=20&network=devnet&chainID=04000001)
- [Tokens endpoint](http://165.227.246.146:9901/api/v3/blockchain/apps/meta/tokens?network=devnet&tokenID=0400000000000000)

reyraa:
### What was the problem?

This PR resolves #4824

### How was it solved?

- Use `useBlockchainApplicationMeta` to retrieve and display the app data
- Use existing hooks to retrieve and pass token data

### How was it tested?

- Test manually
- Update unit tests


src/modules/blockchainApplication/connection/components/RequestSummary/index.js
```diff
@@ -41,32 +45,37 @@ const RequestSummary = ({ nextStep }) => {
...
+  const sendingChainID = request?.chainId.replace('lisk:', '');
+  const tokenData = useAppsMetaTokens({ config: { params: { chainID: sendingChainID } } });
+
   const approveHandler = () => {
     const moduleCommand = joinModuleAndCommand(transaction);
     const transactionJSON = toTransactionJSON(transaction, moduleCommandSchemas[moduleCommand]);
-    // @todo remove the hardcoded normalized
-    // values according to #4824
+    const { recipientChainID } = request?.request?.params ?? {};
+    const sendingChain = metaData.data.data.find((item) => item.chainID === sendingChainID);
```
eniolam1000752: I would suggest that this is optional chained since during the state transition of `metaData` it could be undefined at first.

```suggestion
    const sendingChain = metaData?.data?.data?.find(({ chainID }) => chainID === sendingChainID);
```

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
```diff
@@ -41,32 +45,37 @@ const RequestSummary = ({ nextStep }) => {
...
+    sendingChain.chainID = sendingChainID;
+    const recipientChain = metaData.data.data.find((item) => item.chainID === recipientChainID);
```
eniolam1000752: like wise this too

```suggestion
    const recipientChain = metaData?.data?.data?.find(({ chainID }) => chainID === recipientChainID);
```

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
```diff
@@ -41,32 +45,37 @@ const RequestSummary = ({ nextStep }) => {
...
+    recipientChain.chainID = recipientChainID;
+    const token = tokenData.data.data.length > 0 ? tokenData.data.data[0] : defaultToken;
```
eniolam1000752, line 63:
```suggestion
    const token = tokenData?.data?.data?.length > 0 ? tokenData.data.data[0] : defaultToken;
```

I would also suggest we don't do this because what I see happening here is that due to state transition of `tokenData` it would be undefined first and this would result to the UI possibly showing a wrong token symbol `LSK`. 

I would advise that until the tokenMeta and chainMeta is obtained/resolved, no symbol is shown and all actions are disabled.

eniolam1000752 (changes requested):

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
reyraa: Fair point. I'll apply this.

ManuGowda (changes requested):
Screenshot 2023-04-25 at 10 09 11 AM
Screenshot 2023-04-25 at 10 09 57 AM
Screenshot 2023-04-25 at 10 10 27 AM

eniolam1000752 (changes requested):
### Ui comments

- Padding doesn't look right on small and medium screens
image

- Active account checkmark seems to be too close to the checkbox of the other account at its side
image

- Application breaks on trying to sign request transaction signing
image

-

libs/wcm/hooks/useSession.js
```diff
@@ -1,26 +1,46 @@
...
-      await setSession({
-        ...session,
-        request: false,
-        data: proposalEvents.meta,
-      });
+      await setSessionProposal(proposalEvents.meta);
       const status = await onApprove(proposalEvents.meta, selectedAccounts);
       removeEvent(proposalEvents);
-      refreshPairings();
+      setSessionProposal(null);
```
oskarleonard, line 40: This part looks a bit strange? Calling the same function twice in the same scope, and also one time with `await` and the other call doesn't use `await`

reyraa: Resolved

ManuGowda (approved):

eniolam1000752 (approved):

PR #4911, opened 2023-04-14
Fix account management issues

### LINKED ISSUE #4899: Account management issues on alpha.1
### Expected behavior

- Hide actionable items when current account is not set
   - Dont just hide button with currentAddress && button  we should 

- Fix logic of why isValidator is true when there is no account.

- Ui/UX on both light and dark theme should be properly aligned.

### Actual behavior
- As seen on this screen shoot. No account is selected and therefore these buttons doesn't make sense to show
image

- Paper wallet section is wrongly placed on the create account flow as seen in the screen shot below
image

- Text is not visible in dark mode
image

- Toggle password visibility button doesn’t change the cursor to a pointer type
image

- Creation time in generated encrypted account JSON seems off by 1 hour (I.e my time was 01:26:13 when I generated this account on my local machine) but it read as 00:26:13.
image

- on clicking Add account on the the popover menu, it does not get dismissed even though I have been navigated to the account account options page
image

- Enter password step does not have a back button and its ui doesn’t look like other enter password modals
image

- On the backup you secrete recovery phrase stage, paper wallet is wrongly placed
image

- There is no way to access manage account modal. I was only able to access it after removing an account.
image

- When I don’t have any account selected and I proceed to navigate to the wallet’s page, I get a wrong ui of Manage accounts
image

### Steps to reproduce
- Ensure that no account is selected
- navigate to the validators page
- A user is able to interact with actions that requires an account to be set 

- All versions

eniolam1000752:
### What was the problem?

This PR resolves #4899

### How was it solved?

- [x] Fix dark theme issues
- [x] Hide action buttons on validators monitor page when account is not selected
- [x] Refactor enter password UI
- [x] Refactor UI alignment on Recover passphrase page
- [x] Fix manage account page
- [x] Fix dismissal of account menu when item is clicked 
- [x] Fix unit test

### How was it tested?
- Visually
- Jest

### DIFF
--- a/src/modules/account/components/AccountManagementDropdown/AccountManagementDropdown.js
+++ b/src/modules/account/components/AccountManagementDropdown/AccountManagementDropdown.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import DropdownButton from 'src/theme/DropdownButton';
 import { TertiaryButton } from 'src/theme/buttons';
 import Icon from 'src/theme/Icon';
@@ -9,12 +9,22 @@ import styles from './AccountManagementDropdown.css';
 
 const AccountManagementDropdown = ({ currentAccount, onMenuClick }) => {
   const { name, address, isHW } = currentAccount.metadata;
+  const [isDropdownShown, setIsDropdownShown] = useState(false);
+
+  const handleDismissMenu = () => {
+    setIsDropdownShown(false);
+  };
+  const handleOpenMenu = () => {
+    setIsDropdownShown(true);
+  };
 
   return (
     <DropdownButton
       className={styles.dropDownMenu}
       wrapperClassName={styles.wrapper}
       ButtonComponent={TertiaryButton}
+      isDropdownShown={isDropdownShown}
+      onDropdownOpen={handleOpenMenu}
       buttonClassName={`account-management-dropdown ${styles.dropdownButton}`}
       buttonLabel={
         <div className={styles.accountWrapper}>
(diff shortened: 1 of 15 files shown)


src/modules/account/components/AccountManagementDropdown/AccountManagementDropdown.js
```diff
@@ -9,12 +9,22 @@ import styles from './AccountManagementDropdown.css';
...
 const AccountManagementDropdown = ({ currentAccount, onMenuClick }) => {
   const { name, address, isHW } = currentAccount.metadata;
+  const [isDropdownShown, setIsDropdownShown] = useState(false);
+
+  const handleDismissMenu = () => {
+    setIsDropdownShown(false);
+  };
+  const handleOpenMenu = () => {
+    setIsDropdownShown(true);
+  };
```
ikem-legend: I believe this can be simplified to:
```
const handleDismissMenu = () => {
    setIsDropdownShown(!isDropdownShown);
};

File: src/modules/auth/components/SignupSuccessed/accountCreated.css [hunk 2]
```diff
@@ -34,3 +34,9 @@
   color: var(--color-slate-gray);
   margin-bottom: 30px;
 }
+
+.subheader {
```
ikem-legend: This and the corresponding JS file should be renamed to `subHeader`

ManuGowda (commented):
LGTM, please address the comments and run `npm run format`

ManuGowda (approved):

ikem-legend (approved):

PR #4915, opened 2023-04-17
UI fixes transactions blocks applications

### LINKED ISSUE #4902: UI fixes for domains transaction, blocks, applications
### Expected behavior
1. View transaction stats
[LINK](http://0.0.0.0:8080/#/transactions), Text "Amount per transaction" should be "Fee per transaction" according to design

2. View transaction
Transaction detail page should have height as a link

3. List blockchain applications
[LINK](http://0.0.0.0:8080/#/applications), 
  - No pointer cursor on application in list. 
  - Search input doesn't look like design. 
  - Table shouldn't have any title according to design
  - Show search option when applications are greater than 5

4. Add blockchain application
[LINK](http://0.0.0.0:8080/#/wallet?modal=addApplicationList), 1. Missing total LSK column. 2. Can add an application which is already added

5. List blocks
[LINK](http://0.0.0.0:8080/#/blocks), Date for old blocks are not in correct format, missing time

6. View block
[LINK](http://0.0.0.0:8080/#/block?id=f3fb6cc87334a5033c88fc39d4693e99dbd7b0aef1ed7697c67b8feb8b1516e5), Status should be text and not icon check

oskarleonard:
### What was the problem?

This PR resolves #4902

### How was it solved?
Applying easy fixes to look like design

### How was it tested?

Manually


src/modules/transaction/components/Overview/Overview.js
```diff
@@ -232,7 +232,7 @@ const Overview = () => {
           </div>
         </div>
         <div className={`${styles.column} ${styles.pie}`}>
-          <h2 className={styles.title}>{t('Amount per transaction (LSK)')}</h2>
+          <h2 className={styles.title}>{t('Fee per transaction (LSK)')}</h2>
```
ManuGowda: I think the designs were updated to fee, but the endpoints are not adapted to fee yet? http://devnet-service.liskdev.net:9901/api/v3/transactions/statistics?limit=7&interval=day

so for now we can keep it `Amount per transaction`?

if the API is updated in future we can change in UI.

ManuGowda (commented):
Small change

ManuGowda (approved):

src/modules/transaction/components/Overview/Overview.js
oskarleonard: fixed

ikem-legend (approved):

PR #4917, opened 2023-04-18
Fix application management issues

### LINKED ISSUE #4907: Application management issues
### Expected behavior
- Chain details should include address and copy icon if address exists
- Remove application button for a default application in application dropdown should be properly aligned

### Actual behavior
- Chain details only shows the copy icon when address does not exist
Deposited-20230412-000724

- Remove application button in application dropdown is slightly off for a default application
Select application-20230412-001305

### Steps to reproduce
Application details
- Visit the Applications page
- Select an application from the list to view its details

Application dropdown
- Click the application dropdown in the top bar
- Select a non default application
- Hover on the default application and observe the delete button position

All

ikem-legend:
### What was the problem?

This PR resolves #4907

### How was it solved?
- [x] Updated address and deposit logic on application detail views
- [x] Updated styling of remove application button

### How was it tested?

- Visually


src/modules/blockchainApplication/explore/components/BlockchainAppDetailsHeader/index.js
```diff
@@ -44,21 +44,23 @@ const BlockchainAppDetailsHeader = ({ application, chainAction, loading }) => {
...
-                    className="tx-id"
-                    containerProps={{
-                      size: 'xs',
-                      className: 'copy-address',
-                    }}
-                  />
-                </span>
-              </ValueAndLabel>
-            </Box>
+            {!!address && (
```
oskarleonard: Any reason why we are not just doing `address && (...` here and in RemoveApplicationDetails.js

oskarleonard (dismissed):

src/modules/blockchainApplication/explore/components/BlockchainAppDetailsHeader/index.js
ikem-legend: @oskarleonard Slight oversight while implementing that but I just fixed it. Good catch

eniolam1000752 (approved):

oskarleonard (approved):

Issue #4963, opened 2023-04-28
Amount in rewards modal is not properly formatted


ikem-legend:
### Expected behavior
Amount in rewards modal should be comma formatted

### Actual behavior
Amount in rewards modal is not properly formatted
![lisk-app-test-fdbk-10](https://user-images.githubusercontent.com/31376203/235016141-57aba70a-b3b3-46e6-81cd-64698d9f5419.png)

### Steps to reproduce
- Log in with validator account
- Visit the Validators page
- Select the Stakes tab on the top right
- On the stakes page, click on the "Claim rewards" button

### Which version(s) does this affect? (Environment, OS, etc...)
All

PR #4964, opened 2023-04-28
Fix recovery phrase text word count and rewards amount formatting

### LINKED ISSUE #4959: Recovery phrase text should have the right count of words in the recovery phrase
### Expected behavior
Creating new account returns 24 word passphrase but text should suggest writing down the correct number of words

### Actual behavior
Creating new account returns 24 word passphrase but text suggests writing down 12 words
lisk-app-test-fdbk-6

### Steps to reproduce
- Visit the Add new account page
- Create new account
- Navigate to secret recovery phrase page

All

ikem-legend:
### What was the problem?

This PR resolves #4959 and #4963

### How was it solved?

- Updating description text
- Implementing amount formatting

### How was it tested?

- Visually

### DIFF
--- a/src/modules/wallet/components/passphraseRenderer/index.js
+++ b/src/modules/wallet/components/passphraseRenderer/index.js
@@ -145,7 +145,7 @@ class PassphraseRenderer extends React.Component {
             <h2 className={styles.header}>{t('Secret recovery phrase')}</h2>
             {subheader && (
               <p className={styles.subheader}>
-                {t('Please write down these 12 words carefully, and store them in a safe place.')}
+                {t('Please write down these 24 words carefully, and store them in a safe place.')}
               </p>
             )}
           </>
--- a/src/modules/wallet/components/passphraseRenderer/index.js
+++ b/src/modules/wallet/components/passphraseRenderer/index.js
@@ -146,7 +146,7 @@ class PassphraseRenderer extends React.Component {
             {subheader && (
               <p className={styles.subheader}>
                 {t(
-                  'Please write down these 12/24 words carefully, and store them in a safe place.'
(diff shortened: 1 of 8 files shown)


src/modules/wallet/components/passphraseRenderer/index.js
```diff
@@ -145,7 +145,7 @@ class PassphraseRenderer extends React.Component {
             <h2 className={styles.header}>{t('Secret recovery phrase')}</h2>
             {subheader && (
               <p className={styles.subheader}>
-                {t('Please write down these 12 words carefully, and store them in a safe place.')}
+                {t('Please write down these 24 words carefully, and store them in a safe place.')}
```
ManuGowda: This can be problem, user can have both 12 and 24 words? you need dynamically render this text based on phrase length

ManuGowda: 12/24

src/modules/wallet/components/passphraseRenderer/index.js
```diff
@@ -145,7 +145,9 @@ class PassphraseRenderer extends React.Component {
             <h2 className={styles.header}>{t('Secret recovery phrase')}</h2>
             {subheader && (
               <p className={styles.subheader}>
-                {t('Please write down these 12 words carefully, and store them in a safe place.')}
+                {t(
+                  'Please write down these 12/24 words carefully, and store them in a safe place.'
+                )}
```
ManuGowda:
```suggestion
                  'Please write down these seed values carefully. Ensure that you keep this in a safe place, with access to the seed you can re-create the account.
                )}
```

I think this more generic in my opinion.

ManuGowda (commented):
rest LGTM

ManuGowda (approved):

eniolam1000752 (approved):

PR #4976, opened 2023-05-04
Implement stakes queue clearing on application switch

### LINKED ISSUE #4968: Clear staking queue when switching application
### Expected behavior
- When switching application we need to clear the staking queue, similarly how we do when switching network. This is because you cannot continue that flow after having switched application.
- Show info modal if staking queue have items before switching application
  - info text: Switching application will clear current staking queue. 
- Opening the staking queue on Enevti should not crash application

### Actual behavior
Staking queue is not cleared.

### Steps to reproduce
1. Add stakes to staking queue
2. Switch application
3. Try to continue the staking flow 

Mac

ikem-legend:
### What was the problem?

This PR resolves #4968

### How was it solved?

- Updating stakes queue clearing logic
- Implementing confirmation dialog

### How was it tested?

- Manually
- Unit tests


src/modules/pos/validator/components/StakeForm/StakeForm.js
```diff
@@ -146,7 +146,7 @@ const StakeForm = ({ t, stakes, account, isStakingTxPending, nextStep, history,
 
   const feedback = validateStakes(
     stakes,
-    BigInt(posToken?.availableBalance),
+    BigInt(posToken?.availableBalance ?? 0),
```
ManuGowda, line 149: This is incorrect? the balance should be based on current account

ikem-legend: @ManuGowda The balance is still based on the current account. `posToken` has the balance of the current account in the POS token. It's passed in from the parent component

ManuGowda: hmm ok, the naming is confusing for sure, thanks for clarification

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
```diff
@@ -1,18 +1,48 @@
...
+    // clear stakes list during application switch
+    if (pendingStakes.length) {
+      const state = {
+        header: 'Pending Stakes',
+        content:
+          'Switching your application and (or) network will remove all your pending stakes. Are you sure you want to continue?',
+        cancelText: 'Cancel switch',
+        cancelFn: removeSearchParamsFromUrl(history, ['modal']),
+        confirmText: 'Continue to switch',
+        confirmFn: /* istanbul ignore next */ () => {
```
oskarleonard: Interesting wording, dont we usually do `onConfirm` or `handleConfirm`, etc.. `onCancel`?

ikem-legend: @oskarleonard Fixed

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
```diff
@@ -1,18 +1,48 @@
...
+import { stakesReset } from 'src/redux/actions';
+import {
+  removeThenAppendSearchParamsToUrl,
+  removeSearchParamsFromUrl,
+} from 'src/utils/searchParams';
 import { selectCurrentApplication } from '../store/selectors';
 import { setCurrentApplication } from '../store/action';
 
-export function useCurrentApplication() {
+export function useCurrentApplication(history) {
```
oskarleonard, line 14: Why did we change from using window? Or our own util `import history from 'src/utils/history';` for this?

Dont we now we need to pass in history from all places that uses `setApplication`, for example in `SelectNode`, `useApplicationManagement`, etc..

src/modules/common/components/ConfirmationDialog/ConfirmationDialog.css
```diff
@@ -0,0 +1,47 @@
...
+  max-width: 430px;
+
+  & .container {
+    display: flex;
+    justify-content: center;
+    margin: 0;
+  }
+
+  & header {
+    padding: 20px 20px 0 !important;
```
oskarleonard: You can increase css specificity by moving this block in under container
```
& .container {
  & header {
```

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
ikem-legend: @oskarleonard Using the passed in history lets us set the state for the confirmation dialog via the location state rather than via the URL. When we upgrade the react-router packages then we can use the useHistory hook instead

oskarleonard: But than we need to pass in the history in many other places, my second paragraph?

If that is a big refactor we could do

```
import history from 'src/utils/history';

export function useCurrentApplication() {
...
removeThenAppendSearchParamsToUrl(history, { modal: 'confirmationDialog' }, ['modal'], state);
...
```
And in ConfirmationDialog.js
```
import history from 'src/utils/history';

function ConfirmationDialog() {
const { header, content, cancelText, onCancel, confirmText, onConfirm } = history.location.state ?? {};
...
```

src/modules/common/components/ConfirmationDialog/ConfirmationDialog.css
ikem-legend: Fixed

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
ikem-legend: It's not a big refactor to pass in history because we only actively try to switch applications via the application/network dialog. While bootstrapping the application for instance, staking queue will be empty and hence no need to worry about it. `history.location` is mutable so it's never a good idea to use it. Then the history import you specified is used in only two places excluding tests. Therefore I chose not to use it because I'm unaware of whatever challenges have been had with it. The best route is still the react-router package update and I've added a TODO for that.

oskarleonard (dismissed):

src/modules/blockchainApplication/manage/hooks/useCurrentApplication.js
```diff
@@ -1,18 +1,49 @@
...
+  );
 
   const setApplication = useCallback((application, applicationNode) => {
-    dispatch(setCurrentApplication(application));
-    /* istanbul ignore next */
-    client.create(applicationNode || application.serviceURLs[0]);
+    // clear stakes list during application switch
+    if (pendingStakes.length) {
+      const state = {
+        header: 'Pending Stakes',
```
ManuGowda:
```suggestion
        header: 'Pending stakes',
```

Please remember to name the title and variable with proper casing in the future, i have noticed this common problem at many places 🙏🏼

ManuGowda (changes requested):
the queue should be cleared while
- Switch network
- Switch application
- Switch account

currently its only handled for app switch and not network switch and account switch

please check the edge cases

oskarleonard (approved):

ManuGowda (approved):

PR #4982, opened 2023-05-08
Removed usbPort label since it only works on macbook

### LINKED ISSUE #4960: Hardware wallet USB ID doesn't show up in modal on Windows
### Expected behavior
Hardware wallet USB ID should show up in Hardware wallet details modal
Screenshot 2023-04-27 at 12 55 37

### Actual behavior
Hardware wallet USB ID doesn't show up in modal on Windows
lisk-app-test-fdbk-7

### Steps to reproduce
- Connect hardware wallet
- Observe hardware wallet details modal from top bar

Windows 10

oskarleonard:
### What was the problem?

This PR resolves #4960

### How was it solved?

- Added a ping device link so the user can actually see which is the active device
- Removed USB ID label

### How was it tested?

Manually


src/modules/hardwareWallet/components/SelectHardwareDeviceModal/components/HwDeviceItem.js
```diff
@@ -16,17 +21,18 @@ function HwDeviceItem({ hwDevice }) {
...
+  }
 
   return (
     <div className={styles.hwDevice}>
       <Icon name="iconLedgerDevice" className={styles.hwWalletIcon} />
       <div className={styles.infoContainer}>
         <h5 className={styles.modelInfo}>{`${manufacturer} ${product}`}</h5>
-        <span className={styles.path}>{usbPort}</span>
+        <TertiaryButton className={classNames(styles.btnShowPubkey)} onClick={handleShowPubkeyOnDevice}>
+          {t('Show pubkey on device')}
```
ManuGowda: I think public key option will be harder for users to view and verify, we can use the old approach.

See here 2.3 option.

Screenshot 2023-05-08 at 1 44 56 PM

ManuGowda (changes requested):
Can we use the verify address method similar to 2.3 ?

ManuGowda (approved):

ikem-legend (approved):

PR #4988, opened 2023-05-10
Fix scroll issue with confirmation button on small screens

### LINKED ISSUE #3893: Go To Confirmation button cropped off on small screens
### Expected behavior
When creating a transaction, while using a small screen, I should be able to scroll down to the bottom of the modal to use the Go To Confirmation button.

### Actual behavior
Due to the small screen the button is cropped off and there is no scroll bar, making it impossible to send the transaction.

### Steps to reproduce
- Open Lisk Desktop
- Sign in
- Press send LSK
- Decrease the height of the window to hide the button
- Try to navigate to the bottom :)

Lisk Desktop (version 2.0.2)
MacBook Pro (13-inch, M1, 2020)
MacOs Big Sur (version 11.6)

ikem-legend:
### What was the problem?

This PR resolves #3893

### How was it solved?

- Adding scroll to the styling for the pages

### How was it tested?

- Visually


ManuGowda (changes requested):
Please observe the same issues on other transactions as well

1. Register validator
2. Unlock
3. Vote
4. Claim Rewards

In general this problem can exists on all types of transaction views

src/modules/pos/validator/components/RegisterValidatorView/registerValidator.css
```diff
@@ -1,9 +1,9 @@
 @import '../../../../../../setup/react/app/mixins.css';
 
 .wrapper {
-  max-width: 720px;
   background-color: var(--color-white);
   border-radius: var(--border-radius-box);
+  overflow: scroll;
```
eniolam1000752: Instead of `scroll` lets use `auto` so that on windows we don't have a scroll bar taking up space when there isn't any overflow

src/modules/pos/validator/components/UnlockBalanceView/styles.css
```diff
@@ -0,0 +1,3 @@
+.wrapper {
+  overflow: scroll;
```
eniolam1000752: likewise here, we can use `auto`

src/modules/token/fungible/components/SendForm/form.css
```diff
@@ -9,6 +9,7 @@
   flex-direction: column;
   justify-content: flex-start;
   align-items: flex-start;
+  overflow-y: scroll;
```
eniolam1000752: likewise here, we can use `auto`

src/modules/wallet/components/RegisterMultisigForm/styles.css
```diff
@@ -5,6 +5,7 @@
   width: 100%;
   padding: 20px;
   padding-bottom: 0;
+  overflow: scroll;
```
eniolam1000752: likewise here, we can use `auto`

src/modules/wallet/components/RegisterMultisigView/styles.css
```diff
@@ -6,6 +6,7 @@
   background: transparent;
   width: 100%;
   height: auto;
+  overflow: scroll;
```
eniolam1000752: likewise here, we can use `auto`

eniolam1000752 (changes requested):

ManuGowda (approved):

oskarleonard (approved):

PR #4993, opened 2023-05-11
Create new account with latest protocol spec

### LINKED ISSUE #4992: Create new account with latest protocol spec
### Description

With Lisk Core 4.0 we use custom derivation path for account creation i.e `m/44'/134'/0'` (default path). So that users can manage multiple account with one secret recovery phrase.

### Motivation

- Create new accounts with custom derivation path
- Enable access to legacy Lisk accounts

ManuGowda:
### What was the problem?

This PR resolves #4992

### How was it solved?

- Create new account with latest protocol spec
- Provide access to legacy account

### How was it tested?

- Create new account
- Access legacy account


eniolam1000752 (dismissed):

src/modules/account/components/AddAccountForm/AddAccountForm.js
```diff
@@ -109,27 +108,17 @@ const AddAccountFormContainer = ({
...
-  const [derivationPath, setDerivationPath] = useState(defaultDerivationPath);
-  const derivationPathErrorMessage = useMemo(
-    () => getDerivationPathErrorMessage(derivationPath),
-    [derivationPath]
-  );
-
-  const onDerivationPathChange = (value) => {
-    setDerivationPath(value);
-  };
+  const [derivationPathEvent, setDerivationPathEvent] = useState({
```
oskarleonard: Why did we do this refactor? With this change we are now using two different variables for identical data. This means that we need to keep track of both to never make sure that they differ, which usually results in buggs and a pain to maintain.

image
image

src/modules/auth/components/CustomDerivationPath/index.js
```diff
@@ -1,12 +1,23 @@
...
+  const derivationPathErrorMessage = useMemo(
+    () => getDerivationPathErrorMessage(derivationPath),
+    [derivationPath]
+  );
+
+  const onPathInputChange = ({ target }) => {
+    const value = target.value;
+
+    setDerivationPath(value);
+    onChange?.({ derivationPathErrorMessage, derivationPath: value });
```
oskarleonard: To fix above we need to do
`onChange?.({ derivationPathErrorMessage:  getDerivationPathErrorMessage(derivationPath), derivationPath: value });`

But i really dont recommend it because "duplicate vars holding same data" bugs are real ninjas in hiding. 
We cannot encapsulate derivation path logic here since parent need it.

oskarleonard: If we had a `FieldSet` component this component wouldn't even be needed. The only reason it is called CustomDerivationPath right now is because of the label text. All logic is handled in parents

oskarleonard (dismissed):

eniolam1000752 (approved):

oskarleonard (approved):

Issue #5006, opened 2023-05-17
Peer information fully not displayed


ManuGowda:
### Expected behavior

Should display port, version

### Actual behavior

<img width="1350" alt="Screenshot 2023-05-17 at 9 40 41 PM" src="https://github.com/LiskHQ/lisk-desktop/assets/6449871/babdb568-7ba6-44a6-bada-1618a0792147">

### Which version(s) does this affect? (Environment, OS, etc...)
3.0.0

PR #5011, opened 2023-05-19
Fix status and peer list rending issues

### LINKED ISSUE #5010: Incorrect status shown for transactions and blocks
### Expected behavior

- Remove confirmations from transaction status (we have finality status now)
- Show valid block and transaction status returned by API

### Actual behavior

- Why are we showing the number of confirmations for the transaction status? Doesn't make any sense because we have the finality now.
- The block status shows final whereas the API returns isFinal to be false.

image

image

image

3.0

ManuGowda:
### What was the problem?

This PR resolves #5010 and #5006

### How was it solved?

- Use correct property name exposed by API

### How was it tested?

Visually and Update unit tests


src/theme/Icon/index.js
```diff
@@ -26,6 +26,7 @@ import bookmarksIconEmptyStateDark from '@setup/react/assets/images/icons/bookma
 import checkboxFilled from '@setup/react/assets/images/icons/checkmark-filled.svg';
 import checkboxCircleFilled from '@setup/react/assets/images/icons/checkmark-circle-filled.svg';
 import checkmark from '@setup/react/assets/images/icons/checkmark.svg';
+import inprogress from '@setup/react/assets/images/icons/inprogress.svg';
```
ikem-legend: @ManuGowda Please use camel case for icon names

ikem-legend (commented):
Looks good. Just a minor change

ikem-legend (commented):
Looks good. Just a minor change

eniolam1000752 (changes requested):
Sort by latest block height for a network node seems to be bugged.

When I sort in ascending order, at the bottom, I see that `Unknown` comes before a block height which is lesser that the height just above the `Unknown` value

image

oskarleonard (dismissed):

eniolam1000752 (dismissed):

ikem-legend (dismissed):

ikem-legend (approved):

eniolam1000752 (approved):

Issue #5021, opened 2023-05-24
Clarify add and create account with new and legacy approach


ManuGowda:
### Description

All new accounts after Lisk v4 is recommended to create using new key derivation approach. However the current UI/UX can be confusing to user, we need to improve the accessibility and also convey the user about account addition or creation explicitly.
 
### Acceptance Criteria

- Provide clear option to add/create an account using old or new approach
- Implement the proposed [design](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?type=design&node-id=4087-257413&t=OJjQ5f3PspOG3TJC-4), [design2](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?type=design&node-id=39-47558&t=OJjQ5f3PspOG3TJC-4)
 
#### Feedback from community

I have noticed a potentially major UX issue with the new desktop wallet. Since you have changed the default derivation path, every user who wants to add their old account will encounter a window with a changed derivation path.
Once they proceed, they will be logged into a different account with a balance of 0 LSK. It took me 10 minutes to realize that I needed to access the settings and enable the Enable Access to Legacy Lisk Account and  re-adding the account.
Additionally, once the option is checked in the settings, every address created by the wallet is generated with the old derivation path, leading to further confusion (which I consider a bug).
I can guarantee that there will be a significant number of people adding old addresses with the new derivation path and panicking when they see a balance of 0 LSK.
My suggestion is to move the "legacy account option" from the settings and place it directly below the "custom derivation path" when adding an address, ensuring that users must click it every time they want to add a legacy address (maybe even state that the option is for accounts created before DD.MM.YY)

PR #5033, opened 2023-05-26
Clarify add and create account with new and legacy approach

### LINKED ISSUE #5020: Unable to send maximum possible balance from an account
### Expected behavior

Should be able to send the maximum possible balance from an account

### Actual behavior

Screenshot 2023-05-23 at 4 21 38 PM

### Steps to reproduce

have balance more than 100,000 LSK and try to send 

3.0.0

### LINKED ISSUE #5024: Differences in account creation spec. desktop vs. mobile
### Expected behavior
Same spec.

### Actual behavior

**Recovery Phrase**

- Desktop: 24 words
- Mobile: 12 words

**Account name**

- Desktop: Multibyte cannot be entered.
- Mobile: Multibyte can be entered.

**Enable access to legacy Lisk accounts Switcher**

- Desktop: OFF (default)
- Mobile: ON (default)

If both are used, it would be better to use the same spec. because the differences in spec. are confusing.
Maybe this is a result of UI and UX considerations....

Desktop: v3.0.0-beta.0
Mobile: v3.0.0-beta.0

ManuGowda:
### What was the problem?

This PR resolves #5024 and #5021 

### How was it solved?

- Provide option to switch between legacy and new approach to add existing accounts
- Enabled users to create an account with 12 and 24 words recovery phrase

### How was it tested?

- Create an account using 12 and 24 words
- Add legacy and new account to wallet


ManuGowda:
@ikem-legend please

- Remove the advance settings for legacy account access
- The UI for 12/24 words selection needs to match proposal

eniolam1000752 (approved):

ikem-legend (approved):

PR #5084, opened 2023-06-13
Fix horizontal scrollbar bug in modals

### LINKED ISSUE #5032: UI bug horizontal scrollbar in modals
### Expected behavior
Modals should not have horizontal scroll.
I would suggest that we remove duplicate `overflow: auto` in children. Only one element should have `overflow-y: auto` for the main dialog scroll. 

### Actual behavior
image

### Steps to reproduce
This is probably triggered on different displays with different resolution, i use a 32 inch 4k display. To reproduce test by using different zoom settings in the browser

ikem-legend:
### What was the problem?

This PR resolves #5032

### How was it solved?

- Updating the styling for the different transaction modals

### How was it tested?

Visually


eniolam1000752 (changes requested):
Should the Dialog component its self benefit from not having a horizontal scroll. 

This is so that other uses cased of the component don't suffer from an horizontal scroll.

ikem-legend:
> Should the Dialog component its self benefit from not having a horizontal scroll.
> 
> This is so that other uses cased of the component don't suffer from an horizontal scroll.

I considered doing so but was uncertain about if that might possibly create problems in some other views

eniolam1000752 (dismissed):

src/modules/pos/validator/components/RegisterValidatorView/index.js
```diff
@@ -15,7 +15,7 @@ const RegisterValidator = ({ history, t }) => {
   }, []);
```
oskarleonard, line 15: NRTTL found some more
image
image

I think we should do as we discussed in slack some time ago, this problem got introduced in https://github.com/LiskHQ/lisk-desktop/pull/4988 if my memory serves me correctly, which is caused by children inside children both having scroll overflow

eniolam1000752 (approved):

src/modules/pos/validator/components/RegisterValidatorView/index.js
ikem-legend: Fixed

ManuGowda (approved):
Tested on mac works fine (couldn't test unlock and claim rewards)

PR #5085, opened 2023-06-13
Show error modal if dryrun fails

### LINKED ISSUE #5037: No error state when dryrun endpoint fails
### Expected behavior
If /dryrun req fails with `500`, `503`... we should show an error state.  

### Actual behavior
No error state is shown. 

Currently we only show error state when dryrun request returns `200` with `result: 0`
image

oskarleonard:
### What was the problem?

This PR resolves #5037

### How was it solved?

Check if dryrun endpoint returns an error and in that case show the error dialog when sending an transaction

### How to test?
1. Force dryrun to fail by adding an additional param
image
2. Open a dialog to perform a transaction, for example send token
3. Try to send token
4. It should fail and show a retry button in the last step, Click the retry button.

### DIFF
--- a/src/modules/transaction/store/actions.js
+++ b/src/modules/transaction/store/actions.js
@@ -116,38 +116,19 @@ export const transactionBroadcasted =
       // https://github.com/LiskHQ/lisk-desktop/issues/4698 should handle this logic
     }
 
-    if (error) {
-      dispatch({
-        type: actionTypes.broadcastedTransactionError,
-        data: {
-          error: error.message,
-          transaction,
-        },
-      });
-    } else {
-      if (dryRunResult.data?.result === -1) {
-        dispatch({
-          type: actionTypes.broadcastedTransactionError,
-          data: {
-            error: dryRunResult.data?.errorMessage,
-            transaction,
-          },
-        });
-      }
+    const transactionErrorMessage =
+      error?.message ||
+      (dryRunResult?.data?.result === 0
+        ? dryRunResult?.data?.events.map((e) => e.name).join(', ')
+        : dryRunResult?.data?.errorMessage);
 
-      if (dryRunResult.data?.result === 0) {
-        // @TODO: Prepare error message by parsing the events based on each transaction type
-        // https://github.com/LiskHQ/lisk-desktop/issues/4698 should resolve all the dry run related logic along with feedback
-        const temporaryError = dryRunResult.data?.events.map((e) => e.name).join(', ');
-        dispatch({
-          type: actionTypes.broadcastedTransactionError,
-          data: {
-            error: temporaryError,
-            transaction,
-          },
-        });
-      }
-    }
+    dispatch({
+      type: actionTypes.broadcastedTransactionError,
+      data: {
+        error: transactionErrorMessage,
+        transaction,
+      },
+    });
 
     return false;
   };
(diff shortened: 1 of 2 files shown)


src/modules/transaction/store/actions.js
```diff
@@ -115,27 +116,37 @@ export const transactionBroadcasted =
...
-      });
+      if (dryRunResult.data?.result === 0) {
+        // @TODO: Prepare error message by parsing the events based on each transaction type
+        // https://github.com/LiskHQ/lisk-desktop/issues/4698 should resolve all the dry run related logic along with feedback
+        const temporaryError = dryRunResult.data?.events.map((e) => e.name).join(', ');
+        dispatch({
+          type: actionTypes.broadcastedTransactionError,
+          data: {
+            error: temporaryError,
+            transaction,
```
ManuGowda: In all 3 cases, error/0/-1 only the `error` is different? so why not remove the duplication and just prepare the error and dispatch?

e.g 

```
const transactionError = error || dryRunResult.data?.result === 0 ? dryRunResult.data?.events.map((e) => e.name).join(', ') : dryRunResult.data?.errorMessage;

dispatch({
          type: actionTypes.broadcastedTransactionError,
          data: {
            error: transactionError,
            transaction,
          },
        });
```

oskarleonard: The snippet you provided will always dispatch something while the old code sometimes did not dispatch anything.  Are we sure that we want to make this change? According to previous code the function can in some cases never dispatch anything but return just false. With the snippet in your comment we would always dispatch something, maybe that is what we want?


To make it behave like before we would need something like this 
image

ManuGowda: Ideally if the API or dry run is not successful we should always dispatch an error to abort the operation, so I think the changes you made should be fine?

ManuGowda (dismissed):

src/modules/transaction/store/actions.js
```diff
@@ -96,9 +96,10 @@ export const transactionBroadcasted =
     const moduleCommand = joinModuleAndCommand(transaction);
     const paramsSchema = moduleCommandSchemas[moduleCommand];
     let broadcastResult;
-    const dryRunResult = await dryRun({ transaction, serviceUrl, paramsSchema });
 
-    if (dryRunResult.data?.result === 1) {
+    const [error, dryRunResult] = await to(dryRun({ transaction, serviceUrl, paramsSchema }));
+
+    if (dryRunResult?.data?.result === 1) {
```
eniolam1000752: I think it would be best if we have an enum of dryrun results instead of using magic values here

e.g

```
const dryRunResultTypes = {
  ERROR_1: 1,  // ERROR_1 is just a mock string actual error meaning should be used instead
  ERROR_0: 0, // ERROR_0 is just a mock string actual error meaning should be used instead
}
```

Then consumption can be used as 

```suggestion
    if (dryRunResult?.data?.result === dryRunResultTypes.ERROR_1) {
```

src/modules/transaction/store/actions.js
```diff
@@ -115,28 +116,19 @@ export const transactionBroadcasted =
...
-        type: actionTypes.broadcastedTransactionError,
-        data: {
-          error: dryRunResult.data?.errorMessage,
-          transaction,
-        },
-      });
-    }
+    const transactionErrorMessage =
+      error?.message ||
+      (dryRunResult?.data?.result === 0
```
eniolam1000752: meaning the enum thiny can now be consumed here to as 
```suggestion
      (dryRunResult?.data?.result === dryRunResultTypes.ERROR_0
```

eniolam1000752: I believe this would make it easy for any one reading the code to know what 1 or 0 means at first glance without trying to figure out the Boolean logic.

eniolam1000752 (changes requested):

ManuGowda (approved):

eniolam1000752 (approved):

Issue #5075, opened 2023-06-08
Action buttons "edit" and "delete" lead to the same action page


punkrock:
On the Stakes page: No matter if you click on the "edit" or "trash/delete" button it opens the edit button. I think you should only use the edit button and remove the trash icon or create different pop ups for each action. Why am I suggesting this? Because I clicked on the trash button and then on "confirm" because I thought the vote will be removed. Instead I just got the info that nothing is in the queue.

![image](https://github.com/LiskHQ/lisk-desktop/assets/1260700/e7d18245-c8d4-44e8-af7e-6ebf4e01b5fb)

ManuGowda:
Thanks for the suggestion @punkrock. I think we will simplify this to following

- Edit -> users can increase or decrease the stake and actions will be confirm and cancel
- Delete -> when user clicks on delete then the stake will be removed and show this view

![Screenshot 2023-06-12 at 9 05 14 AM](https://github.com/LiskHQ/lisk-desktop/assets/6449871/39e110df-b73e-48b3-8355-43f63e8c3118)

What do you think?

ManuGowda:
Since the current edit flow ensures both edit and remove stake feature, we decided to remove the trash icon to avoid the duplication.

Issue #5069, opened 2023-06-02
Privacy Policy on the settings page can't be opened


endro42:
### Expected behavior

A user should be able to open the privacy policy.

### Actual behavior

The Anonymous analytics collection toggle is switched instead of navigating to the privacy policy page

![Jun-02-2023 14-03-45](https://github.com/LiskHQ/lisk-desktop/assets/12698415/b948c2db-7b81-4eef-b13f-bd0044e2343a)

### Steps to reproduce

- Open settings page
- Try to open the privacy policy

Issue #5041, opened 2023-05-26
Switched naming for active/standby validators


przemerr:
### Expected behavior
Active validators should be standby validators, and vice versa
### Actual behavior
![image](https://github.com/LiskHQ/lisk-desktop/assets/25085095/c2e722e0-9b22-4ad2-a9b7-5e94fdb47786)

### Steps to reproduce
Go to validators tab in the Lisk Desktop wallet
### Which version(s) does this affect? (Environment, OS, etc...)
Lisk Desktop 3.0-beta

PR #5091, opened 2023-06-15
Validator and staking domain bugfixes

### LINKED ISSUE #5067: Show validator name in staking queue + staking summary
### Description

Show the validator name in the staking queue and also in the staking summary

### Motivation
The staking queue + staking summary is a last moment to double check your changes made to staking. When only showing the validators address it makes the extra check harder for a user. 

### Additional Information
image
image

ManuGowda:
### What was the problem?

This PR resolves #5067, #5041, #5069 and #5075

### How was it solved?

- Remove `trash/delete` icon
- Fix validator name rending by changing property name
- Switched active and standby validator legend and fixed values
- Fix privacy policy link

### How was it tested?

- Stake should now have edit and as part of edit user can remove the stake or edit it
- While staking you should be able to view validator name in queue and summary
- Check active and standby validator legend and see the counts to be displayed based on API
- You should be able to navigate to privacy policy


oskarleonard (dismissed):

eniolam1000752 (changes requested):
## UX issues

- I was able to click on the confirm button even though I did not change the already existing stake on a validator meaning i was able to see the continue to staking queue button and clicking that took me to an obviously empty staking queue which I think is wrong.

image

My suggestion would be that if the stake amount hasn't been changed, then the confirm button should be disabled.

oskarleonard (approved):

eniolam1000752 (approved):

PR #5092, opened 2023-06-16
Display when commission change is allowed again

### LINKED ISSUE #5048: Display When Commission Change is Allowed Again
### Description

Show an approximate date and time countdown that indicates when a commission of a validator can be adjusted again. Similar as the unlock stake/ votes countdown. 

### Motivation

A commission of a validator can only be adjusted after a certain time, ~30 days on Lisk, and a max percentage. To make it more clear in the wallet when commissions can be changed again, after the last change, a time indicator counting down to that moment would be very helpful.

The returned error message after trying to change the commission, shows that the transaction failed. An additional reason for the fail would give the user a better indication why.

image

### Additional Information

Using windows desktop wallet V3.0.0.-beta.0

oskarleonard:
### What was the problem?

This PR resolves #5048

### How was it solved?

Check that the transaction pool for given user doesnt have a changeCommission transaction before enabling the edit commission button

#### Case 1 - Edit button disabled after commission change
1. Go to your validator profile
2. Change the commission **below** the current one
3. Expected: change commission button should be disabled with a tooltip explaining why it is disabled
4. Expected:  Wait ~5 minutes and the new commission should be fetched and the button should be enabled again.
#### Case 2 - Error message when increasing commission second time within short time
1. Go to your validator profile
2. Try to change the commission **higher** than the current one
3. Expected:  You should now see an error message that tells you that you have to wait a month until you can change it.


src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.js
```diff
@@ -73,6 +80,9 @@ export const ChangeCommissionForm = ({ prevState, nextStep }) => {
       inputFeedback = t('Input decimal places limited to 2');
     } else if (!(newCommissionParam >= 0 && newCommissionParam <= 10000)) {
       inputFeedback = t('Commission range is invalid');
+    } else if (lastCommissionChange && isCommissionIncrease(value, currentCommission)) {
+      const timeCanUpdate = moment().to(lastCommissionChange, true);
+      inputFeedback = t(`You can only increase commission in ${timeCanUpdate}`);
     } else if (!isNewCommissionValid) {
```
ManuGowda, line 86: I think with above condition, the below if would never execute?

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
```diff
@@ -57,6 +57,30 @@ describe('ChangeCommissionForm', () => {
...
+    it('should not pass when commission increases more than 5%', async () => {
+      const { getByTestId, getByText } = renderWithRouterAndQueryClient(
+        ChangeCommissionForm,
+        props
+      );
+      const value = '96';
+      const button = getByText(buttonText);
+      const input = getByTestId('newCommission');
+      fireEvent.change(input, { target: { value } });
+      expect(button.closest('button').disabled).toBeTruthy();
```
ManuGowda, line 82: You can also assert the error `You cannot increase commission more than 5%`?

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
```diff
@@ -57,6 +57,30 @@ describe('ChangeCommissionForm', () => {
...
+    it('should not pass when new commission have more than 2 decimal places', async () => {
+      const { getByTestId, getByText } = renderWithRouterAndQueryClient(
+        ChangeCommissionForm,
+        props
+      );
+      const value = '89.0101';
+      const button = getByText(buttonText);
+      const input = getByTestId('newCommission');
+      fireEvent.change(input, { target: { value } });
+      expect(button.closest('button').disabled).toBeTruthy();
```
ManuGowda, line 96: same here assert for the error in input field?

src/modules/pos/validator/components/ChangeCommission/Form/hooks/useLastCommissionChange.js
```diff
@@ -0,0 +1,31 @@
...
+export const useLastCommissionChange = () => {
+  const { data: posConstants } = usePosConstants();
+  const [
+    {
+      metadata: { address },
+    },
+  ] = useCurrentAccount();
+  const { data: transactions } = useTransactions({
+    config: {
+      params: { address, moduleCommand: MODULE_COMMANDS_NAME_MAP.changeCommission, limit: 1 },
```
ManuGowda: is it safer to use sort parameter? so that we alway pick the latest transaction to calculate timestamp properly?

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.js
oskarleonard: It would execute if `!lastCommissionChange` like if the user hasn't changes the commission before. 

But i see that i also need to check if commissionDate isAfter. 
So if commission increase && the user needs to wait, that feedback should take president over the greater than 5% error feedback, which is why i placed it before in the if statement. Please check this https://github.com/LiskHQ/lisk-desktop/pull/5092/commits/95ec5e2b8884f40c7e4f5a384a9b4e4eb681c6ad

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
oskarleonard: Added https://github.com/LiskHQ/lisk-desktop/pull/5092/commits/dcead4abea1ed14f22c48c1697b1bd0fa32c6fe6

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
oskarleonard: Added https://github.com/LiskHQ/lisk-desktop/pull/5092/commits/dcead4abea1ed14f22c48c1697b1bd0fa32c6fe6

src/modules/pos/validator/components/ChangeCommission/Form/hooks/useLastCommissionChange.js
oskarleonard: Yes, we should definitely take the last, i though the last always was first?
What is the params i should add to enforce this?
 Can i enforce it like this? `params: { address, moduleCommand: MODULE_COMMANDS_NAME_MAP.changeCommission, limit: 1,  sort=timestamp:desc},`

ManuGowda: yes that seems correct, https://petstore.swagger.io/?url=https://betanet-service.lisk.com/api/v3/spec#/Transactions/get_transactions

ManuGowda (dismissed):

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
```diff
@@ -57,6 +57,32 @@ describe('ChangeCommissionForm', () => {
...
+        props
+      );
+      const value = '89.0101';
+      const button = getByText(buttonText);
+      const input = getByTestId('newCommission');
+      fireEvent.change(input, { target: { value } });
+      expect(screen.getByText('Input decimal places limited to 2')).toBeTruthy();
+      expect(button.closest('button').disabled).toBeTruthy();
+    });
+
```
eniolam1000752, line 98: I think it would also be nice to assert agains negative commission value inputed @oskarleonard

src/modules/pos/validator/components/ChangeCommission/Form/hooks/useCommissionChangeDate.js
```diff
@@ -0,0 +1,36 @@
...
+  const networkStatus = useNetworkStatus();
+
+  return (
+    lastChangeCommissionTimestamp &&
+    new Date(
+      (lastChangeCommissionTimestamp +
+        posConstants.data?.commissionIncreasePeriod *
+          networkStatus.data?.data?.genesis?.blockTime) *
+        1000
+    )
```
eniolam1000752: to make this easier to read, I think its best to store this in a variable(with a good naming) possibly on line 26 then consume here

src/modules/pos/validator/components/ChangeCommission/Form/hooks/useCommissionChangeDate.test.js
```diff
@@ -0,0 +1,41 @@
...
+      ...mockTransactions.data[0],
+      moduleCommand: MODULE_COMMANDS_NAME_MAP.changeCommission,
+      timeStamp: 123455,
+    };
+
+    useTransactions.mockReturnValue({
+      data: {
+        data: [transaction],
+      },
+    });
```
eniolam1000752: instead of mocking useTransactions this way, you can utilize MSW `Server.use()` function to create a separate server instance with this mock data. 

This way, the natural state transition of the useCommissionChangeDate hook would be maintained.

you can get inspiration from `src/modules/token/fungible/hooks/queries/useGetHasUserAccount.test.js`

src/modules/pos/validator/components/ValidatorProfile/DetailsView.js
```diff
@@ -76,13 +88,43 @@ const DetailsView = ({ data, isMyProfile }) => {
...
+                  ))}
+              </div>
+              <div
+                className={classNames(
+                  styles.value,
+                  styles.capitalized,
+                  hasChangeCommission &&
+                    onEdit &&
+                    typeof onEdit === 'function' &&
+                    styles.textLineThrough
```
eniolam1000752: I think it would be more readable if you use classNames like this. What do you think

```suggestion
                    styles.value,
                    styles.capitalized,
                    { 
                      [styles.textLineThrough] : hasChangeCommission && onEdit && typeof onEdit === 'function'
                    }
```

eniolam1000752 (changes requested):

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
oskarleonard: Added, although that part of the logic was not added in this branch, https://github.com/LiskHQ/lisk-desktop/pull/5092/commits/531c5139a303eb28bfdbd5e2c40b8f5c5493755a

src/modules/pos/validator/components/ChangeCommission/Form/hooks/useCommissionChangeDate.js
oskarleonard: I rather only do the calculation and time creation when it is needed (if `lastChangeCommissionTimestamp` is true). Refactored that code to use a function though, in this https://github.com/LiskHQ/lisk-desktop/pull/5092/commits/3f1c2892ddaa049520e1f0a0f653a5a73c764d81

src/modules/pos/validator/components/ValidatorProfile/DetailsView.js
oskarleonard: 👍 I agree when there are conditional classnames that is how we should do, https://github.com/LiskHQ/lisk-desktop/pull/5092/commits/200b2551d1c4c8f074ca87fc49ba01a3f5ab0e61

src/modules/pos/validator/components/ChangeCommission/Form/hooks/useCommissionChangeDate.test.js
oskarleonard: Ok, lets extend with isLoading
https://github.com/LiskHQ/lisk-desktop/pull/5092/commits/0d5dfab2cafbfb653919f592583fd2ac1ad3394a

eniolam1000752 (approved):

ManuGowda (approved):

Issue #5026, opened 2023-05-25
Tab index on "Set up your account password" is incorrect 


endro42:
### Expected behavior

The tab index when entering a password should automatically move from "Password" to "Password Confirmation"

### Actual behavior

The tab index automatically moves to the "show password confirmation" icon instead of the input field. Other option is too include the "show password" icon as well.

![May-24-2023 22-32-41](https://github.com/LiskHQ/lisk-desktop/assets/12698415/6baf317a-bc65-4a2e-a6a5-37d89189f635)

### Steps to reproduce

- Create an account and try every click via the tab & enter keys

PR #5097, opened 2023-06-20
Handle page breakdown and resolve password index positioning bug

### LINKED ISSUE #5027: Failed to load "wallet tokens" results in a error page that breaks the entire user flow
### Expected behavior

When a request errors or ends up in a timeout, a user might want to see an error for the section that executed the request but still be able to navigate to different pages without entirely switching context (showing error page)

### Actual behavior

Tokens on the wallet page are loading and end up throwing the error page

https://github.com/LiskHQ/lisk-desktop/assets/12698415/8ec15c95-c86f-472a-b658-ce023ed39f71

### Steps to reproduce

- Navigate to the wallet page on Lisk Desktop (Beta)

eniolam1000752:
### What was the problem?

This PR resolves #5027, #5026

### How was it solved?

- [x] Ignore tabIndex on the input type toggle button
- [x] Fix Issue on wallet page crashing

#### How to test solution of #5027
- Navigate to the add account by secret passphrase.
- Go through the flow until the point of entering a password and account name
- You should be able to navigate the input field smoothly by clicking on the tab key.

#### How to test solution of #5026
- Change the url in the useTokenBalance hook to a wrong one (this would allow you mock a failed token balance api call)
- Navigate to the wallet page.
- You should now be seeing an error section with a retry button instead of the application crashing.


ManuGowda (commented):
@eniolam1000752 the changes works fine but it doesn't handle other case similar to token balances

e.g update `/api/${API_VERSION}/events` endpoint to `/api/${API_VERSION}/eventsss` then you dont get retry option, rather no events, which is wrong?

Screenshot 2023-06-20 at 7 31 36 PM
Screenshot 2023-06-20 at 7 32 14 PM

oskarleonard (approved):

ManuGowda (approved):

eniolam1000752:
> @eniolam1000752 the changes works fine but it doesn't handle other case similar to token balances
> 
> e.g update `/api/${API_VERSION}/events` endpoint to `/api/${API_VERSION}/eventsss` then you dont get retry option, rather no events, which is wrong?
> 
> Screenshot 2023-06-20 at 7 31 36 PM Screenshot 2023-06-20 at 7 32 14 PM

would be resolved by this issue https://github.com/LiskHQ/lisk-desktop/issues/5100#issue-1767053311

Issue #5040, opened 2023-05-26
Update nodejs version to use LTS (18.16.0)


ManuGowda:
### Description

Update nodejs version to use LTS (18.16.0)

### Motivation

- v16 is currently in active maintenance and end of life is at 2023/09
- Production release should have longer active and maintenance period, hence update to Release: https://nodejs.org/en/blog/release/v18.16.0/

### Additional Information

https://nodejs.dev/en/about/releases/

- Update nvm to use the current LTS node.js v18.16.0
- Update readme
- Update package.json

PR #5104, opened 2023-06-21
Update node.js to LTS and yarn

### LINKED ISSUE #5103: Replace npm with yarn as package manager
### Description

- Install dependencies in parallel
- Improved Performance and Speed
- More secure

ManuGowda:
### What was the problem?

This PR resolves #5103, #5040 

### How was it solved?

- Update lisk-client to beta.2
- Update node.js to v18

### How was it tested?

- `rm package-lock.json` && `rm -rf node_modules`
- `nvm install`
- `npm i -g yarn` or `npx yarn`
- `yarn run test`


Jenkinsfile
```diff
@@ -27,7 +28,7 @@ pipeline {
 					nvm(getNodejsVersion()) {
 						sh '''
 						rm -rf lisk-service/ # linting will fail otherwise
-						npm run lint
+						yarn run lint
```
ikem-legend: yarn doesn't need to have the `run` keyword before it can run any script so `yarn lint` works fine. Other instances can be updated accordingly

ManuGowda:
This PR needs to remain open until

- Able to notaries the app (Build prod app and test the signed app)

ManuGowda:
> This PR needs to remain open until
> 
> * Able to notaries the app (Build prod app and test the signed app)

Works now.

package.json
```diff
@@ -280,7 +278,8 @@
     "webpack-bundle-analyzer": "4.7.0",
     "webpack-cli": "^4.2.0",
     "webpack-dev-server": "3.11.3",
-    "webpack-merge": "=4.1.5"
+    "webpack-merge": "=4.1.5",
+    "yarn": "1.22.19"
```
oskarleonard, line 282: Any special reason for why need to add yarn here? I have never added it when used yarn in the past, but maybe i missed something 🤔

.gitignore
```diff
@@ -17,6 +17,8 @@ reports
 .history
```
oskarleonard, line 17: NRTTL:

Getting error, i tested like this:
1. `rm -rf node_modules yarn.lock`
2. `nvm install`
3. `npm i -g yarn`
4. `yarn install --registry=https://npm.lisk.com`

image

package.json
ManuGowda: if we do not have yarn installed gloablly and if we have npx then we can do `npx yarn` and use the locally installed yarn

.gitignore
ManuGowda: you can just do yarn without all that.. it appears the error started around `liskhq` installation, which seems to be wrong due to the registry you mentioned, try with `just yarn`

package.json
oskarleonard: Aha, I didnt know yarn could work like that 👍

.gitignore
oskarleonard: works now!

libs/wcm/utils/jsonRPCFormat.js
```diff
@@ -0,0 +1,72 @@
...
+const SERVER_ERROR_CODE_RANGE = [-32000, -32099];
+
+const STANDARD_ERROR_MAP = {
+  [PARSE_ERROR]: { code: -32700, message: 'Parse error' },
+  [INVALID_REQUEST]: { code: -32600, message: 'Invalid Request' },
+  [METHOD_NOT_FOUND]: { code: -32601, message: 'Method not found' },
+  [INVALID_PARAMS]: { code: -32602, message: 'Invalid params' },
+  [INTERNAL_ERROR]: { code: -32603, message: 'Internal error' },
+  [SERVER_ERROR]: { code: -32000, message: 'Server error' },
+};
```
eniolam1000752: I see a duplication in `libs/wcm/utils/error.js`

eniolam1000752 (changes requested):

ikem-legend (dismissed):

oskarleonard (dismissed):

eniolam1000752 (dismissed):

ikem-legend (approved):

eniolam1000752 (approved):

PR #5106, opened 2023-06-21
Implement error state on all tables

### LINKED ISSUE #5100: All tables invoking an api call should have an error state and a retry option when api call fails.
### Description

Presently All table showing a list makes an api call to render this list. Though when this api call fails, a wrong status message of empty state is shown to users instead of an error message. 

We need to implement the following

- All tables invoking an api call should render an error status message when there is an api failure
- A retry button along side the message should be shown possibly as seen in the image below
image

eniolam1000752:
### What was the problem?

This PR resolves #5100

### How was it solved?

- [x] Fix table to display error on api failure
- [x] Fix failing unit test
- [x] Add retry functionality error states on tables 

#### How to test #5100
- Change the api URL on any of the api query hook you would want to test in code.
- Navigate to the table invoking the above altered hook.
- You should see an error state with a retry button.

### DIFF
--- a/src/modules/transaction/components/TransactionEvents/TransactionEvents.js
+++ b/src/modules/transaction/components/TransactionEvents/TransactionEvents.js
@@ -50,6 +50,7 @@ const TransactionEvents = ({ address, transactionID = '', blockID = '', isWallet
     error,
     hasNextPage,
     fetchNextPage,
+    refetch,
   } = useTransactionEvents({
     config: { params },
   });
@@ -97,6 +98,7 @@ const TransactionEvents = ({ address, transactionID = '', blockID = '', isWallet
             isWallet,
           }}
           error={error}
+          retry={refetch}
           emptyState={{
             message: t('There are no events for this account.'),
             illustration: 'emptyEventsIllustration',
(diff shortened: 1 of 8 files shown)


ManuGowda (commented):
Screenshot 2023-06-23 at 12 20 35 PM

when `/api/${API_VERSION}/token/balancess` causes the send or request flow or in general other transaction modal to be not usable we should let user know about the problem?

Can you take care of this edge cases as well when the transaction forms has error we should let the the know about the issue?

eniolam1000752:
> Screenshot 2023-06-23 at 12 20 35 PM
> 
> when `/api/${API_VERSION}/token/balancess` causes the send or request flow or in general other transaction modal to be not usable we should let user know about the problem?
> 
> Can you take care of this edge cases as well when the transaction forms has error we should let the the know about the issue?

okay, though I think this should be handled in this [PR](https://github.com/LiskHQ/lisk-desktop/pulls#issue-1768383341) since already in that PR I would be handling error cases with dry run. @ManuGowda

ManuGowda (approved):

oskarleonard (approved):

Issue #4575, opened 2022-11-10
Upgrade webpack-dev-server


oskarleonard:
### Description
To fix these vulnerabilities:
- braces  <2.3.1 (low)
- glob-parent  <5.1.2 (high)
- node-forge  <=1.2.1 (high)

We need to upgrade to webpack-dev-server version 4

### Acceptance Criteria
Update webpack-dev-server to version 4

### Additional Information
We have at least 2 options that we need to migrate, `contentBase` and `inline`.
Here you can find the [Migration guide](https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md)

PR #5120, opened 2023-06-27
Update electron dependency and sandboxing

### LINKED ISSUE #5109: Electron renderer lacks sandboxing
### Description

The lisk-desktop application creates its main BrowserWindow instance where the application will render without the sandbox option. This option only became default in version 20 of electron; however, lisk-desktop currently uses version 17.

```
win.browser = new BrowserWindow({
    width: width > 1680 ? 1680 : width,
    height: height > 1050 ? 1050 : height,
    minHeight: 576,
    minWidth: 769,
    center: true,
    webPreferences: {
        // Avoid app throttling when Electron is in background
        backgroundThrottling: false,
        // Specifies a script that will be loaded before other scripts run in the page.
        preload: path.resolve(__dirname, '../src/ipc.js'),
        //
        https://www.electronjs.org/docs/latest/tutorial/security#isolation-for-untrusted-con
            tent
nodeIntegration: false,
        contextIsolation: true,
    },
});
```

### Recommendation

- Resolve https://github.com/LiskHQ/lisk-desktop/issues/4663

- Alternatively we can use `sandbox: true` while for the `webPreferences`, [docs](https://www.electronjs.org/docs/latest/tutorial/sandbox)

ManuGowda:
### What was the problem?

This PR resolves #5109, #4575

### How was it solved?

- Updated outdated electron and other dependencies
- With the latest version of `electron` the sandboxing is handled by default, so it was resolved with an update

### How was it tested?

- `rm -rf node_modules && rm -rf app/node_modules`
- `yarn --cwd app && yarn`
- `yarn dev`
- `yarn test`


Jenkinsfile
```diff
@@ -28,7 +28,7 @@ pipeline {
 					nvm(getNodejsVersion()) {
 						sh '''
 						rm -rf lisk-service/ # linting will fail otherwise
-						yarn lint
+						yarn run lint
```
ikem-legend, line 31: Any reason why these scripts were updated to use `yarn run` rather than just `yarn`?

ManuGowda: sometimes the `yarn` wont run the commands correctly, so its safer to keep it `yarn run` for consistent results.

ikem-legend:
The dev script needs to be updated in order to work on Windows. So it should be changed from 
```"dev": "cross-env DEBUG=true NACL_FAST=disable && NODE_OPTIONS='--openssl-legacy-provider' webpack serve --config ./setup/config/webpack.config.dev",```
to
```"dev": "cross-env DEBUG=true NACL_FAST=disable NODE_OPTIONS='--openssl-legacy-provider' webpack serve --config ./setup/config/webpack.config.dev",```
This is because adding `&&` before an environment variable breaks on Windows if such environment variable isn't already global to the machine. Using `cross-env` solves that.

File: package.json [hunk 2]
```diff
@@ -277,7 +278,7 @@
     "webpack": "5.24.2",
     "webpack-bundle-analyzer": "4.7.0",
     "webpack-cli": "^4.2.0",
-    "webpack-dev-server": "3.11.3",
+    "webpack-dev-server": "4.15.1",
```
oskarleonard, line 281: Probably related to webpack upgrade. But now the whole screen is covered in error (before it was only the chrome console). For example when editing a stake i get this 
image

oskarleonard (dismissed):

package.json
ManuGowda: I think its better that way? before we use to ignore these errors, its better that we resolve issues than getting away from it 😅

ikem-legend (approved):

package.json
oskarleonard: Yea, i agree, really dont like the console errors we are having. Hopefully this will make us prioritise them more

oskarleonard (approved):

eniolam1000752 (approved):

PR #5121, opened 2023-06-27
Indicate unclaimed rewards

### LINKED ISSUE #5064: No indication of crediting unclaimed rewards
### Description

Inform users about crediting unclaimed rewards when adjusting or removing stakes.

### Motivation

In the current protocol implementation, if a user has some staking rewards to claim and adjusts the corresponding delegate stake, the rewards are automatically claimed. This is stated in the LIP as follows: `In order to support efficient calculation of rewards using the sharing coefficient, all unclaimed rewards of this stake are credited to the staker before modifying the staked amount.`
https://github.com/LiskHQ/lips/blob/main/proposals/lip-0070.md#efficient-calculation-of-rewards

However, in the current implementation of the Lisk Desktop wallet, the user is not informed about this process when changing the stake. After changing the stake, the unclaimed rewards simply disappear, and there is no trace of the rewards being credited in the transaction list. Consequently, if the user doesn't pay attention to their balance, they may mistakenly assume that the rewards were not credited.

### Additional Information

You can test the flow by staking LSK to a sharing delegate and waiting to receive staking rewards, then readjust the stake.

eniolam1000752:
### What was the problem?

This PR resolves #5064

### How was it solved?

- [x] Implement required UI
- [x] integrated dry run on stake form
- [x] Extend unit tests

### How was it tested?

- Add, Edit or remove a stake from/to a validator
- Navigate to the staking queue modal
- Click on the `Contine` button
- Staked validators with rewards should be visible


src/modules/pos/validator/components/StakeForm/StakeForm.js
```diff
@@ -154,10 +169,44 @@ const StakeForm = ({ t, stakes, account, isStakingTxPending, nextStep, history,
...
+
+        if (
+          name === 'rewardsAssigned' &&
+          module === 'pos' &&
+          data.stakerAddress === currentAccount.metadata.address
+        ) {
+          return {
+            ...result,
+            [data.validatorAddress]: data,
+            total: BigInt(result.total || 0) + BigInt(data.amount),
```
ManuGowda: whats this total? we only need `data.amount` 🤔

ManuGowda: As you can see, the data structure tells us that we should only care about amount https://github.com/LiskHQ/lisk-sdk/blob/98c3d9c588b56845e46eb8a209985a92bf478c67/framework/src/modules/pos/events/rewards_assigned.ts#L18, maybe result property is related to success of failure i guess

src/modules/pos/validator/components/StakeSummary/StakeSummary.test.js
```diff
@@ -176,6 +182,22 @@ describe('StakingQueue.Summary', () => {
     expect(wrapper).toContainMatchingElements(12, '.stake-item-address');
   });
 
+  it('should render rewards', () => {
+    const wrapper = mountWithRouter(Summary, { ...props, edited, removed, added });
+    const addedItemList = wrapper.find('.stake-item').at(0);
+    const editedItemList = wrapper.find('.stake-item').at(4);
+    const removedItemList = wrapper.find('.stake-item').at(9);
+    expect(addedItemList).toHaveText(
+      'lskdwsyfmcko6mcd357446yatromr9vzgu7eb8y110 LSKReward:  0.001 LSK'
```
ManuGowda: This is wrong ? it should always display validator name? i.e `subzero 100 LSK Reward: 0.001 LSK`?

ManuGowda (changes requested):

ManuGowda:
Screenshot 2023-06-28 at 9 19 49 AM

@eniolam1000752 can't we display each element in 1 row? otherwise it can be difficult to view the changes.

src/modules/transaction/components/StakeValidate/StakeValidator.js
```diff
@@ -17,43 +19,62 @@ const ItemList = ({ items, heading }) => {
...
-        <InfoColumn title={t('Transaction Fees')} className="stake-fees">
-          {transaction?.composedFees.Transaction}
+        <InfoColumn
+          title={t('Total rewards')}
+          toolTipText={t(
+            'Total rewards credited to your account in effective to changes in stakes.'
+          )}
+        >
+          <span className={styles.totalReward}>
+            <TokenAmount val={rewards.total.toString()} token={token} />
```
oskarleonard: Crashes here for me
image

src/modules/pos/validator/components/StakeForm/StakeForm.js
eniolam1000752: the reason for total is because we are trying to normalise the response from dry run and total would aggregate all rewards as we normalise the response. This means we don't need to start aggregating total on the summary page with something like this. `Object.values(rewards).reduce((result, { amount}) => result + amount, 0) ` instead we simply just do `rewards.total` (more of like an ahead of time computation rather than a just in time computation of total rewards)

src/modules/pos/validator/components/StakeSummary/StakeSummary.test.js
eniolam1000752: reason was because in the mock prop value, the validator name wasn't provided it was just the address.

src/modules/pos/validator/components/StakeForm/StakeForm.js
```diff
@@ -153,11 +167,46 @@ const StakeForm = ({ t, stakes, account, isStakingTxPending, nextStep, history,
...
+      setIsLoading(true);
+      const [error, dryRunResult] = await to(
+        dryRun({
+          skipVerify: true,
+          paramsSchema,
+          transaction,
+          serviceUrl: mainChainNetwork.serviceUrl,
+        })
+      );
+      setIsLoading(true);
```
oskarleonard: Do we need to call `setIsLoading(true)` twice in this function? See line 176 as well

src/modules/pos/validator/components/StakeForm/StakeForm.js
```diff
@@ -153,11 +167,46 @@ const StakeForm = ({ t, stakes, account, isStakingTxPending, nextStep, history,
...
+      setDryrunError(error);
```
oskarleonard: If we get an error we are still continuing with `nextStep`, is that correct?

src/modules/pos/validator/components/StakeForm/StakeForm.js
```diff
@@ -153,11 +167,46 @@ const StakeForm = ({ t, stakes, account, isStakingTxPending, nextStep, history,
     posToken
   );
   const showEmptyState = !changedStakes.length || isStakingTxPending;
+  const onConfirm = async (formProps, transactionJSON, selectedPriority, fees) => {
+    if (!showEmptyState && moduleCommandSchemas && !isLoading) {
+      const moduleCommand = joinModuleAndCommand(transactionJSON);
```
oskarleonard: Can we scope the reward logic (from line 172 to 205) in a function `const rewards = getRewards()` to easily see the logic in this `onConfirm`?

src/modules/pos/validator/components/StakeForm/StakeForm.js
```diff
@@ -153,11 +167,46 @@ const StakeForm = ({ t, stakes, account, isStakingTxPending, nextStep, history,
     posToken
   );
   const showEmptyState = !changedStakes.length || isStakingTxPending;
+  const onConfirm = async (formProps, transactionJSON, selectedPriority, fees) => {
+    if (!showEmptyState && moduleCommandSchemas && !isLoading) {
```
oskarleonard: the `!isLoading` condition seems redundant here? Isnt the confirm button disabled when `isLoading`? 
Now it reads like

> if isLoading (but this function is somehow called anyway)
> go to history.push(routes.validators.path);

Which feels wrong 🤔

src/modules/pos/validator/components/StakeForm/StakeForm.js
```diff
@@ -231,9 +284,9 @@ const StakeForm = ({ t, stakes, account, isStakingTxPending, nextStep, history,
                   />
```
oskarleonard, line 303: Not related to this line:

Maybe we should rename the `buttonTitle` label to 
`buttonTitle={showEmptyState ? t('Continue to validator list') : t('Continue')}`
Since that seems to be where it is navigating. 

I dont know in which scenario we show this component when showEmptyState. Do you know how to test that flow?

src/modules/transaction/components/StakeItem/StakeItem.js
```diff
@@ -17,11 +17,11 @@ import styles from './StakeItem.css';
  * @param {String} title text to use instead of the address e.g. validator username
  * @param {Boolean} truncate text to use instead of the address e.g. validator username
  */
-const StakeItem = ({ stake, address, title, truncate, token }) => {
+const StakeItem = ({ stake, address, title, truncate, token, reward }) => {
   const accountPath = routes.explorer.path;
 
   return (
-    <span className={styles.container}>
+    <span className={`${styles.container} stake-item`}>
```
oskarleonard: `stake-item` seems to only be used for testing purposes? Maybe we can use test id instead. Confusing to add classes just for testing purposes.

src/modules/transaction/components/StakeValidate/StakeValidator.js
ManuGowda: Screenshot 2023-06-29 at 3 51 07 PM
It can happen that sometimes there is no rewards to claim? so when `Total rewards` is `0` amount we can hide it to not confuse user?

eniolam1000752: hmm okay nice suggestion.

src/modules/transaction/components/StakeValidate/StakeValidator.js
```diff
@@ -17,43 +19,64 @@ const ItemList = ({ items, heading }) => {
...
+              'Total rewards credited to your account in effective to changes in stakes.'
+            )}
+          >
+            <span className={styles.totalReward}>
+              <TokenAmount val={rewards.total.toString()} token={token} />
+            </span>
+          </InfoColumn>
+        )}
       </div>
-      {/* <div className={styles.txFeeContainer}>
```
oskarleonard, line 52: Comment

oskarleonard (approved):

ManuGowda (approved):

src/modules/transaction/components/StakeValidate/StakeValidator.js
ManuGowda: It was a dead code.. i checked the UI the transaction fees are showned.

PR #5123, opened 2023-06-28
Stake page should update by itself

### LINKED ISSUE #5074: Stakes page should update by itself
The "Stakes" page doesn't update itself and is not showing votes users just made. The page should update itself if a vote was removed or added. 

Reproduce: Add or remove a vote and click immediately on "Show all your stakes". You will notice that nothing changes. Users would have to go to another page and then go to their Stakes page to see the changes. It should update by itself.

Lisk Desktop 3.0.0-beta.0
Where? Validators -> Stakes

oskarleonard:
### What was the problem?

This PR resolves #5074

### How was it solved?

- Listen for a users pos transactions and update stakes page for each new transaction.
- Show error status modal if broadcast fails.

### How to test?
1. Use an account with stakes and go to http://0.0.0.0:8080/#/validators/profile/stakes
2. Edit a stake successfully higher
3. Wait ~10 sec (Note: out socket works pretty bad so if nothing happens refresh page and redo the previous steps)
4. Expected: That stake amount should update and flash green briefly

### DIFF
--- a/src/modules/pos/validator/components/SentStakesRow/SentStakesRow.css
+++ b/src/modules/pos/validator/components/SentStakesRow/SentStakesRow.css
@@ -40,7 +40,7 @@
       font-size: var(--font-size-h6);
       transition-property: color;
       transition-timing-function: steps(8, end);
-      transition-duration: 0.5s;
+      transition-duration: 1.5s;
     }
 
     &.animateRed {
--- a/src/modules/pos/validator/components/SentStakesRow/SentStakesRow.css
+++ b/src/modules/pos/validator/components/SentStakesRow/SentStakesRow.css
@@ -28,6 +28,31 @@
   &:hover .pinWrapper {
     display: block;
   }
+
+  & .balance {
+    padding: 0px 10px 0px 10px;
+    display: flex;
+    align-items: center;
+    font-size: var(--paragraph-font-size-s);
+    color: var(--color-content-light);
+
+    &.amountCell {
+      font-size: var(--font-size-h6);
+      transition-property: color;
+      transition-timing-function: steps(8, end);
+      transition-duration: 0.5s;
+    }
+
+    &.animateRed {
+      color: var(--color-red);
+      transition-duration: 0s;
+    }
+
+    &.animateGreen {
+      color: var(--color-deep-green);
+      transition-duration: 0s;
+    }
+  }
 }
 
 .action {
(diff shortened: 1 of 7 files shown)


src/modules/pos/validator/components/SentStakesRow/SentStakesRow.css
```diff
@@ -28,6 +28,31 @@
...
+    display: flex;
+    align-items: center;
+    font-size: var(--paragraph-font-size-s);
+    color: var(--color-content-light);
+
+    &.amountCell {
+      font-size: var(--font-size-h6);
+      transition-property: color;
+      transition-timing-function: steps(8, end);
+      transition-duration: 0.5s;
```
ManuGowda: Increase it to 1.5s ? so that user can view the transition easily...

ManuGowda (commented):
The changes LGTM, left a comment and also as discussed please fix the `new.transactions` event subscribe and unsubscribe issue.

ManuGowda (dismissed):

src/modules/pos/validator/components/SentStakesRow/SentStakesRow.js
```diff
@@ -21,7 +25,14 @@ const SentStakesRow = ({ data: stakes, stakeEdited, token }) => {
         <Balance colSpanXs={1} value={`#${rank}`} />
         <Balance value={<TokenAmount val={validatorWeight} token={token} />} />
         <Balance value={`${commission / 100}%`} />
-        <Balance className={styles.amountCell} value={<TokenAmount val={amount} token={token} />} />
+        <Balance
+          className={classNames({
+            [styles.amountCell]: true,
+            [styles.animateGreen]: prevAmount && parseInt(prevAmount, 10) < parseInt(amount, 10),
+            [styles.animateRed]: prevAmount && parseInt(prevAmount, 10) > parseInt(amount, 10),
```
eniolam1000752: Shouldn't we use BigInt instead of parseInt

eniolam1000752 (changes requested):

src/modules/pos/validator/components/SentStakesRow/SentStakesRow.js
eniolam1000752:
```suggestion
            [styles.animateGreen]: prevAmount && BigInt(prevAmount) < BigInt(amount),
            [styles.animateRed]: prevAmount && BigInt(prevAmount) > BigInt(amount),
```

oskarleonard: True, fixed

ManuGowda (approved):

PR #5135, opened 2023-07-01
Fix staking amount validation error

### LINKED ISSUE #5122: Staking balance error should account for edit stake
### Expected behavior
When editing a stake higher than account balance we should only show below error if the edit stake difference is higher.
In the image below i edited the stake from 220 to 230, which is only a 10 LSK difference, and my account balance contains 225 LSK which means that i can afford to edit my stake 10 LSK higher.

### Actual behavior
image

### Steps to reproduce
1. Have an account with for example 100 LSK
2. Stake 50 LSK with validator X
3. Send 30 LSK to random address (not this accounts address)
4. Try to edit stake with validator X to 60 LSK

eniolam1000752:
### What was the problem?

This PR resolves #5122 

### How was it solved?

- [x] Fix amount validation in the `useStakeAmountField` hook
- [x] update unit tests

### How was it tested?

- Navigate to the validator's page
- Try to add or edit stakes on any validator
- Staking amount which should be the difference between the confirm and unconfired amount should now be properly validated against total wallet balance.


oskarleonard (approved):

ikem-legend (approved):

ManuGowda (approved):
Actually this still have issues with `validateAmount` function.. the amount comparison is incorrect, but for now to this specific issue it solves the problem hence approving.

PR #5148, opened 2023-07-04
Bugfix: IPC exposes overly sensitive functionality

### LINKED ISSUE #5113: IPC exposes overly sensitive functionality
### Expected behavior

expose every functionality (message that the renderer can send, or receive) in a specific function with a clear purpose. Follow Electron’s IPC tutorial to learn how to effectively send messages from the main process to the renderer and vice-versa. This will limit the renderer’s access to Electron’s APIs to the minimum.

### Actual behavior

The electron preload script exposes to the renderer application (the react app) the send , on , once , and removeListener of the ipcRenderer object. This is not recommended as it may expose a lot more functionality to the renderer than necessary.

### Steps to reproduce

https://github.com/LiskHQ/lisk-desktop/blob/8238f41dbae0ac55449149db1a19bf0b69cda7d1/app/src/ipc.js#L5-L24

### Reference

https://www.electronjs.org/docs/latest/tutorial/ipc

oskarleonard:
### What was the problem?

This PR resolves #5113

### How was it solved?
Restricting what we `exposeInMainWorld` from the electron IPC API. Dont allow the client to access functions such as `ipc.send`, `ipc.on`, etc. 

Unfortunately we must have duplicate `const` for the IPC channel names and events since electron doesn't allow import of custom modules into preload scripts (which our ipc is). Check https://github.com/electron/electron/issues/35587 for more info

### How was it tested?
Affected features that most be tested before merging this PR:
1. Hardware wallet
2. Updating electron
3. Locale (language)
4. localStorage


libs/hardwareWallet/ledger/ledgerDeviceListener/subscribeToLedgerDevices.js
```diff
@@ -3,5 +3,5 @@ import { LEDGER_HW_HID_EVENT } from '@libs/hardwareWallet/ledger/constants';
 const IPC = window.ipc;
 
 export const subscribeToLedgerDeviceEvents = (fn) => {
-  IPC.on(LEDGER_HW_HID_EVENT, (action, data) => fn(data));
+  IPC[LEDGER_HW_HID_EVENT]((action, data) => fn(data));
```
ManuGowda, line 6: `on` event is missing? and same below

app/src/ipc.js
```diff
@@ -1,24 +1,140 @@
...
+const IPC_DOWNLOAD_UPDATE_START = 'IPC_DOWNLOAD_UPDATE_START';
+const IPC_DOWNLOAD_UPDATE_PROGRESS = 'IPC_DOWNLOAD_UPDATE_PROGRESS';
+const IPC_DOWNLOAD_UPDATE_COMPLETED = 'IPC_DOWNLOAD_UPDATE_COMPLETED';
+const IPC_UPDATE_QUIT_AND_INSTALL = 'IPC_UPDATE_QUIT_AND_INSTALL';
+const IPC_OPEN_URL = 'IPC_OPEN_URL';
+const IPC_STORE_CONFIG = 'IPC_STORE_CONFIG';
+const IPC_CONFIG_RETRIEVED = 'IPC_CONFIG_RETRIEVED';
+const IPC_RETRIEVE_CONFIG = 'IPC_RETRIEVE_CONFIG';
+const IPC_SET_LOCALE = 'IPC_SET_LOCALE';
+const IPC_DETECT_LOCALE = 'IPC_DETECT_LOCALE';
```
ManuGowda, line 36: reuse from src/const/ipcGlobal?

src/const/ipcGlobal.js
```diff
@@ -0,0 +1,13 @@
+export const IPC_UPDATE_AVAILABLE = 'IPC_UPDATE_AVAILABLE';
```
ManuGowda, line 1: the file name can be `ipcGlobalEvents.js` so while importing it becomes easier to recognise?

ManuGowda (commented):
LGTM, tested on hardware wallet it worked, haven't tested the auto updater...

where possible please move the reused constants to const file and refer 🙏🏼

libs/hardwareWallet/ledger/ledgerDeviceListener/subscribeToLedgerDevices.js
ManuGowda:
```suggestion
  IPC[LEDGER_HW_HID_EVENT]((_, data) => fn(data));
```
you can do the same at other places?

src/const/ipcGlobal.js
oskarleonard: These const are used both as `ipc channel names` and `ipc events` depending on usage which is why i wanted to be more lose in the naming.

For example: 
**channel name**
image

**ipc event objcet**
image

app/src/ipc.js
oskarleonard: Unfortunately we must have duplicate consts for the IPC channel const since electron doesn't allow imports into preload scripts (which our ipc is) in newer versions of electron https://github.com/electron/electron/issues/35587

ManuGowda (dismissed):

ikem-legend (dismissed):

ManuGowda (approved):

ikem-legend (approved):

PR #5152, opened 2023-07-06
Bugfix: Remove unnecessary innerHTML usage

### LINKED ISSUE #5115: Unnecessary use of innerHTML
### Expected behavior

Modify the code to use .innerText instead of .innerHTML . This will accomplish the same without the risk of XSS.

Review every other HTML sink, and, where they are really necessary, add a comment explaining why, and why unsanitized attacker input cannot reach it.

### Actual behavior

The code unnecessarily uses the .innerHTML function to append simple text to the DOM, as shown in figure 59.1. This function allows users to modify the HTML of an element; however, in this case, this is unnecessary because the data-name attribute of icon elements is always simple text without HTML tags. Using .innerText is sufficient.

### Steps to reproduce

https://github.com/LiskHQ/lisk-desktop/blob/8238f41dbae0ac55449149db1a19bf0b69cda7d1/setup/react/assets/fonts/iconfont/icons.html#L497-L506

### Reference

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText
https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

eniolam1000752:
### What was the problem?

This PR resolves #5115

### How was it solved?

- [x] Replace innerHTML with innerText

### How was it tested?

- Serve this directory `setup/react/assets/fonts/iconfont/icons.html` (you can use Live server for ease)
- Clicking on any of the icons doesn't change its behaviour after `innerHTML` is changed to  `innerText`

### DIFF
--- a/setup/react/assets/fonts/iconfont/icons.html
+++ b/setup/react/assets/fonts/iconfont/icons.html
@@ -498,7 +498,7 @@
       (function () {
         document.getElementById('icons').onclick = function (e) {
           e = e || window.event;
-          var name =
+          const name =
             e.target.getAttribute('data-name') || e.target.parentNode.getAttribute('data-name');
           document.getElementById('name').innerText = name;
         };
--- a/setup/react/assets/fonts/iconfont/icons.html
+++ b/setup/react/assets/fonts/iconfont/icons.html
@@ -500,7 +500,7 @@
           e = e || window.event;
           var name =
             e.target.getAttribute('data-name') || e.target.parentNode.getAttribute('data-name');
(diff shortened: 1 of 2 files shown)


setup/react/assets/fonts/iconfont/icons.html
```diff
@@ -500,7 +500,7 @@ <h1>Usage</h1>
           e = e || window.event;
           var name =
```
ManuGowda: change this to const

ManuGowda (approved):

oskarleonard (approved):

PR #5174, opened 2023-07-13
Bug fix: Resolve issue with deviceDisconnectDialog

### LINKED ISSUE #5165: Phishing risk on the deviceDisconnectDialog modal
### Description

Part of the error message shown in the deviceDisconnectDialog modal is controlled by the model URL parameter. Developers expected this field to receive only the name of a device that was disconnected, but it can receive any message, including a phishing message.

### Recommendations

Short term, hardcode the list of supported devices that can be shown in the URL.
Alternatively, if the list of devices is unknown, limit the size of the message.

eniolam1000752:
### What was the problem?

This PR resolves #5165

### How was it solved?

- Removed <DeviceDisconnectDialog /> since its not being used in the application.

### How was it tested?

- Visually


oskarleonard (approved):

ikem-legend (approved):

ManuGowda (commented):
@eniolam1000752 i think in general for all the modals where we have params we can URL parse and if there is malicious data we can show error?

e.g `modal=stake& model=lisk network. Please visit malicious.com to learn how to fix the issue`

ManuGowda (approved):

PR #5176, opened 2023-07-13
Reject wallet connect request should show reject modal

### LINKED ISSUE #5164: lisk-desktop attempts to open a non-existent modal
### Description

The ConnectionSummary component tries to open the connectionSuccess modal which does not exist. The developer likely wanted to open the connectionStatus modal.

### Actual behavior

https://github.com/LiskHQ/lisk-desktop/blob/8238f41dbae0ac55449149db1a19bf0b69cda7d1/src/modules/blockchainApplication/connection/components/ConnectionSummary/index.js#L35-L40

### Recommendations

Short term, fix the issue by opening the correct modal.

Long term, instead of hardcoding a modal's name when opening it, use constants based on the modals defined in the routes.js file. This will remove the possibility of bugs such as the ones defined in this finding, and improve the code's quality.

oskarleonard:
### What was the problem?

This PR resolves #5164

### How was it solved?

Use correct modal

### How was it tested?

1. Reject connection request from wallet connect
2. Expected: Show a modal telling the user the connection was rejected


ikem-legend (dismissed):

src/modules/blockchainApplication/connection/components/ConnectionSummary/index.js
```diff
@@ -51,7 +51,7 @@ const ConnectionSummary = () => {
   const rejectHandler = async () => {
     const result = await reject();
     addSearchParamsToUrl(history, {
-      modal: 'connectionSuccess',
+      modal: 'connectionStatus',
```
eniolam1000752, line 54: Not related to this PR. 

I think the issue raised by the audit team concerning having pages that display information from query params as seen in this [PR](https://github.com/LiskHQ/lisk-desktop/pull/5174#issue-1802687812) would apply to this component. 
`<ConnectionStatus />` as seen from line `16 - 25`. I think we should take their suggestion by either limiting the `name` query param key or have a defined set of values. 

You can find the audit's suggestion in the audit final report `95. Phishing risk on the deviceDisconnectDialog modal`

eniolam1000752 (changes requested):

src/utils/urlUtils.js
```diff
@@ -0,0 +1,18 @@
+export function sanitizeTextFromDomains(text) {
+  if (!text) {
+    return '';
+  }
+  const trimmedText = text.slice(0, 30);
```
ManuGowda, line 5: why 30?

src/utils/urlUtils.js
```diff
@@ -0,0 +1,18 @@
...
+  const httpRegexG =
+    /[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b[-a-zA-Z0-9()@:%_+.~#?&/=]*/g;
```
ManuGowda: /[-\w@:%._+~#=]{1,256}\.\w{1,6}\b[-\w()@:%_+.~#?&/=]*/gi;

reason
The \w metacharacter matches word characters.

A word character is a character a-z, A-Z, 0-9, including _ (underscore).

but please check if it works

src/utils/urlUtils.js
oskarleonard: No specific reason, in the audit report they just recommended to limit it.

src/utils/urlUtils.js
oskarleonard: That seems to match it strangely 
image

And this should add the ['scam.io'] in the array so i know what to remove from the string
image

ManuGowda: `/[-\w@:%.+~#=]{1,64}\.\w{1,6}\b[-\w()@:%+.~#?&/=]*/gi`

src/utils/urlUtils.js
ManuGowda: ah ok.. then i guess you took the 30 length based on the url that wallet connect provides? but do you think it would be safe, just because of this hard length if we have problem then have to make a release just to allow more length for valid URLs 😩

ManuGowda (approved):

src/utils/urlUtils.test.js
```diff
@@ -0,0 +1,30 @@
...
+    const text = 'Web app scam.io Web';
+    const expected = 'Web app  Web';
+    expect(sanitizeTextFromDomains(text)).toEqual(expected);
+  });
+
+  it('Should remove multiple domains from a string', () => {
+    const text = 'Web scam.io phish.com';
+    const expected = 'Web  ';
+    expect(sanitizeTextFromDomains(text)).toEqual(expected);
+  });
```
eniolam1000752, line 35: we caw also have  a test that checks against urls of these format

- https://scam.io
- http://localhost:0000
- 192.168.10.1:9000

eniolam1000752 (changes requested):

src/utils/urlUtils.test.js
ManuGowda: lets merge this one and he has a two other PRs we can fix them there?

ikem-legend (approved):

src/utils/urlUtils.test.js
eniolam1000752: okay cool

eniolam1000752 (approved):

PR #5181, opened 2023-07-18
Resolve issue with send token amount validation

### LINKED ISSUE #5177: Bug in send token amount validation
### Expected behavior
- If an amount lesser than the wallet balance is inputed into the send token amount field, I should't get a `wallet balance amount exceeded` validation error.

### Actual behavior
Presently, if I have a wallet balance greater than the amount intend to send, I get an error stating `Provided amount is higher than your current balance` as seen in the image below

image

eniolam1000752:
### What was the problem?

This PR resolves #5177

### How was it solved?

- Convert amount and balance to BigInt in the validator
- Update unit tests

### How was it tested?

- Navigate to the send form route
- Ensure to have sufficient token balance(possible 20LSK)
- input token amount values of `1`, `3` there should be no form validation error derived.


ManuGowda (dismissed):

ikem-legend (dismissed):

package.json
```diff
@@ -139,7 +139,6 @@
     "html-entities": "1.4.0",
```
oskarleonard, line 139: NRTTL:
Please check if checked in files are prettified

oskarleonard: /walletVisual/index.js is not for example

oskarleonard (approved):

ManuGowda (approved):

ikem-legend (approved):

PR #5199, opened 2023-07-21
Send token button should be disabled when available balance is 0

### LINKED ISSUE #5197: Send token button should be disabled when available balance is 0
### Expected behavior

Send token button should be disabled when available balance is 0

### Actual behavior

User is able to open the send form when an account do not have available balance

### Steps to reproduce

- Create an account
- Add account to wallet
- `Send` button is not disabled

ManuGowda:
### What was the problem?

This PR resolves #5197

### How was it solved?

- Disable `send` button
- Removed unused components and functions

### How was it tested?

- Create an account with 0 balance


src/modules/blockchainApplication/manage/hooks/useSearchApplications.js
```diff
@@ -18,7 +25,7 @@ export const useSearchApplications = () => {
   const setDebounceSearch = (value) => {
     setSearchValue(value);
     clearTimeout(timeout.current);
-    // Validate the URL with debouncer
+    // Validate the URL with denouncer
```
oskarleonard: isnt it `debouncer` ?

src/modules/wallet/components/AllTokens/AllTokens.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable max-statements */
 import React, { useCallback, useMemo, useRef, useState } from 'react';
+import { useLocation } from 'react-router';
```
oskarleonard: I think we should only import from  `react-router-dom`, and when migration is complete remove, `react-router` completely from our package.json. `react-router-dom` is a super-set of `react-router`, no need for both

src/modules/wallet/components/AllTokens/AllTokens.js
```diff
@@ -29,8 +28,13 @@ const AllTokens = ({ history }) => {
   const { t } = useTranslation();
   const timeout = useRef();
 
-  const [search, setSearch] = useState('');
+  const [searchToken, setSearch] = useState('');
   const { filters, setFilter } = useFilter({});
+  const { search } = useLocation();
+  const queryParams = new URLSearchParams(search);
+  const disableSend = Boolean(queryParams.get('disableSend'));
```
oskarleonard: Doesnt this query param contain a string representation of 'true' or 'false'? 
In that case we cannot rely on `Boolean(` since 
image

Better to do:
`const isSendDisabled = queryParams.get('disableSend') === 'true';`
🤔

oskarleonard: As you can see here: 
image

ikem-legend (dismissed):

ikem-legend (approved):

oskarleonard (approved):

PR #5200, opened 2023-07-21
Sign message wallet connect

### LINKED ISSUE #5102: Sign message from wallet connect
### Expected behavior
It should be possible to initiate a sign message from wallet connect

### Actual behavior
Shows an error view "missing schema"

### Steps to reproduce
Try signing a transaction from wallet connect

oskarleonard:
### What was the problem?

This PR resolves #5102

### How was it solved?

Added sign message flow for wallet connect

### How was it tested?

1. sign_message from wallet connect
2. Opens sign message flow in lisk desktop
3. Expected: Should work to sign message


setup/react/app/MainRouter.js
```diff
@@ -15,8 +15,16 @@ const MainRouter = ({ history }) => {
...
+    const event = events.length && events[events.length - 1];
+
+    if (event.name === EVENTS.SESSION_REQUEST) {
+      const method = event.meta?.params?.request?.method;
+
+      if (method === 'sign_message') {
+        addSearchParamsToUrl(history, { modal: 'requestSignMessageDialog' });
+      } else {
+        addSearchParamsToUrl(history, { modal: 'requestView' });
+      }
```
ManuGowda, line 27:
```suggestion
      if (method === 'sign_message') {
        addSearchParamsToUrl(history, { modal: 'requestSignMessageDialog' });
      } else if (method === 'sign_transaction') {
        addSearchParamsToUrl(history, { modal: 'requestView' });
      } else {
        //Show Unsupported method and return the user back
      }
```

eniolam1000752 (changes requested):
Issues observed

- Text not properly styled in dark mode for sign message
image

- on sign message, I was able to freely adjust the dimension of the text area. I think we should lock the width and adjustable height of the textarea
image

- Are there situations where a user could sign more than on transactions in a single wallet connect request. If yes then I think if I only sign a single transition instead of me getting an array when I click on copy signatures, I should just get the signature without the square brackets. Also I think if there is only one tx been signed, I should see `copy signature `not `copy signatures`
image

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
```diff
@@ -160,7 +160,7 @@ const RequestSummary = ({ nextStep, history }) => {
             )}
           </ValueAndLabel>
           {!errorMessage && (
-            <ValueAndLabel className={styles.labeledValue} label={t('Selected account')}>
+            <ValueAndLabel className={styles.labeledValue} label={t('Sender account')}>
```
ManuGowda:
```suggestion
            <ValueAndLabel className={styles.labeledValue} label={t('Signing account')}>
```
As discuss

ManuGowda (changes requested):
LGTM, except eniolas comments

eniolam1000752 (approved):

ManuGowda (approved):

PR #5202, opened 2023-07-24
Unused route property and un-routed modal

### LINKED ISSUE #5194: Lisk-desktop: unused route property
### Actual behavior

The isSigninFlow property of routes is never used.

oskarleonard:
### What was the problem?

This PR resolves #5194 and #5240

### How was it solved?

Removal of unused logic

### How was it tested?

Manually

### DIFF
--- a/src/routes/routes.js
+++ b/src/routes/routes.js
@@ -66,7 +66,6 @@ export default {
   register: {
     path: '/register',
     isPrivate: false,
-    isSigninFlow: true,
     forbiddenTokens: [],
   },
   addAccountByFile: {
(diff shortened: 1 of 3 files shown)


src/routes/routes.js
```diff
@@ -216,10 +214,6 @@ export const modals = {
     isPrivate: false,
     forbiddenTokens: [],
   },
-  setPassword: {
```
ManuGowda, line 219: can you cross check if there such routes which are redundant?

ManuGowda (commented):
rest of the changes LGTM

src/routes/routes.js
oskarleonard: Yes, i couldnt find that such a route is invoked anywhere. Nor do we have such a component

ikem-legend (approved):

ManuGowda (approved):

PR #5214, opened 2023-07-28
Bugfix: Lisk-desktop: Incorrect setState callbacks

### LINKED ISSUE #5191: Lisk-desktop: Incorrect setState callbacks
### Actual behavior

Several setState callbacks use this.props or this.state , which is incorrect as explained in React’s documentation . Using the js/react/inconsistent-state-update CodeQL rule , we found the following problematic instances:

- lisk-desktop/src/modules/bookmark/components/BookmarksList/Bookmarks
List.js#L89-L92
- lisk-desktop/src/modules/wallet/components/passphraseRenderer/index.js#
L126-L133
- lisk-desktop/src/modules/wallet/components/searchBarWallets/passphraseR
enderer/index.js#L126-L133
- lisk-desktop/src/theme/tabs/tabsContainer/tabsContainer.js#L49-L54
- lisk-desktop/src/theme/Tooltip/tooltip.js#L32-L34
- lisk-desktop/src/utils/withData.js#L113-L120
- lisk-desktop/src/utils/withFilters.js#L39-L44
- lisk-desktop/src/utils/withFilters.js#L65-L76
- lisk-desktop/src/utils/withLocalSort.js#L26-L28

ikem-legend:
### What was the problem?

This PR resolves #5191

### How was it solved?

- Refactored valid use cases
- Removed unused files

### How was it tested?

- Manually


src/modules/block/components/blocks/blocks.js
```diff
@@ -22,41 +22,48 @@ export const defaultFilters = {
 };
 
 // eslint-disable-next-line max-statements
-const Blocks = ({ filters, applyFilters, clearFilter, clearAllFilters, sort, changeSort }) => {
+const Blocks = () => {
```
oskarleonard, line 25: Changes here probably introduces regression, in that filters are not cleared. 
1. Add height filter (or any other filter)
2. Remove height filter by clicking x
3. Expected: Height field in filter dropdown should be cleared

(Same for remove all filters)

ikem-legend: It seems like a general problem as the same behaviour exists in all the other places where the hook filter is used. We'll need to create a ticket to fix them all

src/modules/wallet/components/passphraseRenderer/index.js
```diff
@@ -130,24 +130,24 @@ class PassphraseRenderer extends React.Component {
     const otherIndex = indexes.find((index) => index !== selectedIndex);
     const shouldDisplayOptions = Object.values(chosenWords).length < 2;
 
-    this.setState({
-      ...this.state,
+    this.setState((state) => ({
+      state,
```
oskarleonard: This will result in a state object inside a state object. 
We need to spread the state to prevent buggy behaviour like, the choose words still showing when button enabled. 
image


`...state,`

src/modules/block/components/blocks/blocks.js
oskarleonard: This problem doesnt exists in the `release/3.0.0` branch though, feels strange to introduce a new bug. 

Problem with this behaviour is that after cleaning and values are still in the dropdown the apply button does nothing.

src/modules/wallet/components/passphraseRenderer/index.js
ikem-legend: Fixed

src/modules/block/components/blocks/blocks.js
ikem-legend: It exists in the other pages like Transactions monitor that use the useFilter hook. That's what I was referring to as a general problem

oskarleonard (approved):

eniolam1000752 (approved):

PR #5221, opened 2023-08-02
Incorrect js docs

### LINKED ISSUE #5190: Lisk-desktop: incorrect JDocs
### Actual behavior

Lisk-desktop makes use of JDocs to document functions; however several are malformed, missing, or include unknown parameters. 

### Steps to reproduce
Use the following CodeQL rules to find and fix these problems:
-  js/jsdoc/malformed-param-tag : Finds parameters missing a description
- js/jsdoc/missing-parameter : Finds undocumented parameters
- js/jsdoc/unknown-paramete r: Finds documented parameters that are not function argument

oskarleonard:
### What was the problem?

This PR resolves #5190

### How was it solved?

Removed most of our Js docs since much of it is deprecated and difficult to have up to date. Future focus will be to have self documented code with type script.

### How was it tested?

Manually


src/modules/pos/validator/store/actions/staking.js
```diff
@@ -45,15 +45,9 @@ export const stakeDiscarded = (data) => ({
 
 /**
  * Defines the new stake amount for a given validator.
- * The reducer will add a new stake if if didn't exist before
+ * The reducer will add a new stake if i didn't exist before
```
ManuGowda:
```suggestion
 * The reducer will add a new stake if it didn't exist before
```

ManuGowda (dismissed):

ManuGowda (approved):

src/modules/auth/store/action.js
```diff
@@ -5,32 +5,11 @@ import actionTypes from './actionTypes';
...
- *
- * @param {Object} data - for hardware wallets it contains publicKey and hwInfo,
- *    otherwise contains passphrase
- * @param {String} data.passphrase - BIP39 passphrase of the account
- * @param {String} data.publicKey - Lisk publicKey used for hardware wallet login
- * @param {Object} data.hwInfo - info about hardware wallet we're trying to login to
- */
-
-/**
- * Store second passphrase in the Redux store
```
eniolam1000752, line 19: I think we can still retain this line

src/modules/auth/store/action.js
```diff
@@ -5,32 +5,11 @@ import actionTypes from './actionTypes';
...
- * @param {string} passphrase - Valid Mnemonic passphrase
- * @returns {object} Pure action object
- */
 export const secondPassphraseStored = (passphrase) => ({
   type: actionTypes.secondPassphraseStored,
   data: passphrase,
 });
 
-/**
- * Removes the second passphrase from the Redux store
```
eniolam1000752, line 30: can 't we also retain this

src/modules/blockchainApplication/explore/hooks/queries/useBlockchainApplicationExplore.js
```diff
@@ -2,24 +2,6 @@ import { BLOCKCHAIN_APPS } from 'src/const/queries';
 import { LIMIT as limit, API_VERSION } from 'src/const/config';
 import { useCustomInfiniteQuery } from 'src/modules/common/hooks';
 
-/**
- * Creates a custom hook for blockchain applications list queries
```
eniolam1000752, line 6: can't we retain this

src/modules/blockchainApplication/explore/hooks/queries/useBlockchainApplicationStatistics.js
```diff
@@ -2,17 +2,6 @@ import { BLOCKCHAIN_APPS_STATISTICS } from 'src/const/queries';
 import { API_VERSION } from 'src/const/config';
 import { useCustomQuery } from 'src/modules/common/hooks';
 
-/**
- * Creates a custom hook for blockchain applications meta queries
```
eniolam1000752, line 6: can't we retain this too

eniolam1000752 (changes requested):

src/modules/auth/store/action.js
oskarleonard: To me all of these descriptions are redundant.
Imo, it is better to have self documenting code, which for these cases i feel that we have.
File is in a folder called /store (so i interpret redux) and the file is called action (so i interpret functions here modifies the redux state).

How do you feel the comment brings value?

I guess `setSecondPassphrase` would have been a better name, but still `secondPassphraseStored` is understandable.

src/modules/blockchainApplication/explore/hooks/queries/useBlockchainApplicationExplore.js
oskarleonard: But we agreed that we should remove all this params because too many docs in the code base have incorrect values. I think it gives a false sense of security, better to just look up swagger when i need to know what parameters i can send

oskarleonard: Feels strange to leave it in some places and remove it in other places is mostly my concern :/

eniolam1000752 (approved):

PR #5253, opened 2023-08-16
Bugfix related to transaction overview, dryrun and error report

### LINKED ISSUE #5251: Transaction overview doesn't show statistics
### Expected behavior

Transaction overview should show stats even when an account is not added

### Actual behavior

Screenshot 2023-07-07 at 12 09 54 PM

### Steps to reproduce

Download the wallet and ensure you do not have any accounts added to wallet. Navigate to transactions tab and observer the stats.

3.0

ManuGowda:
### What was the problem?

This PR resolves #5251, #5256 and #5257

### How was it tested?

- Overview bug: open app in incognito window without any accounts and check if you can view all the stats of transaction
- Dry run: Use transaction composer to modify the fee to be lower and other kinds of scenario to test how dry run results are handled
- Error report: Send a transaction that is invalid and check if Retry works and also report email should show valid information like chain name, id, url and transaction and error message when exists.


src/modules/legacy/components/Summary/Summary.js
```diff
@@ -34,8 +34,11 @@ const Summary = ({ balanceReclaimed, nextStep, wallet, t, fees }) => {
     }),
     [wallet.legacy?.balance, module, command, wallet.sequence?.nonce, wallet.summary?.publicKey]
   );
+  const { data: networkFees } = useFees();
+
   const { initializationFees } = useGetInitializationFees({
     address: wallet.summary?.address,
+    tokenID: networkFees?.data?.feeTokenID,
   });
```
eniolam1000752: aren't we suppose to invoke the fee estimate endpoint here. Though I think the above still works though

src/utils/validators.js
```diff
@@ -110,6 +114,19 @@ export const validateAmount = ({
         return MIN_ACCOUNT_BALANCE > BigInt(accountBalance) - BigInt(amountInBase);
       },
     },
+    FEE_RANGE: {
+      message: i18n.t(
+        `Fee must be greater than ${convertFromBaseDenom(
+          minValue,
+          token
+        )} and less than ${convertFromBaseDenom(accountBalance, token)}.`
+      ),
```
eniolam1000752: shouldn't this be provided via interpolation

```suggestion
      message: i18n.t(
        `Fee must be greater than {{ maxFee }} and less than {{ minFee }}.`,
        {
          minFee: convertFromBaseDenom(accountBalance, token),
          maxFee: convertFromBaseDenom(
            minValue,
            token
          ),
        }
      ),
```

eniolam1000752 (changes requested):

eniolam1000752 (dismissed):

ikem-legend (approved):

eniolam1000752 (approved):

PR #5263, opened 2023-08-24
Fix issues raised on beta.1

### LINKED ISSUE #5229: Small bug findings from release v3.0
### Expected behavior
1. Screenshot 2023-08-08 at 10 59 09 AM
2. Allowed modals should open
3. Use metadata endpoint to search for token. Also do not show the search option when we have less than 10 token balances.
4. Do not show decimals
5. Extend the search to address and name
6. Fix the typo

### Actual behavior
1. Wallet page some items in table have duplicate dates
4. Cannot open request token modal if Lisk application is closed/open. Try to open this link when Lisk app is closed,  lisk://wallet?modal=send&recipient=lskyxq3carbt6ev86k2v2th9js5qbuewzezkz8gtv&amount=10&recipientChain=04000000. 
8. Wallet page / View all token/ Search token not working
9. Validators reporting page. Validators information graph show half validators. Show only integers.
Screenshot 2023-08-08 at 11 12 27 AM

10. Generators/ Validors search. Extend placeholder to be "Search for name or address"
11. Transaction monitoring page, Should be PoS Staking not Pos Staking.

eniolam1000752:
### What was the problem?

This PR resolves #5229

### How was it solved?

- Fixed issues raised on #5229

### How was it tested?

Follow steps on #5229


src/modules/pos/validator/components/ValidatorsMonitorView/Validators.js
```diff
@@ -196,7 +196,7 @@ const ValidatorsMonitor = ({ watchList }) => {
               value={search}
               className={`${styles.filterValidators} filter-by-name`}
               size="m"
-              placeholder={t('Search by name')}
+              placeholder={t('Search for name or address')}
```
ManuGowda:
```suggestion
              placeholder={t('Search by name or address')}
```

src/modules/transaction/components/TransactionRow/components.js
```diff
@@ -25,6 +25,10 @@ import TransactionRowContext from '../../context/transactionRowContext';
 import TransactionTypeFigure from '../TransactionTypeFigure';
 import TransactionAmount from '../TransactionAmount';
 
+const remapModuleCommandNames = {
+  [MODULE_COMMANDS_NAME_MAP.stake]: 'PoS:stake',
+};
```
ManuGowda: I think we don't need to do this, this shouldn't been considered bug at first place 🤔 

the reason is SDK always return `pos` as module so we can not handle such module names specifically? lets remove it as bug and remove this logic?

ManuGowda (approved):

ikem-legend (approved):

PR #5267, opened 2023-08-25
Fix avatar chosen during account creation

### LINKED ISSUE #5236: The avatar choosed during account creation is not used.
### Expected behavior
The avatar choose when creating account should be generated from the "m/44'/134'/0'" address.

### Actual behavior
The avatar choose when creating account will be generated from the legacy address.

### Steps to reproduce
1. Select avatar.
image

2. After creating an account. (different avatar)
image

3. Sign in legacy account. (same avatar)
image

beta.1
Windows 11

ikem-legend:
### What was the problem?

This PR resolves #5236

### How was it solved?

- Updating avatar logic to include custom derivation settings

### How was it tested?

- Unit tests
- Manually


src/modules/auth/components/ChooseAvatar/chooseAvatar.js
```diff
@@ -1,103 +1,113 @@
...
-    super();
+// eslint-disable-next-line max-statements
+const ChooseAvatar = ({ accounts, selected, handleSelectAvatar, nextStep }) => {
+  const [deselect, setDeselect] = useState({});
+  const [accountList, setAccountList] = useState([]);
+  const prevDeselect = useRef({});
+  const wrapperRef = useRef();
+  const timeout = useRef();
+  const { t } = useTranslation();
+  const { enableAccessToLegacyAccounts } = useSettings('enableAccessToLegacyAccounts');
```
ManuGowda, line 21: Since all new accounts are created with default custom derivation path, you do not need this logic, `enableAccessToLegacyAccounts` is false by default.

src/modules/auth/components/ChooseAvatar/chooseAvatar.js
```diff
@@ -1,103 +1,113 @@
...
-    this.setWrapperRef = this.setWrapperRef.bind(this);
-  }
+  useEffect(() => {
+    if (accounts.length) {
+      const getAcctList = async () => {
+        await Promise.all(
+          accounts.map(async (account) => {
+            const options = {
+              passphrase: account.passphrase,
+              enableAccessToLegacyAccounts,
```
ManuGowda, line 33: remove this, it will be false by default.

src/modules/auth/components/ChooseAvatar/chooseAvatar.js
oskarleonard: Noticed a strange scenario which might affect the intended behaviour regarding this.
1. Go to http://0.0.0.0:8080/#/account/add/secret-recovery
2. Check `Enable access to legacy Lisk accounts` 
3. Go to http://0.0.0.0:8080/#/register?strength=128
4. Continue the flow and the user will encrypt the new account with this API `cryptography.legacy.getKeys(passphrase)`.

ikem-legend: @ManuGowda The `extractKeyPair` function requires `enableAccessToLegacyAccounts` and even if it's false by default, changing it affects the account generation. The bug occurred because it wasn't taken into account previously

ikem-legend: @oskarleonard I followed the steps and it's fine because prior to the SDK update, accounts were generated based on that function

oskarleonard: What i mean was that the account would in that case not be created with the default custom derivation Manu talked about, but instead with our legacy one. Which is why it wont work to remove it in this step.

If feels strange that we removed the `enableAccessToLegacyAccounts` from settings if it can be used in this way. Feels like ✅  "Enable access to legacy Lisk accounts" should not be preserved when creating a new account. Ie that the `enableAccessToLegacyAccounts` state should only be used when adding an account from Secret recovery phrase 🤔

ManuGowda: @oskarleonard new account created should always be using custom derivation path, ikem now has fixed it

src/modules/auth/components/ChooseAvatar/chooseAvatar.js
ikem-legend: @ManuGowda Done

src/modules/auth/components/ChooseAvatar/chooseAvatar.js
oskarleonard: @ManuGowda It will still be possible to create a new account using this legacy API `cryptography.legacy.getKeys(passphrase)`. Is that intended?

src/modules/auth/components/ChooseAvatar/chooseAvatar.js
```diff
@@ -1,103 +1,113 @@
...
+              derivationPath: defaultDerivationPath,
+            };
+            const { publicKey } = await extractKeyPair(options);
+            const address = extractAddressFromPublicKey(publicKey);
+            return { ...account, address };
+          })
+        ).then((list) => {
+          setAccountList(list);
+        });
```
eniolam1000752, line 42:
```suggestion
        ).then(setAccountList);
```

eniolam1000752 (changes requested):

PR #5276, opened 2023-08-30
Hardware wallet account management

### LINKED ISSUE #5237: Improve hardware wallet account management
### Description

- The hardware wallet accounts are fetch frequently, optimise the account fetching
  - Switch account shouldn't re-render hardware wallet accounts all the time 
  - Fetch and save the accounts once and invalidate the cache based on account availability onchain. Check the status of all hardware wallet account status with batch query.  
- The hardware wallet accounts are rendered incorrectly at times, we can resolve this by using new interface
Screenshot 2023-08-07 at 11 55 48 AM

oskarleonard:
### What was the problem?

This PR resolves #5237

### How was it solved?

Adding a separate modal to import hardware account from

### How was it tested?

1. `git pull`, `yarn build`, `yarn dev`, `LISK_DESKTOP_URL="http://localhost:8080" DEBUG=true yarn run start`
2. Connect HW and enter pin
3. Click "Import Account" from the Hw Status icon in the topbar
4. **Expected**: Show a modal asking user to open the LiskApp
5. Open Lisk App
6. **Expected**: 
 1 Show Loading state 
 2 Show 3 Accounts
7. Click Load More
8. **Expected**: 
 1 Show Loading state 
 2 Show 6 Accounts
9. Click on 2nd account to import it
10. **Expected**: 
 1 Modal closes 
 2  Imported account becomes current account
11. Rename the HW account to "Test" and Click "Import Account" from the Hw Status icon in the topbar
12. **Expected**:   
 1 Show Loading state
 2 Show 3 Accounts where the second account is named "Test"
13. Close modal and Try to do a send transaction
14. **Expected**: It should work to send a transaction


src/modules/hardwareWallet/hooks/useHWAccounts.js
```diff
@@ -1,74 +1,57 @@
...
+            const accountIndex = index + 1;
+            const name =
+              getNameFromAccount(address, persistedHwAccounts) || `Account ${accountIndex}`;
+
+            return {
+              hw: currentHWDevice,
+              metadata: {
+                address,
+                pubkey: pubKey,
+                accountIndex,
```
ManuGowda:
```suggestion
                accountIndex: index,
```
otherwise the index used while signing will be wrong.

ManuGowda (commented):
The current load more keeps fetching accounts? imaging this
account1 was initialised and then randomly someone initialises account 10? then its hard to follow which account is initialised or not?

src/modules/hardwareWallet/hooks/useHWAccounts.js
oskarleonard: Nice catch! 
Btw, do you think the default name for the first account should be Account 0 or Account 1?

oskarleonard:
> The current load more keeps fetching accounts? imaging this account1 was initialised and then randomly someone initialises account 10? then its hard to follow which account is initialised or not?

Yes this is one of the pitfalls (it takes time to load accounts and is is not possible to ask the HW wallet for used accounts). They cannot though "easily" randomly initialises account 10, they would need to click "Load more" a couple of times, and the loading time is pretty long so this will prevent most users from doing this (if they dont really want to do that)

src/modules/hardwareWallet/hooks/useHWAccounts.js
ManuGowda: Account 1 is fine i think

ManuGowda:
> > The current load more keeps fetching accounts? imaging this account1 was initialised and then randomly someone initialises account 10? then its hard to follow which account is initialised or not?
> 
> Yes this is one of the pitfalls (it takes time to load accounts and is is not possible to ask the HW wallet for used accounts). They cannot though "easily" randomly initialises account 10, they would need to click "Load more" a couple of times, and the loading time is pretty long so this will prevent most users from doing this (if they dont really want to do that)

Hmmm I guess if we showed the balance for each account then that would have given an opportunity for users to know which account has balance and which one to use when. I guess you are eliminating the limitation for technical feasibility. I guess with this approach its more responsibility for users to ensure they use the correct account at correct index. 

In the past we dint had batch fetch so we had issues, so with this approach we would be ok to load more quickly and that shouldn't be a problem.

oskarleonard:
@ManuGowda > 
Yes, a big improvement would be to add balances, which i think we should do that eventually. I also think LiskDesktop should be as dynamic/non-limiting as possible. 
If a user uses Account 10 on another Lisk wallet, LiskDesktop should be able to import that wallet easily (which would be very difficult before this PR, they would basically need to add some Lisk to each wallet < 10).

If we are going to improve this i think these incremental steps would be the best:

1. Allow as much as possible, basically allow for browsing X nr of wallets (This PR)
2. Show Balance for accounts.
3. Allow a user to search for account by index. Maybe they know which derivationPath they used when adding a wallet.
4. Add a feature to loop through X amounts of accounts (for example 1-100), find the used ones, ask if a user wants to import all used accounts from 1-100.

src/modules/hardwareWallet/components/HardwareAccountManagerModal/HardwareAccountManagerModal.js
```diff
@@ -0,0 +1,78 @@
...
+  const currentHWDevice = useSelector(selectCurrentHWDevice);
+  const { isAppOpen } = currentHWDevice;
+
+  function onSelect(account) {
+    dispatch(updateHWAccount(account));
+    setCurrentAccount(account);
+  }
+
+  function onLoadMore() {
+    setNrOfAccounts((prevState) => prevState + 3);
```
ManuGowda: define the number as constant which can be later easily modified from one place

ManuGowda (commented):
LGTM

ManuGowda (commented):
Screenshot 2023-08-31 at 2 53 57 PM
Screenshot 2023-08-31 at 8 54 12 AM

Observed these two problems

for the second problem its difficult to reproduce, but its around memory leak

Screenshot 2023-08-31 at 2 57 13 PM

Please check the dark mode

src/modules/hardwareWallet/components/HardwareAccountManagerModal/HardwareAccountManagerModal.js
oskarleonard: https://github.com/LiskHQ/lisk-desktop/pull/5276/commits/01cba24e9db045e1ea9796abcc3a4b7ffc46d9f7

ManuGowda (approved):

PR #5282, opened 2023-08-31
Handle add/switch account flow on WC message signing method

### LINKED ISSUE #5278: Handle wallet connect switch and add account user journey on message signing
### Description

A user can request to sign a message from Lisk App, where a signing account is not the same as Lisk Desktop current account or the signing account might not exists.

### Acceptance Criteria

- Allow users to switch from current account to signing account
- Inform users to add signing account if it doesn't exists in Lisk Desktop

eniolam1000752:
### What was the problem?

This PR resolves #5278

### How was it solved?

- [x] Added `RequstSummary` component to the RequestSignMessageDialog multi step
- [x] Refactor `RequestSummary` to work properly in message mode
- [x] Fix failing unit tests

### How was it tested?

- Request message signing via wallet connect from an external dApp.
- Test for scenarios if signing account is not available in the user's account list.
- Test for scenarios if signing account is not the current account.
- Try to sign a wallet connect request message using the hardware wallet


ManuGowda (dismissed):

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
```diff
@@ -47,31 +48,29 @@ export const rejectLiskRequest = (request) => {
...
-  const { getAccountByAddress } = useAccounts();
+  const { getAccountByAddress, accounts } = useAccounts();
   const [currentAccount, setCurrentAccount] = useCurrentAccount();
   const { events } = useEvents();
   const [request, setRequest] = useState(null);
   const [transaction, setTransaction] = useState(null);
   const [senderAccount, setSenderAccount] = useState(null);
   const [errorMessage, setErrorMessage] = useState('');
-  const { sessionRequest } = useSession();
+  const { sessionRequest } = useSession({ isEnabled: !message });
```
oskarleonard, line 60: If we have a message this is not enabled 🤔 ?

libs/wcm/hooks/useSession.js
```diff
@@ -119,7 +119,7 @@ export const useSession = () => {
   );
 
   useEffect(() => {
-    if (signClient?.session && !hasLoaded) {
+    if (signClient?.session && !hasLoaded && isEnabled) {
```
oskarleonard, line 122: Shouldnt we loadSessions if isEnabled become enabled? Dont we need it in the dep array in that case?

oskarleonard (dismissed):

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
eniolam1000752: yhea reason is because we don't need it to be invoked for the message view

ManuGowda (approved):

PR #5296, opened 2023-09-06
Mailto domain is not working

### LINKED ISSUE #5293: Mailto domain is not working
### Expected behavior

Users should be able to report the error via email. Support `mailto` domain

### Actual behavior
Screenshot 2023-09-05 at 10 48 10 AM

### Steps to reproduce

- Simulate transaction failure scenario
- Click on `Report the error via email` the email is not triggered.

eniolam1000752:
### What was the problem?

This PR resolves #5293

### How was it solved?

- [x] Refactor white list to consider protocol
- [x] Fix unit test

### How was it tested?

- Open a mailto link (this can be gotten on the status page of a failed transaction flow)
- Clicking on the mailto link should pop up the default mail application of the system


app/src/modules/win.js
```diff
@@ -92,9 +92,13 @@ const win = {
     const handleRedirect = (e, url) => {
       try {
         const urlData = new URL(url);
-        const isAllowedUrl = WHITE_LISTED_DOMAIN.includes(urlData.origin);
 
-        if (!isAllowedUrl) return e.preventDefault();
+        const isUrlAllowed = WHITE_LISTED_URLS.some(
```
oskarleonard: Couldnt we have a function for this `isUrlAllowed(url)` which we import here instead of `WHITE_LISTED_URLS`? Bc the WHITE_LISTED_URLS  is a pretty complex object but the important part is the implementation which would be shown easily in a function.

app/src/modules/win.js
```diff
@@ -2,7 +2,7 @@ import localeHandler from './localeHandler';
 import menu from '../menu';
```
oskarleonard, line 2: NRTTL:
Adding this 
```
        <a target="_blank" href="mailto:mobiledev@lisk.com" rel="noopener noreferrer">
          mailto:mobiledev@lisk.com
        </a>
 ```
 It still opens a new window when clicked, is that supposed to happen?

Comment (eniolam1000752) [reply app/src/modules/win.js:2]: yhea reason is because the navigation events on electron isn't triggered when the `<a />`  is triggered

**Review by oskarleonard — DISMISSED** [review]

File: app/src/utils.js [hunk 3]
```diff
@@ -10,7 +10,19 @@ export const WHITE_LISTED_DEEP_LINKS = [
...
-export const WHITE_LISTED_DOMAIN = ['https://lisk.com'];
+const WHITE_LISTED_URLS = [
+  { protocol: 'https:', urlKey: 'host', domains: ['lisk.com'] },
+  { protocol: 'mailto:', urlKey: 'pathname', domains: ['desktopdev@lisk.com'] },
+];
+
+export const isUrlAllowed = (url) => {
+  const urlData = new URL(url);
+
+  WHITE_LISTED_URLS.some(
```
oskarleonard: we need return here :D. So maybe we say
`getIsUrlAllowed()`?
Will naming be easy
`const isUrlAllowed = getIsUrlAllowed(url)`?

oskarleonard (dismissed):

ManuGowda (dismissed):

ManuGowda (approved):

oskarleonard (approved):

setup/config/webpack.config.prod.js
```diff
@@ -31,6 +32,10 @@ const config = {
       PRODUCTION: JSON.stringify(true),
       VERSION: JSON.stringify(version),
     }),
+    new SubresourceIntegrityPlugin({
```
oskarleonard, line 35: Shouldnt we disable this in devmode?

ManuGowda: This is only triggered for prod? so in dev mode it wont call this as we have .dev.js config?

Issue #5273, opened 2023-08-29
Adapt argon2id protocol recommendation


ManuGowda:
### Description

Adapt encryption and decryption as per recommendation. Bump lisk-client to latest version

Refer to the LIP changes to the protocol: https://github.com/LiskHQ/lips/pull/465/files

##### Argon2id for Wallets

As some mobile devices have less memory, we recommend to use the second recommended option of RFC 9106 which requires less memory but uses more iterations instead:

* iterations=3,
* parallelism=4 lanes,
* memory=65536 KiB (64 MiB),
* 16 bytes salt,
* 32 bytes output.

ManuGowda:
Refer: https://github.com/LiskHQ/lips/pull/465

PR #5298, opened 2023-09-07
Implemented reward estimations while staking a validator

### LINKED ISSUE #5295: Implement reward estimations for staking a validator
### Description

Enable users to estimate the rewards while staking.

### Motivation

As a network participant I want to stake tokens for a validator "validator 1", the wallet should provide an ability to get the expected reward for a daily, monthly and yearly.

### Additional Information

- getExpectedValidatorRewards in [the dynamic reward modul](https://github.com/LiskHQ/lisk-sdk/blob/2d04e23b5b172a7b84cdad2669b985a51156e671/framework/src/modules/dynamic_rewards/endpoint.ts#L42)e, gives expected reward for a validator for several cases (day, month, year)
- getExpectedSharedRewards([ in the pos module](https://github.com/LiskHQ/lisk-sdk/blob/2d04e23b5b172a7b84cdad2669b985a51156e671/framework/src/modules/pos/endpoint.ts#L361) gives the expected shared reward if staking a validator with a particular amount.

To achieve the estimations:

- call getExpectedValidatorRewards(validator1.address),  and expected daily/monthly/yearly reward
- then call pos.getExpectedSharedRewards(validator1.address, daily/monthly/yearly, myStake = 500) and this gives my expected shared reward

- Also enable users to sort validators by `Validator weight, Commission, Rank`

ManuGowda:
### What was the problem?

This PR resolves #5295 , #5273

### How was it solved?

- Fetch reward estimations based on validator, reward and stake amount
- Implemented argo2 changes suggested by research team

### How was it tested?

- Stake a validator having different commission percentage to see the reward estimations
- Add unit tests


src/modules/pos/validator/components/EditStake/index.js
```diff
@@ -60,6 +69,7 @@ const EditStake = ({ history, stakeEdited, network, staking }) => {
 
   const [maxAmount, setMaxAmount] = useState(0);
   const [isForm, setIsForm] = useState(true);
+  const [selectedDuration, setSelectedDuration] = useState('MONTHLY');
```
oskarleonard: `const [selectedDuration, setSelectedDuration] = useState(REWARD_DURATIONS.monthly);`

src/modules/pos/validator/components/EditStake/index.js
```diff
@@ -97,6 +107,26 @@ const EditStake = ({ history, stakeEdited, network, staking }) => {
...
+  const isMonthly = selectedDuration === REWARD_DURATIONS.monthly;
+  const queryConfig = {
+    options: {
+      enabled: !!validator.address && !stakeAmount.error && Number(stakeAmount.value) > 0,
+    },
+    config: {
+      params: {
+        validatorAddress: validator.address,
+        stake: convertToBaseDenom(stakeAmount.value, token),
+        validatorReward: '0',
```
oskarleonard, line 121: should this be hardcoded to 0?

src/modules/pos/validator/components/EditStake/index.js
```diff
@@ -97,6 +107,26 @@ const EditStake = ({ history, stakeEdited, network, staking }) => {
...
+      },
+    },
+  };
+
+  const debouncedQueryConfig = useDebounce(queryConfig, 500);
```
oskarleonard, line 126: this looks strange 🤔  Why this debounce?

src/modules/pos/validator/components/EditStake/index.js
```diff
@@ -201,6 +239,25 @@ const EditStake = ({ history, stakeEdited, network, staking }) => {
...
+                    changeCategory={handleSelectDuration}
+                    categories={[
+                      { value: REWARD_DURATIONS.monthly, label: t('Monthly') },
+                      { value: REWARD_DURATIONS.yearly, label: t('Yearly') },
+                    ]}
+                  />
+                  <span>:</span>
+                </div>
+                {!isFetchingRewards ? (
+                  <TokenAmount val={expectedReward?.data.reward || 0} token={token} />
```
oskarleonard, line 254: Shall we not default to `-` here instead of `0`

src/modules/pos/validator/components/EditStake/index.js
```diff
@@ -142,14 +172,18 @@ const EditStake = ({ history, stakeEdited, network, staking }) => {
...
-    edit: t('After changing your stake amount, it will be added to the staking queue.'),
+    edit: t('Edit your stake by modifying stake amount or removing existing stake.'),
     add: titles.description,
   };
   const headerTitles = {
     edit: t('Edit Stake'),
     add: titles.title,
   };
 
+  const handleSelectDuration = ({ target: { value } }) => {
```
oskarleonard, line 183: This function and the  setSelectedDuration is never used

src/modules/pos/validator/components/EditStake/index.js
```diff
@@ -17,22 +17,26 @@ import WalletVisual from '@wallet/components/walletVisual';
 import routes from 'src/routes/routes';
 import TokenAmount from '@token/fungible/components/tokenAmount';
 import WarnPunishedValidator from '@pos/validator/components/WarnPunishedValidator';
+import { usePosExpectedSharedRewards } from '@pos/reward/hooks/queries/useStakingRewards';
 import { useLatestBlock } from '@block/hooks/queries/useLatestBlock';
 import { useAuth } from '@auth/hooks/queries';
 import { PrimaryButton, SecondaryButton, WarningButton } from 'src/theme/buttons';
+import RewardDurationSwitch from '@wallet/components/RegisterMultisigForm/CategorySwitch';
```
oskarleonard: File is called CategorySwitch while we import this as RewardDurationSwitch, unnecessary confusion.

src/modules/wallet/components/RegisterMultisigForm/CategorySwitch.js
```diff
@@ -1,39 +1,28 @@
...
-      <label
-        htmlFor={`optional-${index}`}
-        className={`optional select-optional ${styles.memberCategory} mandatory-toggle`}
-      >
-        {t('Optional')}
-      </label>
-    </div>
-  );
-};
+const CategorySwitch = ({ changeCategory, categories = [], value, index = 0 }) => (
```
oskarleonard: Always nice to prefix callbacks with `on`

src/modules/wallet/components/RegisterMultisigForm/CategorySwitch.js
```diff
@@ -1,39 +1,28 @@
...
+  <div className={styles.categoryWrapper} onChange={changeCategory}>
```
oskarleonard: div element doesnt have an `onChange` attribute.

oskarleonard: Is also warned about in chrome devtools 

image

src/modules/pos/validator/components/EditStake/index.js
ManuGowda: this is a bug on service side.. we will remove this once service bug is fixed..

src/modules/pos/validator/components/EditStake/index.js
ManuGowda: because the users can enter stake amount and the API will be triggered for every input changes.. which is unnecessary API calls for estimations?

src/modules/wallet/components/RegisterMultisigForm/CategorySwitch.js
oskarleonard: Seems to work though, but we can probably move that to the <input> element instead to get rid of the error

src/modules/pos/validator/components/EditStake/index.js
oskarleonard: true, looks good when looking at it more closely 👍

src/modules/pos/validator/components/EditStake/index.js
ManuGowda: This will be a major change since this component is reused everywhere? maybe it should be handled separately after discussing the approach?

src/modules/pos/validator/components/EditStake/index.js
ManuGowda: I think eniola wanted this be an alias so that its easy to understand what that switch is about, but not a deal breaker we can rename if there is strong case

oskarleonard: Yes, this is very confusing. I remember we talked about this long time ago that we should only allow named export (not export default, to prevent this behaviour.

oskarleonard (dismissed):

eniolam1000752 (dismissed):

eniolam1000752 (approved):

oskarleonard (approved):

PR #5316, opened 2023-09-14
Make transaction fees editable

### LINKED ISSUE #5241: Transaction priority is not always editable, although it looks like it
### Description

When I create a transaction with Lisk Desktop, it shows the "priority" as a blue link, indicating one can click on it.

But actually, it has no effect if you click on it.

Apparently, this is only possible if the network traffic is high, although there is no indication for the user when exactly this is the case.

My suggestion is to always allow editing the priority of the transaction, because it will remove the confusion around it, and maybe replace it with a visual indicator about the current network traffic, so users can decide themselves if and when they want to change the prio of the transaction.

### Additional Information
image

eniolam1000752:
### What was the problem?

This PR resolves #5241

### How was it solved?

- [x] Make transaction fees customization
- [x] Fix unit tests

### How was it tested?

-


src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
```diff
@@ -158,22 +158,33 @@ describe('ChangeCommissionForm', () => {
       fireEvent.change(input, { target: { value } });
```
oskarleonard, line 158: NRTTL:
I cannot edit to make this 0.0024
image

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
```diff
@@ -158,22 +158,33 @@ describe('ChangeCommissionForm', () => {
...
       fireEvent.click(button);
```
oskarleonard, line 160: When choosing custom fee and adding a message, the fee input gets a life of its own, not listening to my custom fee inputs.

https://github.com/LiskHQ/lisk-desktop/assets/8784876/54a39515-75bf-4621-912e-3ecfd5000b04

src/modules/transaction/components/TransactionPriority/FeeViewer.js
```diff
@@ -45,47 +47,69 @@ const FeesViewer = ({
...
+
+  const onInputKeyUp = (e, label) => {
+    if (e.keyCode !== keyCodes.enter) return;
+    setShowEditInput((state) => ({ ...state, [label]: false }));
   };
 
   const onInputBlur = (e) => {
     e.preventDefault();
-    setShowEditIcon(true);
+    setShowEditInput({}, 100);
```
oskarleonard: This Looks invalid, what does the number 100 mean here?

src/modules/pos/validator/components/ChangeCommission/Form/ChangeCommissionForm.test.js
```diff
@@ -158,22 +158,33 @@ describe('ChangeCommissionForm', () => {
...
 
+      const feeToken = { availableBalance: 40000000, chainID: '04000000', symbol: 'LSK' };
+
```
oskarleonard, line 162: NRTTL:
When using custom fee it can be displayed different in first step from the second step, feels like the user could feel tricked by this.  

https://github.com/LiskHQ/lisk-desktop/assets/8784876/dafc4395-eb70-47b2-8b36-3c185202773d

src/modules/transaction/components/TransactionPriority/TransactionPriority.js
```diff
@@ -25,18 +22,18 @@ const TransactionPriority = ({
...
   customFee,
   minRequiredBalance,
+  formProps,
+  computedMinimumFees,
 }) => {
   const [showEditIcon, setShowEditIcon] = useState(false);
-  const [inputValue, setInputValue] = useState();
-
-  const maxFee = MODULE_COMMANDS_MAP[moduleCommand].maxFee;
+  const [inputValue, setInputValue] = useState({});
```
oskarleonard: Why do inputValue need to be an object. With such a name i think of it as a string 🤔

oskarleonard (dismissed):

ManuGowda (approved):

PR #5318, opened 2023-09-14
E2E pipeline, common hooks, test for request token

### LINKED ISSUE #5080: E2E pipeline, common hooks, test for request token
### Description

Please describe the specification of new features/improvements

### Acceptance Criteria

1. Investigate why E2E pipeline are failing. 
   1. Setup screenshoots for failing tests 
3. Setup a before hook which adds a network that is used by all test features
4. Add a single steps for adding an account, (most test will need it in the background) 
5. Request tokens to an account
   - Check for successful copy link validity
   - Check for successful QR code scanning

oskarleonard:
### What was the problem?

This PR resolves #5080

### How was it solved?

1. Use hooks pattern instead of `setWorldConstructor`
   1. Add screenshoots in an after hook
   2. Close: browser, context and page in corresponding hooks
   3. Add the network to be used for 99% of the E2E tests
2. Added a common step to easily set up an account (should be used in most features background)
3. Added a feature test for Request Token
4. Increased default test timeout to 10 seconds, same that we use for our API network request.

### How was it tested?
Please be aware of that the test can sometimes fail if our E2E service network responds slowly. In that case, just rerun the test before starting to investigate why tests are failing.

#### Pass scenario
1. Run the branch with `yarn dev`
2. Run the E2E test in another terminal window/tab 
`PW_BASE_URL=http://localhost:8080/# yarn run cucumber:playwright:open`
3. **Expected**: Tests should pass

#### Fail scenario
1. Go to file `RequestToken.feature`
2. Change `And "lisk_mainchain"...` to `And "none_existing_text"...`
3. Run tests: `PW_BASE_URL=http://localhost:8080/# yarn run cucumber:playwright:open`
4. **Expected**: Tests should fail and produce this folder `e2e/assets/screenshots` where you can see a screenshoot of the failed test.


e2e/features/requestToken/RequestToken.feature
```diff
@@ -0,0 +1,16 @@
...
+    Given I click on a button with text "Request"
+    Then I should see "Request tokens"
+    And "lisk_mainchain" should be selected in "Recipient application" dropdown
+    And "Lisk" should be selected in "Token" dropdown
+    And button with text 'Copy link' should be disabled
+    Given I type "10" in "amount"
+    Then button with text "Copy link" should be enabled
+    And I click on a button with text "Copy link"
+    Then Clipboard should contain "lisk://wallet?modal=send&recipient=lskm9syv4wrjcjczpegz65zqxhk2cp9dkejs5wbjb&amount=10&token=0400000000000000&recipientChain=04000000"
+
```
ManuGowda, line 24: Missing scenarios?

`Request token "Copy link" disabled state` when required fields are not provided?

e2e/hooks/hooks.mjs
```diff
@@ -0,0 +1,62 @@
...
+Before(async function () {
+  context = await browser.newContext({
+    ignoreHTTPSErrors: true,
+    permissions: ['clipboard-read', 'clipboard-write'],
+  });
+  const newPage = await context.newPage();
+  await newPage.goto(`${process.env.PW_BASE_URL}`);
+
+  const networkName = 'devnet';
+  const serviceUrl = 'http://devnet-service.liskdev.net:9901';
```
ManuGowda: This approach can be specific and not generic? if we were to change the nature of network and URLs then they would not work? maybe define such values as config file and expose them from environment so that when required we can modify at one place?

ikem-legend (dismissed):

e2e/features/requestToken/RequestToken.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Request Token
+  Background: Add an account and navigate to wallet
+    Given I add an account with passphrase "peanut hundred pen hawk invite exclude brain chunk gadget wait wrong ready" password "Password1$" name 'test_acc'
+    Then I go to page "wallet"
```
eniolam1000752: this should be 

```suggestion
    And I go to page "wallet"
```

since its an action not an assertion

e2e/features/requestToken/RequestToken.feature
```diff
@@ -0,0 +1,16 @@
...
+
+  Scenario: Request token should generate a copy link
+    Given I click on a button with text "Request"
+    Then I should see "Request tokens"
+    And "lisk_mainchain" should be selected in "Recipient application" dropdown
+    And "Lisk" should be selected in "Token" dropdown
+    And button with text 'Copy link' should be disabled
+    Given I type "10" in "amount"
+    Then button with text "Copy link" should be enabled
+    And I click on a button with text "Copy link"
```
eniolam1000752: This should be `Given` not `And` since it comes after an assertive phrase 
```suggestion
    Given I click on a button with text "Copy link"
```

e2e/features/requestToken/RequestToken.feature
eniolam1000752: exactly. Also consider situations like

- Failure to load supported tokens
- empty amount field
- When a message is provided
- Also assert that the QR code gets blurred when the form fields are invalid

e2e/fixtures/page.mjs
```diff
@@ -0,0 +1,3 @@
+export const fixture = {
+  page: undefined,
```
eniolam1000752:
```suggestion
  page: null,
```

e2e/steps/RequestToken.mjs
```diff
@@ -0,0 +1,17 @@
...
+Then(
+  '{string} should be selected in {string} dropdown',
+  async function (selectedText, dropdownTitle) {
+    const title = fixture.page
+      .getByText(dropdownTitle, { exact: true })
+      .locator('..')
+      .getByTestId('selected-menu-item')
+      .getByText(selectedText, { exact: true });
+    const content = await title.textContent();
+    expect(content).toBeTruthy();
```
eniolam1000752: I think this should be in common since it might be used by some other test in the feature

eniolam1000752 (changes requested):

e2e/features/requestToken/RequestToken.feature
oskarleonard: Request token disabled states are in background when opening the modal, this is also the default state when opening the modal @eniolam1000752.

But i have now also added tests for:
- A check that QR image is disabled
- When amount is specifically 0
- When a message is provided

As for this "Failure to load supported tokens". If we are going to add tests for these types of states it would require us to spend unreasonable amount of time/resources for testings. It would require that we mock fail network requests in our scenarios, or am i missing something @eniolam1000752.

e2e/steps/RequestToken.mjs
oskarleonard: Fixed

e2e/hooks/hooks.mjs
oskarleonard: Hmm, somehow i cannot import our global network file to our e2e folder. Tried a few different imports:
```
import networks, { networkKeys } from 'src/modules/network/configuration/networks';
import networks, { networkKeys } from '../../src/modules/network/configuration/networks';
import networks, { networkKeys } from '@network/configuration/networks';
```
That way we could use our global network file. I guess another way would be to add it to process.env when we run the project, but from how our current networks is configured it doesn't feel like we want to do that?

ManuGowda: Ok, strange that you can not import the keys 🤔 but converting them to environment variables can deviate from what we have. 

maybe then for now you can create a global constants that can be imported this e2e which could be reused elsewhere?

e2e/features/requestToken/RequestToken.feature
eniolam1000752: hmmmm that would be true and it would mean we have to like start setting up mocks which I don't think we should explore now

eniolam1000752 (dismissed):

e2e/features/requestToken/RequestToken.feature
```diff
@@ -1,16 +1,32 @@
...
+    And I wait for "2 seconds"
+    And I click on a button with text "Request"
     Then I should see "Request tokens"
     And "lisk_mainchain" should be selected in "Recipient application" dropdown
     And "Lisk" should be selected in "Token" dropdown
     And button with text 'Copy link' should be disabled
+    And Element 'qrContainer' should contain class 'disabled'
+
+  Scenario: Request token should not generate a copy link if amount is 0
+    Given I type "0" in "amount"
```
ManuGowda, line 14: 👁️ we had a feature to allow users to send `0` amount with some message. 

The use case was for exchanges or something to do a check transaction or just to use data field for communication in network.

Good that we found the issue with e2e test 😉

eniolam1000752:
the CI is red

e2e/features/requestToken/RequestToken.feature
oskarleonard: Yea, we seem to have a check for this specifically 
image

I will change the test to, should be able to create a copy link with 0 😆

e2e/hooks/hooks.mjs
oskarleonard: I think it is because of the .mjs extensions :/. Tried to fix that quick but think it requires more research, dont know if we have time for that now.

e2e/features/requestToken/RequestToken.feature
oskarleonard: I created this issue for this. https://github.com/LiskHQ/lisk-desktop/issues/5326

ikem-legend: But while it makes sense to send 0 amount with message, I don't think it makes sense to request 0 amount. Users who request usually need tokens to carry out transactions so they have less reason to request 0 amount. If a new user were to request 0 amount and receive it, they'd still be unable to send 0 amount in a transaction because they wouldn't have sufficient transaction fees

oskarleonard: @ikem-legend but it is for the scenario Manu wrote about earlier, exchange use case.

It should be allowed to generate a request transaction with 0 amount but just a message. This request link can someone click to to just send me a message

ikem-legend: I get the scenario but I'm just not sure it requires changes on the request end, only because of how less likely it might be. The exchanges or similar accounts who need to do such transactions can do them straight without needing to request. The point of the request token feature, in my opinion, is for users to actually get tokens for transactions so requesting 0 tokens just to make it faster for another user to send them a message seems less likely.

ManuGowda (approved):

e2e/features/requestToken/RequestToken.feature
oskarleonard: I can see this point as well, which is why i didnt want to make any changes regarding that in this PR. Right now it works as you describe

ikem-legend (approved):

Issue #5323, opened 2023-09-15
Wrong delegate under staking button


przemerr:
### Actual behavior

![wrong_delegate](https://github.com/LiskHQ/lisk-desktop/assets/25085095/127ab37f-fe8d-4c90-9d2b-0b33c236f020)

Desktop 3.0.0-rc.0

ManuGowda:
![Screenshot 2023-09-18 at 3 23 27 PM](https://github.com/LiskHQ/lisk-desktop/assets/6449871/6b60e15a-1bec-4c35-bd7c-9357d4f42f2f)


@przemerr I am unable to reproduce this issue, can you share me steps?

ManuGowda:
![Screenshot 2023-09-18 at 3 24 21 PM](https://github.com/LiskHQ/lisk-desktop/assets/6449871/ec709437-21c0-47b5-aa12-0935eb044d3c)

Reference from betanet as well.

ManuGowda:
![Screenshot 2023-09-18 at 3 31 04 PM](https://github.com/LiskHQ/lisk-desktop/assets/6449871/b121ca28-e9f8-45a9-ad56-592862084a59)

przemerr:
Hi @ManuGowda ,
It only applies to betanet,
you need to have a delegate of your own
you need to explicitly go to Validators Tab, and My Validator profile (don't search for a delegate) and stake from there
![image](https://github.com/LiskHQ/lisk-desktop/assets/25085095/f6774bb3-809e-4cc8-87f7-1cd90ff5e19f)

przemerr:
@ManuGowda can you reproduce it?

ManuGowda:
> @ManuGowda can you reproduce it?

Yes was able to reproduce 👍🏼

Issue #5322, opened 2023-09-15
Stakers don't change while browsing


przemerr:
### Actual behavior

![stakers_dont_change](https://github.com/LiskHQ/lisk-desktop/assets/25085095/5145d6cd-1b74-4f8a-8982-0ce812905f6d)

Desktop 3.0.0-rc.0

Issue #5320, opened 2023-09-15
Wrong account name


przemerr:
### Actual behavior

![account_name](https://github.com/LiskHQ/lisk-desktop/assets/25085095/f71556cd-cadb-42f2-9b53-c3bf0ab56bc3)

Desktop 3.0-rc.0

PR #5324, opened 2023-09-15
Bugfix pom and validator domain

### LINKED ISSUE #5319: Sign multisignature issues
### Expected behavior

1. `Please paste the JSON here` should allow users to past the JSON
2. The border issue should be resolved
 
### Actual behavior

1. `Please paste the JSON here` doesn't allow you to paste the JSON, insted it opens up the read files.
2. Screenshot 2023-09-15 at 9 50 34 AM 

### Steps to reproduce
Design [link](https://www.figma.com/file/KcrDpvWEKQhdGwNd4CZ5NY/Desktop-Prototype?type=design&node-id=71%3A75707&mode=design&t=GwJdL2kQmBaW3dTD-1)

1. Select `Sign multisignature`
2. Copy a valid JSON data
3. Try to paste, and observe the read file dialog opens

### LINKED ISSUE #5321: Banned delegate profile page doesn't work
### Actual behavior

banned_delegates

Desktop 3.0.0-rc.0

ManuGowda:
### What was the problem?

This PR resolves #5319, #5321, #5320, #5322 and #5323

### How was it solved?

- Handle the pom edge case with null check
- Ensure staking validator is disabled when a validator is banned
- Updated outdated blog post links

#### Additional improvements

- Restrict window actions (copy, paste, cut, etc..) to prevent hijacking password

### How was it tested?

- For all issues a screen recoding has been attached, so we can refer them for validation
- The commits are also atomic so please review them based on commit fix


src/modules/common/components/uploadJSONInput/uploadJSONInput.js
```diff
@@ -42,13 +42,11 @@ const UploadJSONInput = ({
     accept: {
       'application/JSON': ['.json'],
     },
+    noClick: true,
```
oskarleonard, line 45: Why did we remove this feature? I cannot see it in any issue nor description 🤔 

Clicks are more useful than drag and drop imo, especially since users see what files are accepted easily
image

oskarleonard: Ah, never mind, i see we still have 
`Restore from JSON file` which users can click

oskarleonard: **But feels like the text of these should be more similar.**

**Add account also allows for paste now (so we should probably remove the `placeholderText` here?)**
And maybe change the title from Restore from JSON file to Select JSON file 🤔 
image

___________________

**The header above the drag and drop here feels strange.**
1 header with 2 actions. Maybe remove the `prefixLabel` and make it just say "Select JSON file with transaction value"
image

src/modules/common/hooks/useFilter.js
```diff
@@ -42,6 +42,10 @@ export default function useFilter(defaultFilters = {}) {
     [filters]
   );
 
+  useEffect(() => {
+    setFilters(inputFilters);
```
oskarleonard, line 46: **Syncing** props and state is usually an anti-pattern. 
It is OK though to **initialise** the state but not trying to keep them the sync since it leads to multiple sources of truth. Which probably was why the props name initially was `defaultFilters` (`initialFilters` would be fine too).

We need to test this thoroughly.

oskarleonard: Dont know if it is related to the ani pattern but going to http://0.0.0.0:8080/#/wallet now crashes the app 
image

ManuGowda: Yes I dint like this fix, probably we can discuss a better fix as we know the problem

oskarleonard: What i did was playing around with the filters on different pages then going to /wallet. Cannot replicate this though, so it might be working. But please also make such sanity check 🙃

src/modules/common/components/uploadJSONInput/uploadJSONInput.test.js
```diff
@@ -82,7 +82,7 @@ describe('Upload JSON input component', () => {
 
     render(<UploadJSONInput {...props} />);
 
-    const inputNode = screen.getByLabelText('Restore from JSON file');
+    const inputNode = screen.getByLabelText('Select transaction JSON file');
```
ManuGowda:
```suggestion
    const inputNode = screen.getByLabelText(props.label);
```

src/modules/common/components/uploadJSONInput/uploadJSONInput.test.js
```diff
@@ -98,7 +98,7 @@ describe('Upload JSON input component', () => {
 
     const invalidJson = '{"result":true, "count":42';
     const file = new File([invalidJson], 'file.json', { type: 'application/json' });
-    const inputNode = screen.getByLabelText('Restore from JSON file');
+    const inputNode = screen.getByLabelText('Select transaction JSON file');
```
ManuGowda:
```suggestion
    const inputNode = screen.getByLabelText(props.label);
```

src/modules/auth/hooks/useRestrictWindowActions.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const useRestrictWindowActions = ({ actions, window }) => {
+  const restrictAction = useCallback((event) => {
+    event.preventDefault();
+  }, []);
+
+  useEffect(() => {
+    actions?.forEach((action) => action && window.addEventListener(action, restrictAction));
+    return () => {
+      actions?.forEach((action) => action && window?.removeEventListener(action, restrictAction));
```
ikem-legend: I feel like `window` would still exist when the hook is unmounted, hence no need for the optional chaining, but other than that the hook works as expected

oskarleonard (dismissed):

src/modules/auth/components/EnterPasswordForm/EnterPasswordForm.js
```diff
@@ -9,6 +9,7 @@ import Box from 'src/theme/box';
 import BoxContent from 'src/theme/box/content';
 import { PrimaryButton, TertiaryButton } from 'src/theme/buttons';
 import styles from './EnterPasswordForm.css';
+import { useRestrictWindowActions } from '../../hooks/useRestrictWindowActions';
```
oskarleonard: alias

src/modules/auth/hooks/useRestrictWindowActions.js
```diff
@@ -0,0 +1,14 @@
...
+    event.preventDefault();
+  }, []);
+
+  useEffect(() => {
+    actions?.forEach((action) => action && window.addEventListener(action, restrictAction));
+    return () => {
+      actions?.forEach((action) => action && window.removeEventListener(action, restrictAction));
+    };
+  }, [actions, window, restrictAction]);
+};
```
eniolam1000752: do we need to pass in window as an argument to the hook?. Since `window` is a global variable and available any where we should be able to get it in the hook without passing it to the hook right ?

ikem-legend (dismissed):

src/modules/wallet/components/overview/overview.js
```diff
@@ -80,7 +80,7 @@ const Overview = ({ isWalletRoute, history }) => {
         isBanned,
         pomHeight: pomHeights ? pomHeights[pomHeights.length - 1] : 0,
         readMore: () => {
-          const url = 'https://lisk.com/blog/development/lisk-staking-process';
+          const url = 'https://lisk.com/blog/posts/lisk-staking-process';
```
eniolam1000752, line 83: can't we have this in a .env file  just as we do for the environment variable, PRODUCTION or DEBUG

eniolam1000752 (changes requested):

src/modules/wallet/components/overview/overview.js
ManuGowda: this was a category change.. in website they changed the category of research, development, posts, etc.. so this is just a one time thing.

oskarleonard (approved):

eniolam1000752 (approved):

ikem-legend (approved):

PR #5329, opened 2023-09-21
Fixed request 0 amount

### LINKED ISSUE #5326: Users should be able to get a request link with 0 amount
### Expected behavior
- Since i can send a transaction with 0 and only a message, i should also be able to create a request link with that parameter.
- Make sure that leaving the amount field as untouched should not be possible. 
- Update E2E test to correspond this behaviour. This scenario needs to be modified `Request token should not generate a copy link if amount is 0`

### Actual behavior
image

### Steps to reproduce
1. Open Request modal 
2. Type 0 in amount

ikem-legend:
### What was the problem?

This PR resolves #5326

### How was it solved?

Updated amount validation

### How was it tested?

- E2E tests

### DIFF
--- a/src/modules/wallet/components/request/request.js
+++ b/src/modules/wallet/components/request/request.js
@@ -89,6 +89,7 @@ const Request = () => {
   const applicationExploreAndMetaData = useApplicationExploreAndMetaData();
   const networkSupportedTokens = useNetworkSupportedTokens(state.recipientChain.value);
   const { recipientChain, token, amount, reference } = state;
+  const selectedToken = networkSupportedTokens.data?.find(({ tokenID }) => tokenID === token.value);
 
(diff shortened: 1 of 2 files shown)


src/modules/wallet/components/request/request.js
```diff
@@ -123,7 +123,9 @@ const Request = () => {
         ? validateAmount({
             amount: target.value,
             locale: i18n.language,
-            token: state.token.value,
+            // eslint-disable-next-line no-use-before-define
```
ManuGowda: this eslint rule can be removed?

ikem-legend: Fixed

ManuGowda (approved):

oskarleonard (approved):

PR #5330, opened 2023-09-21
Prevent a user from switching to a custom network which fails

### LINKED ISSUE #5309: Cannot switch back to a working network after being on another network
### Expected behavior
Switching to a working network should always work.

### Actual behavior
I need to click twice on a working network before it switches to it after being on a nonworking network. 

https://github.com/LiskHQ/lisk-desktop/assets/8784876/088fa38f-b3cb-4d4c-825f-9924623bc1a2

oskarleonard:
### What was the problem?

This PR resolves #5309

### How was it solved?

Make sure our networkClient is set before making the request. This is to prevent race conditions.

### How was it tested?

1. Switch from Betanet to Testnet
2. Expected: Show error in the dropdown
3. Do this a couple of times, switching between Betanet, Testnet and your locally added devnet
4. Expected: Should not get into a freeze state as in the issue.


src/modules/network/components/networkSwitcherDropdown/NetworkSwitcherDropdown.js
```diff
@@ -52,15 +53,21 @@ function NetworkSwitcherDropdown({ noLabel, onNetworkSwitchSuccess }) {
...
       });
-      networkStatus.refetch().then((res) => {
+      setSelectedNetwork(network);
+    },
+    [networkStatus]
+  );
+
+  useEffect(() => {
+    if (selectedNetwork.serviceUrl) {
+      networkStatus?.refetch?.()?.then((res) => {
```
eniolam1000752, line 63: There seems to be a maximum call stack exceeded error when I click on the network `DialogNetworkApplicationSelector`. It seems because this is now in a useEffect, there are states scenarios that un-wantedly triggers the setValue from setting and since the deepMerge logic is reclusive it could have been called a lot of time resulting lots of calls to deepMerge (this is an assumption though).

image

src/modules/network/components/networkSwitcherDropdown/NetworkSwitcherDropdown.js
eniolam1000752, line 1: On switching between networks, while I was on betanet, I observed I had duplicate main chain one from devenet and the other from betanet

image

eniolam1000752, line 1: There is also the issue of residual persisted network/application information lingering around the local storage even though they have been removed.

Below is a list of all the networks which I don't even see in my network list

image

image

eniolam1000752 (changes requested):

eniolam1000752 (changes requested):
- On editing a previously added custom network, the name on the selected network becomes empty.

image

ManuGowda (approved):

eniolam1000752 (approved):

PR #5334, opened 2023-09-22
Add staking e2e tests

### LINKED ISSUE #5082: Add e2e tests for staking
### Acceptance Criteria

Pre-requisite: Ensure you have an account enough balance, and the account is added to the wallet and one account has been selected and we are in `Validators` page.

1. Select few random validators
2. Add e2e tests for Stake validators
3. Add e2e tests for Unstake validators
   a. Edit stake 
   b. Remove stake
4. View stakes

ikem-legend:
### What was the problem?

This PR resolves #5082

### How was it solved?

Implementing the following tests:
- [x] Select few random validators
- [x] Add e2e tests for Stake validators
- [x] Add e2e tests for Unstake validators
- Edit stake
- Remove stake
- [x] View stakes

### How was it tested?

- Manually


e2e/steps/common.mjs
```diff
@@ -67,32 +84,32 @@ Given('I wait for {string}', async (timeout) => {
   await new Promise((res) => setTimeout(res, +time * unitMultiplier[unitKey]));
 });
 
-Then('I should see {string}', async function (textContent) {
+Then('I should see {string}', async (textContent) => {
   await expect(fixture.page.getByText(textContent, { exact: true })).toBeVisible();
 });
 
-Then('I should possibly see {string}', async function (textContent) {
+Then('I should possibly see {string}', async (textContent) => {
```
oskarleonard: I think we need to use the function here. Especially if we want to be able to scale and be able to use World and this... Most tutorials i saw recommended to use function here instead of arrow function

ikem-legend: Prettier removes it on save. It wasn't a manual change

oskarleonard: https://github.com/cucumber/cucumber-js/issues/2004

ikem-legend: While I can see that issue, there was never a PR for it and so far, we don't use the `this`/`world` context. Furthermore, any other change to that file by anyone else would still format it on save except Prettier is disabled locally

cucumber.js
```diff
@@ -5,7 +5,6 @@ module.exports = {
     paths: ['./e2e/features/**/*.feature'],
     import: ['./e2e/hooks/hooks.mjs', './e2e/steps/**/*.mjs', './e2e/steps/**/*.js'],
     dryRun: false,
-    publishQuite: true,
```
oskarleonard, line 8: I think we need this one. Think CI could pass even if it shouldnt pass if we remove it

ikem-legend: There was a typo there and `publishQuiet` is no longer a valid config option based on https://github.com/cucumber/cucumber-js/blob/main/docs/configuration.md though `--publish-quiet` might work on the command line

e2e/steps/common.mjs
oskarleonard: This is probably some local linting you have 🤔.
It doesnt format it for me, i dont think prettier nor eslint settings formats this since it basically mean two different things 🤔. If the code produced identical results it would make sense that these tools formatted it, but not when they mean different things.

Another motivation from me to keep previous syntax (except for being able to add world) is that it will cause merge conflict with my PR. I see no value in this refactor.

ikem-legend: Fixed

oskarleonard (approved):

eniolam1000752 (approved):

PR #5335, opened 2023-09-24
Bugfix: Validators can not increase commission after commission decrease

### LINKED ISSUE #5327: Validators can not increase commission after commission decrease.
### Expected behavior

Validators are not allowed to increase the commission more than once per month. That means, if they increase their commission at some point, then they can not increase again for next 28 days. 

On the other hand, if a validator has never increased their commission but has only decreased it (recall the initial commission value is 100%), then they should be able to increase it at any time by at most 5%.

### Actual behavior

Even if a validator decreases the commission, Lisk Desktop app does not allow them to increase it afterwards. 

Example:  In devnet, validator "genesis_23" decreased their commission twice, first from 100% to 15% and then from 15% to 12%, as shown in the screenshots below. 

image

image

Then, the validator should be allowed to increase their commission up to 17%. Instead, Lisk Desktop does not allow them to increase the commission as shown below. 

image

ManuGowda:
### What was the problem?

This PR resolves #5327

### How was it solved?

- Transaction dry run ensures the commission increase and other validations are handled, so we do not need to handle them specifically at form level. Check the command [verification logic](https://github.com/LiskHQ/lisk-sdk/blob/2d04e23b5b172a7b84cdad2669b985a51156e671/framework/src/modules/pos/commands/change_commission.ts#L56-L72) 
- Added more time for feedback for users to read

### How was it tested?

- Follow the steps mentioned [here](https://github.com/LiskHQ/lisk-desktop/issues/5327) and submit to view the dry run feedback

Screenshot 2023-09-24 at 11 56 11 AM


src/modules/transaction/components/TxComposer/txComposer.css
eniolam1000752, line 1: I don't think there is a need to have the feedback fade out on its own. Because this is an error feedback component fading it out means if a user looses focus out of the application coming back to the application, the user would feel like why did my transaction not go through since the feedback disappears on its own. 

I think it would be best to ensure that the feedback leave the page only when there is a change to any of the form input field. That way the error message stays on the screen so far the user doesn't make an update to any of the form input

eniolam1000752 (changes requested):

eniolam1000752 (approved):

Issue #5347, opened 2023-09-28
Same Icon for Incoming/outgoing transaction


przemerr:
### Actual behavior
I believe the icons should be different for an icnoming and outgoing transaction 
![image](https://github.com/LiskHQ/lisk-desktop/assets/25085095/889f13a3-246c-4c5a-84c5-6e7fb7fd5426)

PR #5356, opened 2023-10-02
Bugfix account balance calculation

### LINKED ISSUE #5349: Can't send transactions leaving the account with less than 0.05 LSK
### Expected behavior
With SDK version 6, the minimum balance of 0.05 LSK is no longer required. Users should now be able to send a transaction such that their balance is reduced to less than 0.05 LSK.

This account `lsk4uztac84qks4u84yxphnambadmp6tv73ojwrg3` is a proof that you can send a transaction even with balance lower than 0.05 LSK, using for example CLI

### Actual behavior
Users can't send transactions such that, after sending the transaction, the account balance is less than 0.05 LSK
image

### Steps to reproduce
Try sending a transaction from an account that result with balance lower than 0.05 LSK
Desktop 3.0.0-rc.1

ManuGowda:
### What was the problem?

This PR resolves #5349, #5347

### How was it solved?

- Fix account balance and usable balance bug by check against available balance and amount to spend including fee
- Fix transaction incoming and outgoing arrow

### How was it tested?

- Check wallet for incoming and outgoing transaction
- Send entire balance (considering fees)


eniolam1000752 (dismissed):

src/modules/transaction/api/index.js
```diff
@@ -68,24 +74,26 @@ export const broadcast = async ({ transaction, serviceUrl, moduleCommandSchemas
...
 
-const getEventDataResultError = (events) => {
+const getEventDataResultError = (events, moduleCommand) => {
   const event = events?.find((e) => e.data?.result && e.data?.result !== 0);
 
   if (event) {
-    return EVENT_DATA_RESULT[event.data.result];
+    return moduleCommand === MODULE_COMMANDS_NAME_MAP.registerValidator
+      ? VALIDATOR_EVENT_DATA_RESULT[event.data.result]
+      : TOKEN_EVENT_DATA_RESULT[event.data.result];
```
ikem-legend: There are also errors tied to other module commands like staking. So maybe a case statement might be the best solution for this case

src/modules/transaction/constants.js
```diff
@@ -36,3 +36,10 @@ export const VALIDATOR_EVENT_DATA_RESULT = {
   3: 'The BLS key has already been registered in the chain, please check and add a valid BLS key.',
   4: 'Invalid proof of possession for the given BLS key, please check and add a valid proof of possession.',
 };
+
+export const POS_EVENT_DATA_RESULT = {
+  1: 'Failed to stake for an unregistered validator.',
+  2: 'The unstaking attempt failed due to parameters that were not valid.',
+  3: 'Failed due to an excessive number of stakes being unlocked.',
+  4: 'Failed to stake due to exceeding the staking limit.',
+};
```
eniolam1000752, line 45: This constant wasn't consumed anywhere else

eniolam1000752 (changes requested):

src/modules/transaction/api/index.js
```diff
@@ -78,12 +78,19 @@ const getEventDataResultError = (events, moduleCommand) => {
...
-    return moduleCommand === MODULE_COMMANDS_NAME_MAP.registerValidator
-      ? VALIDATOR_EVENT_DATA_RESULT[event.data.result]
-      : TOKEN_EVENT_DATA_RESULT[event.data.result];
+    switch (moduleCommand) {
+      case MODULE_COMMANDS_NAME_MAP.transfer || MODULE_COMMANDS_NAME_MAP.transferCrossChain:
+        return TOKEN_EVENT_DATA_RESULT[event.data.result];
+      case MODULE_COMMANDS_NAME_MAP.registerValidator:
+        return VALIDATOR_EVENT_DATA_RESULT[event.data.result];
+      case MODULE_COMMANDS_NAME_MAP.stake || MODULE_COMMANDS_NAME_MAP.unlock:
+        return TOKEN_EVENT_DATA_RESULT[event.data.result];
```
ikem-legend: Should be `return POS_EVENT_DATA_RESULT[event.data.result]`

ikem-legend (approved):

eniolam1000752 (approved):

PR #5387, opened 2023-10-13
Dont pass transaction in URL in multi signature continuous flow

### LINKED ISSUE #5374: Dont pass transaction in URL in multi signature continuous flow
### Description

When switching account in a continuous multi sig flow we currently pass ([This PR](https://github.com/LiskHQ/lisk-desktop/pull/5366)) the transaction url as a url params.

It would be better if we use:
1. Url state (Dont think it will work since we link to a modal)
2. Redux for this temporary state

### Motivation

- Current way creates a messy URL and could probably cause problems when a lot of multi signature accounts since url params have a limit of 2048 character.

oskarleonard:
### What was the problem?

This PR resolves #5374 #5388

### How was it solved?

- Moving transaction string to url state since its max size is 640k characters compared to url params max size of 2k characters.
- Since URL is no longer unique between certain switches i also had to add a `generateUniqueId` in the url.
- Check for required remaining members to determine if a transaction is complete

### How was it tested?
Because of issue #5388 we unfortunately need to test all edge cases again to prevent regression in another state.
So please test the cases from [this PR](https://github.com/LiskHQ/lisk-desktop/pull/5366)
As well as the issue raised in this issue  #5388

### DIFF
--- a/src/modules/account/hooks/useCurrentAccount.js
+++ b/src/modules/account/hooks/useCurrentAccount.js
@@ -22,6 +22,8 @@ export function useCurrentAccount() {
   // eslint-disable-next-line max-statements
   const setAccount = (encryptedAccount, referrer, redirect = true, urlState) => {
     // clear stakes list during login or accounts switch
+    const relativeUrlPath = referrer || routes.wallet.path;
+
     if (pendingStakes.length) {
       const onCancel = /* istanbul ignore next */ () =>
         removeSearchParamsFromUrl(history, ['modal']);
@@ -29,7 +31,7 @@ export function useCurrentAccount() {
         dispatch(setCurrentAccount(encryptedAccount));
 
         dispatch(stakesReset());
-        history.push(referrer || routes.wallet.path);
+        history.push(relativeUrlPath);
       };
       const state = createConfirmSwitchState({
         mode: 'pendingStakes',
@@ -42,7 +44,6 @@ export function useCurrentAccount() {
       dispatch(setCurrentAccount(encryptedAccount));
       if (redirect) {
         if (urlState) {
-          const relativeUrlPath = referrer || routes.wallet.path;
           const { pathname, search } = new URL(relativeUrlPath, window.location.origin);
           history.push({
             pathname,
@@ -50,7 +51,7 @@ export function useCurrentAccount() {
             state: urlState,
           });
         } else {
-          history.push(referrer || routes.wallet.path);
+          history.push(relativeUrlPath);
         }
       }
     }
--- a/src/modules/account/hooks/useCurrentAccount.js
+++ b/src/modules/account/hooks/useCurrentAccount.js
@@ -19,7 +19,8 @@ export function useCurrentAccount() {
     (stake) => stake.confirmed !== stake.unconfirmed
   );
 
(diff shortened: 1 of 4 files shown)


src/modules/account/hooks/useCurrentAccount.js
```diff
@@ -40,7 +41,17 @@ export function useCurrentAccount() {
     } else {
       dispatch(setCurrentAccount(encryptedAccount));
       if (redirect) {
-        history.push(referrer || routes.wallet.path);
+        if (urlState) {
+          const relativeUrlPath = referrer || routes.wallet.path;
+          const { pathname, search } = new URL(relativeUrlPath, window.location.origin);
```
ikem-legend: Since this same variable is used in line 53, it can be defined before the conditions then used in each condition. The current implementation is also fine

ikem-legend (dismissed):

src/modules/transaction/components/Multisignature/Multisignature.js
```diff
@@ -28,15 +29,18 @@ export const PartiallySignedActions = ({
...
-  const [, setCurrentAccount] = useCurrentAccount();
+  const [currentAccount, setCurrentAccount] = useCurrentAccount();
 
   const handleSwitchAccount = () => {
     const stringifiedTransaction = encodeURIComponent(JSON.stringify(transactionJSON));
+    const uniqueUrlIdToPreventHashError = generateUniqueId();
     setCurrentAccount(
       nextAccountToSign,
-      `/wallet?modal=signMultiSignTransaction&stringifiedTransaction=${stringifiedTransaction}`,
+      `/wallet?modal=signMultiSignTransaction&prevAccount=${currentAccount.metadata.address}&uniqueId=${uniqueUrlIdToPreventHashError}`,
```
eniolam1000752, line 39: I don't see where uniqueId and prevAccount is consued

eniolam1000752 (dismissed):

src/modules/transaction/components/Multisignature/Multisignature.js
oskarleonard: They are used for the URL, not inside the app, it was added to make the url unique because otherwise there were some strange bugs, which is why i named it `uniqueUrlIdToPreventHashError`. Got a warning as well in chrome console in one of the cases when i didnt add them (dont remember which one of all cases i tested). I think it was, if you use react router hash and push two identical urls to the history stack it have some unintended consequences regarding the url state... 

Probably `prevAccount` is not needed now since url should be unique with the help of `generateUniqueId()` but i didnt want to remove it and have to test all cases again for that change. Especially since we might want to use it in the future (in another issue) when we will allow for switching back to the starting account

src/modules/account/hooks/useCurrentAccount.js
oskarleonard: Make sense, fixed in latest commit

src/modules/transaction/components/Multisignature/Multisignature.js
eniolam1000752: okay cool

eniolam1000752 (approved):

ikem-legend (approved):

PR #5390, opened 2023-10-13
Implement creating future transactions

### LINKED ISSUE #5317: Enable users to create future transactions
### Description

Multisignature accounts can not create future transactions from an account.

### Acceptance criteria

- Enable users to initiate multiple ordered transactions from an account
- Inform users about the future (nonce) transactions being created by taking network status into account
- Ensure the downloaded file has ordered nonce specified
- Add unit/e2e tests for future transactions (only applicable to multisignature accounts)

ikem-legend:
### What was the problem?

This PR resolves #5317

### How was it solved?

- [x] Create account nonce sync hook
- [x] Implement future transaction warning

### How was it tested?

- Creating multiple transactions and the nonce should increase for subsequent transactions provided the previous ones haven't been signed. Nonce warning should be displayed when transaction nonce is greater than on-chain nonce
- Unit tests


src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,55 @@
...
+  const incrementNonce = useCallback(() => {
+    let localNonce = currentAccountNonce;
+    localNonce += 1;
+    setNonceByAccount(currentAccountAddress, localNonce);
+  }, []);
+
+  return { accountNonce, onChainNonce, incrementNonce };
+};
+
+export default useNonceSync;
```
ManuGowda:
```suggestion
import { useEffect, useState, useCallback } from 'react';
import { useCurrentAccount, useAccounts } from '@account/hooks';
import { useAuth } from './queries';

const useNonceSync = () => {
  const [{ metadata: { address } }] = useCurrentAccount();
  const { setNonceByAccount, getNonceByAccount } = useAccounts();
  const currentAccountNonce = getNonceByAccount(address);

  const { data: authData } = useAuth({
    config: { params: { address } },
  });
  const onChainNonce = authData?.data.nonce;
  const [accountNonce, setAccountNonce] = useState(onChainNonce);

  // Store nonce by address in accounts store
  const handleLocalNonce = (currentNonce) => {
    let localNonce = parseInt(currentAccountNonce || 0, 10);
    if (localNonce < currentNonce) {
      localNonce = currentNonce;
    }
    setNonceByAccount(address, localNonce);

    setAccountNonce(localNonce);
  };

  useEffect(() => {
    handleLocalNonce(parseInt(onChainNonce, 10));
  }, [onChainNonce]);

  // Call incrementNonce after transaction signing
  const incrementNonce = useCallback(() => {
    let localNonce = currentAccountNonce;
    localNonce += 1;
    setNonceByAccount(address, localNonce);
  }, []);

  return { accountNonce, onChainNonce, incrementNonce };
};

export default useNonceSync;
```

src/modules/transaction/components/TxSignatureCollector/TxSignatureCollector.js
```diff
@@ -104,8 +106,12 @@ const TxSignatureCollector = ({
     });
   };
 
-  const onEnterPasswordSuccess = ({ privateKey }) =>
+  const onEnterPasswordSuccess = ({ privateKey }) => {
+    if (isTransactionAuthor) {
+      incrementNonce();
```
ManuGowda: the increment should happen once user clicks `Download`? otherwise following can happen

- The user intentionally closes the window without download, hence he has no file to sign, in that case the nonce should still be old one?
- The same applies for unintentionally close scenario as well?
- Now that we have switch scenario if the user clicks `Switch account` we might increase the nonce, so best to increment the nonce when user downloads the file or copies

src/modules/transaction/components/TxSummarizer/index.js
```diff
@@ -100,6 +107,16 @@ const TxSummarizer = ({
             </div>
           </section>
         )}
+        {nonceWarning && (
+          <div className={styles.nonceWarning}>
+            <Icon name="warningYellow" />
+            <p>
+              Please ensure to send the previously initiated transactions. Check the transaction
```
ManuGowda: Translation?

src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,55 @@
...
+  };
+
+  useEffect(() => {
+    handleLocalNonce(parseInt(onChainNonce, 10));
+  }, [onChainNonce]);
+
+  // Call incrementNonce after transaction signing
+  const incrementNonce = useCallback(() => {
+    let localNonce = currentAccountNonce;
+    localNonce += 1;
```
oskarleonard: Do we need to mutate localNonce here? Why not just using const?
Something like
`let localNonce = currentAccountNonce + 1;`
or
`let localNonce = parseInt(currentAccountNonce) + 1;`

src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,55 @@
...
+  };
+  const baseUrl = mainChainNetwork?.serviceUrl;
+  const config = useAuthConfig(customConfig);
+  const authData = queryClient.getQueryData([AUTH, chainID, config, baseUrl]);
+  const onChainNonce = authData?.data.nonce;
+  const [accountNonce, setAccountNonce] = useState(onChainNonce);
+
+  // Store nonce by address in accounts store
+  const handleLocalNonce = (currentNonce) => {
+    let localNonce = parseInt(currentAccountNonce || 0, 10);
```
oskarleonard: Could also be written in a non mutable way. 
```
const accountNonce = parseInt(currentAccountNonce || 0, 10);
const localNonce = accountNonce < currentNonce ? currentNonce : accountNonce;
```
Generally much easier to read and less error prone when writing purely immutable

src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,55 @@
...
+import { useQueryClient } from '@tanstack/react-query';
+import { useCurrentApplication } from '@blockchainApplication/manage/hooks';
+import { useCurrentAccount, useAccounts } from '@account/hooks';
+import useSettings from '@settings/hooks/useSettings';
+import { AUTH } from 'src/const/queries';
+import { useAuthConfig } from './queries';
+
+// eslint-disable-next-line max-statements
+const useNonceSync = () => {
+  const queryClient = useQueryClient();
```
oskarleonard, line 11: Is there a way to group these vars with space separators, to scope connected logic. For example 

```
 const [currentAccount] = useCurrentAccount();
  const currentAccountAddress = currentAccount.metadata.address;
  const { mainChainNetwork } = useSettings('mainChainNetwork');
  const queryClient = useQueryClient();
  const chainID = currentApplication.chainID;
  const customConfig = {
    params: {
      address: currentAccountAddress,
    },
  };
  const baseUrl = mainChainNetwork?.serviceUrl;
  const config = useAuthConfig(customConfig);
  const authData = queryClient.getQueryData([AUTH, chainID, config, baseUrl]);
  const onChainNonce = authData?.data.nonce;

const [accountNonce, setAccountNonce] = useState(onChainNonce);

const currentAccountNonce = getNonceByAccount(currentAccountAddress);
  ```

File: src/modules/auth/hooks/useNonceSync.js [hunk 7]
```diff
@@ -0,0 +1,55 @@
...
+  const currentAccountAddress = currentAccount.metadata.address;
+  const currentAccountNonce = getNonceByAccount(currentAccountAddress);
+  const { mainChainNetwork } = useSettings('mainChainNetwork');
+  const chainID = currentApplication.chainID;
+  const customConfig = {
+    params: {
+      address: currentAccountAddress,
+    },
+  };
+  const baseUrl = mainChainNetwork?.serviceUrl;
```
oskarleonard: `const serviceUrl = mainChainNetwork?.serviceUrl;`

src/modules/auth/hooks/useNonceSync.js
oskarleonard: Ah, i think that is what Manu is suggesting in https://github.com/LiskHQ/lisk-desktop/pull/5390/files#r1365008159. Much easier to understand

src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,55 @@
...
+  }, [onChainNonce]);
+
+  // Call incrementNonce after transaction signing
+  const incrementNonce = useCallback(() => {
+    let localNonce = currentAccountNonce;
+    localNonce += 1;
+    setNonceByAccount(currentAccountAddress, localNonce);
+  }, []);
+
+  return { accountNonce, onChainNonce, incrementNonce };
```
oskarleonard: Since nonce is string in API, couldnt this hook expose it as a string as well, no need to mix types except for when doing calculations

src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,55 @@
...
+  const customConfig = {
+    params: {
+      address: currentAccountAddress,
+    },
+  };
+  const baseUrl = mainChainNetwork?.serviceUrl;
+  const config = useAuthConfig(customConfig);
+  const authData = queryClient.getQueryData([AUTH, chainID, config, baseUrl]);
+  const onChainNonce = authData?.data.nonce;
+  const [accountNonce, setAccountNonce] = useState(onChainNonce);
```
oskarleonard: Here we initialize it as string type? (from API) 
Later we set this var to an int.
Feels strange mixing types, cant we just always store and expose it as a string and just doing int conversions when need to calculate

src/modules/auth/hooks/useNonceSync.js
ikem-legend: The difference between this approach and my approach is that this makes an additional API call and pulls from the cache if necessary. However, the auth endpoint is already called a number of times especially during app bootstrap and after a new transaction is received. So the cache is almost always up-to-date, hence my approach of retrieving data from the cache without triggering an additional API call

src/modules/auth/hooks/useNonceSync.js
ikem-legend: Well it could be parsed as a number here but I'll assess this then change accordingly

src/modules/transaction/components/TxSummarizer/index.js
ikem-legend: Fixed

src/modules/auth/hooks/useNonceSync.js
ikem-legend: Fixed

src/modules/auth/hooks/useNonceSync.js
ikem-legend: Fixed

src/modules/auth/hooks/useNonceSync.js
ikem-legend: Fixed

src/modules/auth/hooks/useNonceSync.js
ikem-legend: Fixed. It's now calculated as a BigInt but stored as a string

src/modules/auth/hooks/useNonceSync.js
ikem-legend: FIxed

src/modules/account/hooks/useAccounts.js
```diff
@@ -22,11 +22,24 @@ export function useAccounts() {
   const getAccountByPublicKey = (pubkey) =>
     accounts.find((account) => account.metadata.pubkey === pubkey);
 
+  const setNonceByAccount = useCallback(
+    (address, nonce, transactionHex) => dispatch(setAccountNonce(address, nonce, transactionHex)),
+    []
+  );
+
+  const getNonceByAccount = (address) => {
+    const accountNonceMap = useSelector(selectAccountNonce)[address] ?? {};
```
oskarleonard: We should not call hooks inside of regular functions. They must be top layer in a component or custom hook. We might want to install [eslint](https://www.npmjs.com/package/eslint-plugin-react-hooks) for to prevent side effects that can happen from this. But i have not seen this patterned use anywhere else in our code (we are bad at passing in dependencies thought 😓)

src/modules/account/hooks/useAccounts.js
```diff
@@ -22,11 +22,24 @@ export function useAccounts() {
...
+    const nonceList = Object.values(accountNonceMap).length ? Object.values(accountNonceMap) : [0];
```
oskarleonard: We can do the Object.values call once
```
const accountNonceValues = Object.values(accountNonceMap);
const nonceList = accountNonceValues?.length ? accountNonceValues : [0];
```

src/modules/account/hooks/useAccounts.js
oskarleonard: This might be the sideeffect of doing this, or the side effect will happen somewhere else 🤔 
image

src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,54 @@
...
+  const authNonce = typeof onChainNonce === 'bigint' ? onChainNonce.toString() : onChainNonce;
+
+  const [accountNonce, setAccountNonce] = useState(onChainNonce.toString());
+  const currentAccountNonce = getNonceByAccount(currentAccountAddress);
+
+  // Store nonce by address in accounts store
+  const handleLocalNonce = (currentNonce) => {
+    const storedNonce = BigInt(currentAccountNonce || 0);
+    const localNonce = storedNonce < currentNonce ? currentNonce : storedNonce;
+    setNonceByAccount(currentAccountAddress, localNonce.toString());
```
oskarleonard: Can call toString once and use that
`const localNonceStr = localNonce.toString()`

src/modules/transaction/components/TxSignatureCollector/TxSignatureCollector.js
```diff
@@ -104,8 +107,17 @@ const TxSignatureCollector = ({
     });
   };
 
-  const onEnterPasswordSuccess = ({ privateKey }) =>
+  const onEnterPasswordSuccess = ({ privateKey }) => {
+    // const { signatures, id, ...rest } = transactionJSON;
```
oskarleonard: comment

ikem-legend: Fixed

src/modules/account/hooks/useAccounts.js
ikem-legend: Fixed

src/modules/auth/hooks/useNonceSync.js
ikem-legend: Fixed

src/modules/account/hooks/useAccounts.js
ikem-legend: Fixed

src/modules/transaction/components/TransactionInfo/TransactionInfo.js
```diff
@@ -71,7 +75,13 @@ const TransactionInfo = ({
               <label>{date || '-'}</label>
             </div>
             <div className={styles.col}>
-              <label>{t('Nonce')}</label>
+              <label>
+                {t('Nonce')}{' '}
+                {nonceWarning && <Icon name="warningYellow" className={styles.warning} />}
+                {canResetNonce && (
```
oskarleonard, line 81: Why this new var. If we have a nonceWarning we should be able to reset 🤔

src/modules/account/hooks/useAccounts.js
```diff
@@ -22,11 +23,25 @@ export function useAccounts() {
   const getAccountByPublicKey = (pubkey) =>
     accounts.find((account) => account.metadata.pubkey === pubkey);
 
+  const setNonceByAccount = useCallback(
+    (address, nonce, transactionHex) => dispatch(setAccountNonce(address, nonce, transactionHex)),
+    []
+  );
```
eniolam1000752: there would be no need for a call back here since it might impact on memory.

src/modules/transaction/components/TransactionInfo/TransactionInfo.js
ikem-legend: Nonce should should only be reset if transaction hasn't been signed yet. That's why another variable was needed

src/modules/account/hooks/useAccounts.js
eniolam1000752, line 1: The present implementation of the nonce caching is scary in the sense that there is no history clean up strategy. 

Since for every multi sig transaction a user creates, we are always keeping a record of it at a point a user would have like eg 200 transactions possibly from different accounts on that device and since there is no way a transaction done at index 0 would have a nonce greater than the transaction done at index 100.

I think there should be a sanitisation of this states so as not to fill up local storage space (remember it has a max size of 10mb) since transactions HEX can be quite very lengthy.

src/modules/account/store/reducer.js
```diff
@@ -58,14 +58,33 @@ export const list = (state = {}, { type, encryptedAccount, accountDetail, addres
...
+      if (state[address]?.[transactionHex]) {
+        return state;
+      }
+      return {
+        ...state,
+        [address]: {
+          ...state[address],
+          [transactionHex]: nonce,
+        },
+      };
```
eniolam1000752, line 73: As a follow up to my above statement this create a history of transactions created, though there is not clean up strategy.

src/modules/auth/hooks/useNonceSync.js
```diff
@@ -0,0 +1,55 @@
...
+    const localNonce = storedNonce < currentNonce ? currentNonce : storedNonce;
+    const localNonceStr = localNonce.toString();
+    setNonceByAccount(currentAccountAddress, localNonceStr, 'defaultNonce');
+
+    setAccountNonce(localNonceStr);
+  };
+
+  useEffect(() => {
+    handleLocalNonce(onChainNonce);
+  }, [onChainNonce]);
```
eniolam1000752, line 44: This can be done in a memo

eg


```suggestion
const accountNonce = useMemo(() => {
    const storedNonce = BigInt(currentAccountNonce || 0);
    const localNonce = storedNonce < currentNonce ? currentNonce : storedNonce;
    const localNonceStr = localNonce.toString();
    setNonceByAccount(currentAccountAddress, localNonceStr, 'defaultNonce');
    
    return localNonceStr
}, [currentAccountNonce, currentAccountAddress, ])
```

eniolam1000752 (changes requested):

src/modules/account/hooks/useAccounts.js
ikem-legend: The fact that a transaction might most likely be completed on a different device than it was initiated on plays a role in why this can be tricky to solve. However, the fix for clearing all stored nonce can play a role in this

eniolam1000752 (approved):

oskarleonard (approved):

PR #5395, opened 2023-10-17
Show the validator that gave claimable rewards

### LINKED ISSUE #5386: Make pos_getClaimableRewards return value for each stake
### Description

Make `pos_getClaimableRewards` return rewards amount for each staked validator.

### Motivation

With the current implementation, `pos_getClaimableRewards` returns an aggregated value of claimable rewards without considering the contribution of each stake towards the rewards. It's understandable that the rewards are guaranteed by the protocol, and users can trust that the number is correct. However, there are many instances where having a clear distribution of the rewards would be beneficial:

- when a validator starts missing many blocks, the staker receives less than they should,
- when a validator gets PoMed or banned, the staker receives no rewards,
- when a validator at some point has too low self-vote to accommodate for the vote weight, the staker receives lower rewards than usual,
- when a validator starts increasing its commission.

All of these instances will be hard to notice if a staker votes for many validators.

The change is primarily aimed at improving the wallet experience and may benefit future third-party tools. 

If you decide to implement it, please inform the Lisk Service/Desktop team so they can incorporate it into the desktop wallet.

oskarleonard:
### What was the problem?

This PR resolves #5386

### How was it solved?

Added design to let users know which validators gave staking rewrads

### How was it tested?

1. Use account 0 derivation path (it has rewards)
2. Click Claim rewards on http://0.0.0.0:8080/#/validators/profile/stakes
3. Expected: You should see which validators gave you the reward

### DIFF
--- a/src/modules/pos/reward/hooks/queries/useValidatorRewardsWithToken.js
+++ b/src/modules/pos/reward/hooks/queries/useValidatorRewardsWithToken.js
@@ -1,0 +1,59 @@
+import { useCurrentAccount } from '@account/hooks';
+import { useEffect, useState } from 'react';
+import { useCommandSchema } from '@network/hooks';
+import { useAuth } from '@auth/hooks/queries';
+import { useAppsMetaTokensConfig } from '@token/fungible/hooks/queries';
+import { addTokensMetaData } from '@token/fungible/utils/addTokensMetaData';
+import defaultClient from 'src/utils/api/client';
+import { isEmpty } from 'src/utils/helpers';
+import { splitModuleAndCommand } from '@transaction/utils';
+import { MODULE_COMMANDS_NAME_MAP } from '@transaction/configuration/moduleCommand';
+import { getRewards } from '@pos/validator/components/StakeForm/StakeForm';
+
+export const useValidatorRewardsWithToken = () => {
+  const [
+    {
+      metadata: { pubkey, address },
+    },
+  ] = useCurrentAccount();
+  const [validatorRewardsWithToken, setValidatorRewardsWithToken] = useState();
+  const { moduleCommandSchemas } = useCommandSchema();
+  const { data: auth } = useAuth({ config: { params: { address } } });
+  const authNounce = auth?.data?.nonce;
+  const createMetaConfig = useAppsMetaTokensConfig();
+  const transformToken = addTokensMetaData({ createMetaConfig, client: defaultClient });
+
+  useEffect(() => {
+    if (isEmpty(moduleCommandSchemas) || !authNounce || !address || !pubkey) {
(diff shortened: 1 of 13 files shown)


src/modules/pos/reward/hooks/queries/useValidatorRewardsWithToken.js
```diff
@@ -0,0 +1,62 @@
...
+    },
+  ] = useCurrentAccount();
+  const { moduleCommandSchemas } = useCommandSchema();
+  const { data: auth } = useAuth({ config: { params: { address } } });
+  const authNounce = auth?.data?.nonce;
+  const createMetaConfig = useAppsMetaTokensConfig();
+  const transformToken = addTokensMetaData({ createMetaConfig, client: defaultClient });
+
+  useEffect(() => {
+    if (isEmpty(moduleCommandSchemas) || !authNounce || !address || !pubkey) {
```
ManuGowda:
```suggestion
    if (isEmpty(moduleCommandSchemas) || !authNonce || !address || !pubkey) {
```

ManuGowda (approved):

src/modules/pos/reward/hooks/queries/useValidatorRewardsWithToken.test.js
```diff
@@ -0,0 +1,28 @@
...
+  moduleCommandSchemas: mockCommandParametersSchemas.data.commands.reduce(
+    (result, { moduleCommand, schema }) => ({ ...result, [moduleCommand]: schema }),
+    {}
+  ),
+});
+
+describe('useValidatorRewardsWithToken', () => {
+  it('Should toggle load when finished loading validators', async () => {
+    const { result, waitFor } = renderHook(() => useValidatorRewardsWithToken(), { wrapper });
+    await waitFor(() => result.current.isLoading === false);
```
ikem-legend, line 25: `await waitFor(() => !result.current.isLoading)` is another way this could have been implemented but this is fine

ikem-legend (approved):

src/modules/pos/reward/hooks/queries/useValidatorRewardsWithToken.test.js
oskarleonard: Since we initialise isLoading to null `const [isLoading, setIsLoading] = useState(null);` And when we start to fetch we set value to True and then when fetched is complete we set it to False. We need to explicitly check for false since !null is true so it would never wait for the complete fetch.

Sure it could probably be done if rewriting the hook a bit, but for now isLoading is
null: when hook is first initialized
true: when hook is loading
false: when hook has finished loading (not `null`)

PR #5408, opened 2023-10-27
Fix nonce undefined in certain state

oskarleonard:
### What was the problem?
This PR resolves https://github.com/LiskHQ/lisk-desktop/issues/5411, https://github.com/LiskHQ/lisk-desktop/issues/5412

### How was it solved?

Use our regular auth hook to make sure we send correct nonce. Code is also much less and easier to follow. I think a possible extra network call is worth it to always guarantee that we send with a correct nonce.

#### Prerequisite
If you have tested this feature before. Clean your redux first (since structure was wrong before)
1. Go to `src/modules/account/store/reducer.js`
2. Remove `localNonce` from whitelist
3. Go back to running application and make a page refresh. 
4. Add `localNonce` back to whitelist

#### How to test
1. Create a new multi sig account
5. Try sending from that account right after successfully creating it
6. Expected: Should work without reseting the nonce

### DIFF
--- a/src/modules/account/hooks/useAccounts.js
+++ b/src/modules/account/hooks/useAccounts.js
@@ -23,18 +23,18 @@ export function useAccounts() {
   const getAccountByPublicKey = (pubkey) =>
     accounts.find((account) => account.metadata.pubkey === pubkey);
 
-  const setNonceByAccount = (address, nonce, transactionHex) =>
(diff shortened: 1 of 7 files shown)


ikem-legend (dismissed):

ManuGowda (approved):
Code changes LGTM.

Need to test it by clearing redux or open incog window.

ikem-legend (approved):

PR #5410, opened 2023-10-29
Implemented additional value column on transaction table

### LINKED ISSUE #5348: Add additional columns to the transaction list 🙏
image

also here

image

It's not a bug it's a feature request, those are important metrics, lisk-mobile has it, no reason desktop wallet shouldn't have it 🙏

### LINKED ISSUE #5349: Can't send transactions leaving the account with less than 0.05 LSK

eniolam1000752:
### What was the problem?

This PR resolves #5348

### How was it solved?

- [x] Introduce the new value column to the transaction monitor table
- [x] Introduce the new value column and refactor columns on the wallet transaction table
- [x] Fix failing unit tests 

### How was it tested?

- Navigate to the transactions table on the transaction monitor page
- Navigate to the transactions table on the wallet page
- The new value column should be visible.

### DIFF
--- a/src/modules/transaction/components/TransactionRow/components.js
+++ b/src/modules/transaction/components/TransactionRow/components.js
@@ -1,6 +1,6 @@
 import React, { useContext } from 'react';
 import { Link } from 'react-router-dom';
-import { getTransactionAmount } from '@transaction/utils/transaction';
+import { /* getTransactionAmount, */ getTransactionValue } from '@transaction/utils/transaction';
 import { MODULE_COMMANDS_NAME_MAP } from 'src/modules/transaction/configuration/moduleCommand';
 import DateTimeFromTimestamp from 'src/modules/common/components/timestamp';
 import WalletVisual from '@wallet/components/walletVisual';
@@ -22,7 +22,7 @@ import { getValidatorDetailsClass } from '@pos/validator/components/ValidatorsTa
 import styles from './row.css';
 import TransactionRowContext from '../../context/transactionRowContext';
 import TransactionTypeFigure from '../TransactionTypeFigure';
-import TransactionAmount from '../TransactionAmount';
+// import TransactionAmount from '../TransactionAmount';
 
 export const ID = ({ isWallet }) => {
   const { data, address } = useContext(TransactionRowContext);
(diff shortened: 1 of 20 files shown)


src/modules/transaction/components/TransactionRow/components.js
```diff
@@ -1,6 +1,6 @@
 import React, { useContext } from 'react';
 import { Link } from 'react-router-dom';
-import { getTransactionAmount } from '@transaction/utils/transaction';
+import { /* getTransactionAmount, */ getTransactionValue } from '@transaction/utils/transaction';
```
ManuGowda: Clean this and below comment `// import TransactionAmount from '../TransactionAmount';`

src/modules/transaction/utils/transaction.js
```diff
@@ -91,7 +93,13 @@ const getTransactionAmount = ({ module, command, params = {} }) => {
   ) {
     return params.amount;
   }
-
+  console.log(
```
ManuGowda: Remove

src/modules/transaction/utils/transaction.js
```diff
@@ -101,6 +109,37 @@ const getTransactionAmount = ({ module, command, params = {} }) => {
   return '0';
 };
 
+export const getTransactionValue = (transactionJSON, token) => {
+  const { params, moduleCommand } = transactionJSON;
+
+  const getTokenWithSymbol = (tokenInBaseDenom) => (
+    <TokenAmount val={tokenInBaseDenom} token={token} />
+  );
+  const trasnactionValueMap = {
```
ManuGowda:
```suggestion
  );

  const trasnactionValueMap = {
```

src/modules/transaction/utils/transaction.js
```diff
@@ -101,6 +105,47 @@ const getTransactionAmount = ({ module, command, params = {} }) => {
...
+  const genrateLink = (text, href) => (
+    <Link to={href} onClick={(e) => e.stopPropagation()}>
+      {text}
+    </Link>
+  );
+
+  const trasnactionValueMap = {
+    [MODULE_COMMANDS_NAME_MAP.changeCommission]: () => `${+params.newCommission / 100}%`,
+    [MODULE_COMMANDS_NAME_MAP.claimRewards]: () => null,
+    [MODULE_COMMANDS_NAME_MAP.unlock]: () => null,
```
ManuGowda, line 124: For unlock and claim rewards we can use events tab? which has data? similar to `Signatories` logic you can provide a link for events? 

i.e instead of `-` for these trx we can use `events` and make that as a link to redirect to events tab next to `Details`?

ManuGowda: Ignore this suggestion.

ManuGowda (approved):

oskarleonard (approved):

PR #5414, opened 2023-11-01
Wallet connect multisignature continuous flow

### LINKED ISSUE #5371: Multisignature wallet connect account should initiate the continuous signing flow
### Description

We need to initiate a continuous signing flow for wallet connect accounts which are multi signature.

When signing an account from wallet connect which is a multisignature account we should be given a switch account action in this view:
image
If the mandatory account exists in users accounts

### Motivation

It should be easy to sign a multi signature transaction if all required accounts are in the users account management

### Additional Information
Relevant [PR](https://github.com/LiskHQ/lisk-desktop/pull/5366) where this was solved for non wallet connect accounts

oskarleonard:
### What was the problem?
Wait to merge this. We probably dont want this continues flow for wallet connect accounts since the wallet connect app should probably be responsible to handle this.

This PR resolves #5371


ManuGowda:
Please do not delete this branch.

PR #5417, opened 2023-11-07
Substitute empty fiat values for a dash(-)

### LINKED ISSUE #5398: Show dash symbol when we dont have the fiat value
### Description

For sidechains we cannot provide the fiat value. So for example on this view
image

We should make sure to show `-` in the fiat column

eniolam1000752:
### What was the problem?

This PR resolves #5398

### How was it solved?

- Add `-` as emptyPlaceholder on the converter component

### How was it tested?

- Navigate to the claim rewards and/or Token list page. It would be observed that token's without a fiat value would have the `-` value in place


oskarleonard (dismissed):

src/modules/pos/validator/components/ClaimRewardsForm/index.js
```diff
@@ -65,8 +65,9 @@ const ClaimRewardsForm = ({ nextStep }) => {
           </p>
           <div className={styles.tableContainer}>
             <Table
-              isLoading={!validatorRewardsWithToken}
               showHeader
+              emptyPlaceholder="-"
```
ikem-legend: The `emptyPlaceholder` prop is never used in the table component

ikem-legend (approved):

oskarleonard (approved):

PR #5421, opened 2023-11-07
Fix word overlap on edit staking modal

### LINKED ISSUE #5418: Words overlapping in Edit Stake modal
### Actual behavior
image

### Steps to reproduce
go to the edit stake modal with an account that has a lot of digits in their balance

desktop 3.0.0-rc.3, windows

eniolam1000752:
### What was the problem?

This PR resolves #5418

### How was it solved?

- Change `Available balance` to `Usable balance`
- Round balance to multiple of 10

### How was it tested?

- Navigate to the edit stake modal. Ensure a user has lots of token balance value with 8 decimal points. I should be observed that the decimal points are truncated remaining only a rounded up value with amounts to a multiple of 10.


src/modules/search/hooks/useDebounce.js
```diff
@@ -18,7 +18,7 @@ export function useDebounce(value, delay) {
     return () => {
       clearTimeout(timeoutHandler.current);
     };
-  }, [value, delay]);
+  }, [JSON.stringify(value), delay]);
```
ikem-legend, line 21: Why stringify the value?

ikem-legend (approved):

src/modules/search/hooks/useDebounce.js
oskarleonard: Yea, i dont like this either, doing this for all places using this hook while the only place this is needed is in the EditStake component?
Looks like we can handle this outside of the component?
Something like:
`const stakeValue = useDebounce(convertToBaseDenom(stakeAmount.value, token), 500);` 
instead of 
`const debouncedQueryConfig = useDebounce(queryConfig, 500);` 

Or maybe extend the useDebounce hook, so this could work
```
const debouncedQueryConfig = useDebounce(queryConfig, 500, queryConfig.config.params.stake);


useDebounce(value, time, updateString) {
   const valueUpdater = updateString || value;
...
   [valueUpdater, delay]);
}
```

oskarleonard (approved):

PR #5422, opened 2023-11-08
Prevent auto update of generator table

### LINKED ISSUE #5399: Keep sort in validator and generator table even if list updates
### Description

When I sort the validators according to their commission, sometimes (but not always) after a few seconds they get again randomly re-shuffled. Other times, the list stays same an I have time to browse validators according to their commission.

### Motivation

It is bad UX and makes it difficult to scroll the list

oskarleonard:
### What was the problem?

This PR resolves #5399

### How was it solved?

Replacing refetchInterval on generator table with ui button to enable refreshes of this table.

### How was it tested?

1. Go to http://localhost:8080/#/validators
2. Sort by commission
3. Wait 20 sec
4. Expected: List should not have been update
5. Click Refresh
6. Expected: List is now updated


src/theme/QueryTable/QueryTable.js
```diff
@@ -31,12 +31,13 @@ export const QueryTable = ({
     // When the header is fixed at the top, the position is 50px
     // Therefore the page should only scroll into the view if the header is not at the top
     if (
-      document.querySelector(`.${styles.header}`).getBoundingClientRect().top - window.scrollY <=
+      document.querySelector(`.${styles.header}`)?.getBoundingClientRect().top - window.scrollY <=
```
eniolam1000752:
```suggestion
      (document.querySelector(`.${styles.header}`)?.getBoundingClientRect()?.top || window.scrollY) - window.scrollY <=
```

eniolam1000752 (changes requested):
- I get this error when I navigate to  other tabs like`Validators` and `Sanctioned` tab.

image

- Placement seems disturbing as it distorts the view of the other tabs
image

ikem-legend (approved):

eniolam1000752 (approved):

PR #5424, opened 2023-11-09
Fix yarn run start

oskarleonard:
### What was the problem?

With the latest electron release we regressed and broke our `yarn run start` command. 
image

### How was it solved?

Downgrading a few packages

### How was it tested?

1. `yarn install`
2. `yarn build`
3. `LISK_DESKTOP_URL="http://localhost:8080" DEBUG=true yarn run start`
4. Expected: Should not cause error


shuse2:
Closing due to inactivity

Issue #5427, opened 2023-11-10
Misalignment in blocks transactions header


eniolam1000752:
### Expected behavior
- Table header columns should align with the table body columns

### Actual behavior
<img width="1480" alt="image" src="https://github.com/LiskHQ/lisk-desktop/assets/32064769/e6f63f8f-f59b-413c-8762-e8467d746938">

Issue #5426, opened 2023-11-10
Application freezes when required signatures provided is large


eniolam1000752:
### Expected behavior
- Register/Edit multi signature should only allow a maximum of 64 required signatures.

### Actual behavior
- The required signature field should only have a maximum number of 64. Though it is able to take in more than that which freezes the application due to trying to render a log of public key input field.

### Steps to reproduce
- Try to create or edit a multi-signature account
- Provide a large number for required signatures
- It would be observed that the application freezes.

Issue #5425, opened 2023-11-10
Signing a message takes forever


eniolam1000752:
### Expected behavior
A user should be able to sign a message

### Actual behavior
I am unable to sign a message because it takes too long even after 5mins I was still unable to sign a simple text

### Steps to reproduce
- Navigate to the sign message page
- Try to sign a message after providing a password

JesusTheHun:
@ManuGowda I'm still experiencing the issue with Lisk Desktop `v3.0.0-rc.3` on MacOS Somona 14.0 on Apple Silicon.
Check the video at 30:18

Issue #5419, opened 2023-11-07
Sender/Recipient logic in the transaction page


przemerr:
### Expected behavior
The transactions table should show a correct recipient/sender
### Actual behavior
![image](https://github.com/LiskHQ/lisk-desktop/assets/25085095/0aaada34-fbbe-4cdf-80f4-5b78fa4c7302)
It seems like the recipient column (`lskw7488a9nqy6m3zkg68x6ynsp6ohg4y7wazs3mw`) always shows a sender, for the best UI experience:

- in case of an outgoing transaction it should show the recipient,
- in case of an incoming transaction it should show the sender

or just show both in any case.

Additionally (I don't want to create a new github issue) it would be nice if the transaction detail page would show the recipient if there's any, currently you need to expand the parameters to see it.

![image](https://github.com/LiskHQ/lisk-desktop/assets/25085095/ce86e719-be88-4399-9fbf-696bba4ccd0a)

ManuGowda:
Thank you @przemerr for the feedback.. in the wallet page we can show the `Sender` but adding `Recipient` would make the table congested and on a small screen it would look bad.

However, for the details page we will add the `Recipient` that should be good enough?

Also note, not all transactions have `Recipient` and we wanted this table to be generic for all types of sidechain applications.

If lets say a sidechain ABC has different types of modules with different properties then the wallet wont be able to show them generically, hence balancing the fields to display.

przemerr:
Hi @ManuGowda my point for the transaction list is that, if i browse for example my account.. i want to quickly see from were the transaction is coming from or were was it sent

Lisk Desktop v2 does exactly this:

- in case of an outgoing transaction it shows the recipient,
- in case of an incoming transaction it shows show the sender

and v3 always shows the sender (and obviously I'm the sender if I'm browsing my account), so you have to pretty much always click the transaction to see were you sent it, I understand the wallet has to be compatible with different chains, but still it's a regression. If there's no recipient just leave it empty, or have `N/A` graphic If nothing can be done about it, it's fine, I'm just pointing out that Desktop v2 had the correct logic.

As for the details page adding `Recipient` is fine :), but i see it also misses the `amount/value` so if it's not a problem it would be good to have it also there.

Thanks for taking the feedback into consideration 🙏

PR #5430, opened 2023-11-10
Fix issues observed on 3.0.0-rc.3

### LINKED ISSUE #5429: Invalid column name on transaction table
### Expected behavior
The first column of the transactions table in the overview page should be `Sender` not `Recipient`

### Actual behavior
The first column of the transactions table says `Recipient`

### Steps to reproduce
- Navigate to the wallets page you would see that the first column is `Recipient`

eniolam1000752:
### What was the problem?

This PR resolves #5429, #5427, #5426, #5425 and #5419

### How was it solved?

- Added validation to multi-signature required signature input field
- Fix block transactions header alignment
- Fix first column of the wallet's page transaction table
- Fix issue with sign message

### DIFF
--- a/src/modules/transaction/components/TxSignatureCollector/TxSignatureCollector.js
+++ b/src/modules/transaction/components/TxSignatureCollector/TxSignatureCollector.js
@@ -108,12 +108,14 @@ const TxSignatureCollector = ({
   };
 
   const onEnterPasswordSuccess = ({ privateKey }) => {
-    const paramsSchema = moduleCommandSchemas[moduleCommand];
-    const transaction = fromTransactionJSON(transactionJSON, paramsSchema);
-    const buffer = encodeTransaction(transaction, paramsSchema);
-    const transactionHex = cryptography.utils.hash(buffer).toString('hex');
-    if (isTransactionAuthor) {
-      incrementNonce(transactionHex);
+    if (transactionJSON.nonce) {
+      const paramsSchema = moduleCommandSchemas[moduleCommand];
+      const transaction = fromTransactionJSON(transactionJSON, paramsSchema);
+      const buffer = encodeTransaction(transaction, paramsSchema);
+      const transactionHex = cryptography.utils.hash(buffer).toString('hex');
+      if (isTransactionAuthor) {
+        incrementNonce(transactionHex);
+      }
     }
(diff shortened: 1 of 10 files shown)


src/modules/transaction/components/TxSignatureCollector/TxSignatureCollector.js
```diff
@@ -108,12 +108,14 @@ const TxSignatureCollector = ({
   };
 
   const onEnterPasswordSuccess = ({ privateKey }) => {
-    const paramsSchema = moduleCommandSchemas[moduleCommand];
-    const transaction = fromTransactionJSON(transactionJSON, paramsSchema);
-    const buffer = encodeTransaction(transaction, paramsSchema);
-    const transactionHex = cryptography.utils.hash(buffer).toString('hex');
-    if (isTransactionAuthor) {
-      incrementNonce(transactionHex);
+    if (transactionJSON.nonce) {
```
ManuGowda: I think this was not the bug? we need to find the root cause?

it could be the `isTransactionAuthor` which is declared in global scope? please check the root of the problem and fix

src/modules/wallet/components/RegisterMultisigForm/index.js
```diff
@@ -124,6 +126,12 @@ const Form = ({ nextStep, prevState = {}, onNext, authQuery }) => {
   const changeNumberOfSignatures = (e) => {
     const value = e.target.value ? Number(e.target.value) : undefined;
 
+    if (value > 64) {
+      setRequiredSignatureError('Cannot have more than 64 required signature');
```
ManuGowda: Use the constant MAX_MULTI_SIG_MEMBERS

ManuGowda (changes requested):
The status column value needs to be aligned 

Screenshot 2023-11-13 at 9 11 33 AM

src/modules/transaction/components/TxSignatureCollector/TxSignatureCollector.js
oskarleonard: Is this one related to signing a message? I think it is a good idea to move that logic inside the if statement because it is not used outside of the if. However, should we not still use `if(isTransactionAuthor)`

oskarleonard (approved):

ManuGowda (approved):

PR #5457, opened 2023-11-16
Stake state is not cleaned when changing account

### LINKED ISSUE #5442: Stake state is not cleaned when changing account
### Expected behavior
Redux state for stakes are not account specific, it is always store for current account, therefore  we need to clean that state when changing account, application or network. 

### Actual behavior
https://github.com/LiskHQ/lisk-desktop/assets/8784876/4b30bd1d-5bd9-472e-b9de-430d13159c0b

### Steps to reproduce
1. staking 10 lsk to validator1 from account1, 
2. Switch to account 2 (which hasnt stake anywhere yet)
3. Go to validator1
4. Expected: Not show that edit stake for that validator

oskarleonard:
### What was the problem?

This PR resolves #5442

### How was it solved?

Always clean stakes when switching account, application or network. 

### How was it tested?

1. Stake 10 LSK to validator1 from account1
2. Switch to an account which has not staked to validator1
3. Navigate to validator1
4. Expected: Not show that edit stake for that validator


src/modules/account/hooks/useCurrentAccount.js
```diff
@@ -19,51 +19,45 @@ export function useCurrentAccount() {
     (stake) => stake.confirmed !== stake.unconfirmed
   );
 
+  function switchAccount({ encryptedAccount, relativeUrlPath, redirect, urlState }) {
```
ikem-legend: Ideally this should be consistent with other function definitions. We shouldn't be having functions defined with `const` and `function` keywords in the same file

ikem-legend (dismissed):

ikem-legend (approved):

eniolam1000752 (approved):

Issue #5470, opened 2023-11-18
Need to Close "Transaction Signature Successful" Modal to Initiate New Transaction from blockchain application


RaphaelCornelis:
### Actual behavior
In the Lisk Desktop application, after completing a transaction through a dApp, a modal appears indicating "Transaction signature successful". Users are required to manually close the modal to be able to proceed with a new transaction. This behavior interrupts the user experience and could potentially confuse or hinder users from performing multiple transactions seamlessly.

### Steps to reproduce
- Complete a transaction.
- Observe the "Transaction signature successful" modal.
- Attempt to initiate a new transaction without closing the modal.

### Which version(s) does this affect? (Environment, OS, etc...)
V3.0.0-rc.4

ManuGowda:
@RaphaelCornelis could you please add the `Expected behavior`?

ManuGowda:
I guess you have expressed the expected behaviour in this issue https://github.com/LiskHQ/lisk-desktop/issues/5469, shall we consider the same?

Issue #5468, opened 2023-11-18
No of event trigger when rejecting Wallet Connect transaction


corbifex:
### Expected behavior
When a user rejects a wallet connect session request, an event should be triggered by the wallet to indicate the rejection of the session request.

### Actual behavior
No event is fired by the wallet upon rejecting a wallet connect session request. As a result, there's no notification or indication provided regarding the rejection of the session request.

### Steps to reproduce
Open the application using version 3.0.0-rc.4 that integrates wallet connect functionality.
Initiate a wallet connect session request from a wallet connect connection.
When prompted with the session request, choose `cancel` within the wallet.

### Which version(s) does this affect? (Environment, OS, etc...)
V3.0.0-rc.4

PR #5474, opened 2023-11-20
Wallet connect improvements

### LINKED ISSUE #5469: Wallet connect DAPP sign transaction is only triggered on “Copy Signature”
### Expected behavior
Let the DAPP know the transaction has successfully been signed. This should happen when password has been entered successfully not when clicking copy signature.

### Actual behavior
https://user-images.githubusercontent.com/8784876/266607400-b6912062-6aa2-4c0e-bba7-53163032c19d.mov 

(https://github.com/LiskHQ/lisk-desktop/issues/5302)

oskarleonard:
### What was the problem?

This PR resolves #5469 #5468 #5470

### How was it solved?

Changed flow of wallet connect responses

#### Notify on success
1. Initiate transaction from wallet connect
2. enter correct password
3. Expected: Wallet connect should receive response

#### Notify on cancel
1. Initiate transaction from wallet connect
2. Click cancel
3. Expected: Wallet connect should receive response

##### Allow requestView modal to receive new wallet connect requests
1. Initiate transaction from wallet connect
2. enter correct password
3. Initiate another transaction from wallet connect
4. Expected: New request should be seen in desktop wallet


src/modules/wallet/components/RequestSignStatus/index.js
```diff
@@ -29,12 +29,21 @@ const successData = (t) => ({
   ),
 });
 
+const getStringifiedTransactionJSON = (signedTransaction) => {
+  const moduleCommand = joinModuleAndCommand(signedTransaction);
+  const paramSchema = moduleCommandSchemas[moduleCommand];
+  const transactionJSON = toTransactionJSON(signedTransaction, paramSchema);
+  return JSON.stringify(transactionJSON);
+};
```
eniolam1000752: I think it would be best to get `moduleCommandSchemas` from `useCommandSchema` hook.

eniolam1000752 (changes requested):

src/modules/wallet/components/RequestSignStatus/index.js
oskarleonard: Do you think it is safe? This above code is how we did it previously. I just moved it out of the component. 
image

I am worried about loading state if we introduce that hook, could affect this component  🤔

oskarleonard: Looks like props contain the schema. Maybe we can use that instead 🤔 
image

image

eniolam1000752: hmmm okay

ManuGowda (commented):
Screenshot 2023-11-22 at 8 50 37 AM

Now that we changed from signature to full transaction we should updated the text?

ManuGowda (approved):

ikem-legend (approved):

eniolam1000752 (approved):

PR #5475, opened 2023-11-20
Fix issue with balance validation on request token deep linking

### LINKED ISSUE #5467: Deeplink does not validate minimum balance requirements
### Expected behavior

1. Check if the requested tokens can be processed from selected account
2. Check if there are other accounts which has balance to fulfil the request
3. Provide users an option to switch to the account which has balance

### Actual behavior

Screenshot 2023-11-18 at 11 10 02 AM

### Steps to reproduce

1. Create a request link for any token `lisk://wallet?modal=send&recipient=lskbgyrx3v76jxowgkgthu9yaf3dr29wqxbtxz8yp&amount=1&token=0100000100000000&recipientChain=01000001`
4. Switch to an account which has less balance than required to process the request

eniolam1000752:
### What was the problem?

This PR resolves #5467

### How was it solved?

- [x] Apply token balance checks on send token form
- [x] Remove token balance checks on overview component
- [x] Fix failing unit tests

### How was it tested?

- Switch to an account without available token balance.
- Navigate to the send modal by applying the search query params `?modal=send` to the current url
- You should be able to see the not available to token modal popup

### DIFF
--- a/src/modules/token/fungible/components/SendForm/SendForm.js
+++ b/src/modules/token/fungible/components/SendForm/SendForm.js
@@ -1,6 +1,7 @@
 /* eslint-disable max-lines */
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import Piwik from 'src/utils/piwik';
+import { useHistory } from 'react-router-dom';
 import { MODULE_COMMANDS_NAME_MAP } from '@transaction/configuration/moduleCommand';
 import AmountField from '@common/components/amountField';
 import TokenAmount from '@token/fungible/components/tokenAmount';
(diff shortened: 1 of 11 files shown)


src/modules/token/fungible/components/SendForm/SendForm.js
```diff
@@ -115,6 +129,22 @@ const SendForm = (props) => {
     return areFieldsValid && isTokenValid;
   }, [amount, recipient, reference, recipientChain, sendingChain, token]);
 
+  useEffect(() => {
+    if (!isLoadingFeeBalance && !tokenBalanceQuery.isLoading) {
```
oskarleonard, line 130: Can we move  `const hasTokenWithBalance` calculation here (lines 89-91). Inside this if statement since this is the only part of the component which uses that logic.

oskarleonard (dismissed):

ikem-legend (approved):

oskarleonard (approved):

Issue #5489, opened 2023-11-23
Incomplete special characters in request token message after using deep link


ikem-legend:
### Expected behavior
The same special characters added to a token request should be returned after opening the deep link

### Actual behavior
- Some special characters are skipped after the request token deep link is opened
[Request token deep link bug](https://www.loom.com/share/692ad8d899cc41da96b1c1b1a4900b58?sid=753029d8-bfae-4f88-8bcc-b38a8ad4836d)

- After a deep linking to the send token, the number of bytes left on the message field is always 64bytes. When an onchange event is triggered on the input field, then it starts to compute the bytes left correctly.
![image (5)](https://github.com/LiskHQ/lisk-desktop/assets/32064769/1223593f-3507-4b53-9401-0a9f21539ac8)

### Steps to reproduce
- Open the request token modal
- Click to add message
- Add message with special characters

### Which version(s) does this affect? (Environment, OS, etc...)
All

PR #5491, opened 2023-11-24
Fix crash on edit fee

### LINKED ISSUE #5487: Custom fee edit of more than 8 decimals crashes the app
### Expected behavior
Editing custom transaction fee past 8 decimals shouldn't crash the app and should show a validation error similar to the amount field

### Actual behavior
Editing custom transaction fee past 8 decimals crashes the app

### Steps to reproduce
https://github.com/LiskHQ/lisk-desktop/assets/31376203/8c478fdd-865c-49c9-b984-455d244c4215

All

eniolam1000752:
### What was the problem?

This PR resolves #5487, #5489 and #5493

### How was it solved?

- [x] Add MAX_ACCURACY to validation list on FeeViewer

### How was it tested?
- Navigate to send token form
- Initiate an edit fee action
- input Floating point value whose size is more than 8 (e.g 1.0000000033)
- it should be observed that a max accuracy error should be shown.

### DIFF
--- a/e2e/features/requestToken/RequestToken.feature
+++ b/e2e/features/requestToken/RequestToken.feature
@@ -43,8 +43,8 @@ Feature: Request Token
     Then button with text "Copy link" should be enabled
     And Element 'qrContainer' should not contain class 'disabled'
     Given I click on a button with text "Add message (Optional)"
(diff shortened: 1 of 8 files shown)


ManuGowda (approved):
LGTM

The hover issue is still not 100% resolved, so I moved the issue out of this PR.

What has done now in regards to hover is better than before.

oskarleonard (approved):

PR #5528, opened 2023-12-06
Prevent crash when a transaction is pending in a row

### LINKED ISSUE #5526: Changing self-stake crashes wallet page
Lisk Desktop 3.0.0

I changed my self-stake and it crashed the wallet page. It doesn't come back, even if I switch accounts. Only a restart of Lisk Desktop helps.

Error:
```
TypeError: Cannot read properties of undefined (reading 'timestamp'):

 in h
 in a
 in ForwardRef
 in ForwardRef
 in j
 in withI18nextTranslation(j)
 in withI18nextTranslation(j)
 in A
 in u
 in D
 in div
 in BoxContent
 in div
 in Box
 in O
 in div
 in BoxContent
 in div
 in Box
 in section
 in E
 in T
 in t
 in ErrorBoundary
 in withI18nextTranslation(ErrorBoundary)
 in main
 in j
 in t
 in div
 in j
 in withRouter(j)
 in b
 in section
 in main
 in section
 in a
 in div
 in Unknown
 in Connect(Component)
 in I
 in D
 in Q
 in withRouter(Q)
 in E
 in t
 in t
 in O
 in t
 in u
 in Unknown
```

image

oskarleonard:
### What was the problem?

This PR resolves #5526 

### How was it solved?

Check if a transaction is pending before trying to access its block. If it is pending the `block` object is undefined.

### How was it tested?
Doesnt always happen, need to make a transaction pending and keep it pending. Currently testnet has such a transaction so the above how to test works while it is still pending. 

1. Go to /transactions page.
2. Expected Not crash: Height and Date (which comes from `block` object) should display a pending symbo, check first row.
image

### DIFF
--- a/src/modules/transaction/__fixtures__/mockTransactions.js
+++ b/src/modules/transaction/__fixtures__/mockTransactions.js
@@ -33,7 +33,7 @@ const transaction = (index) => ({
     isFinal: true,
   },
   confirmations: 22 + index,
-  executionStatus: 'pending',
+  executionStatus: 'Final',
   meta: {
     recipient: {
       address: 'lsktk7bj2yadx5vq3f87gh5cwca7ptpk5djpxhhc3',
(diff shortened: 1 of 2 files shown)


eniolam1000752 (approved):
Changes LGTM

ManuGowda (approved):

PR #5535, opened 2023-12-07
Improve user experience when blockchain client or service is indexing

### LINKED ISSUE #5524: Improve user experience when blockchain client or service is indexing
### Description

Currently, we show a toaster to let users know that blockchain client is syncing or service is indexing.
However from user experience point of view this is not very good, as the basic users might not get to know why the wallet is not functioning.

### Motivation

- Provide a clear guidance or information to end users when blockchain client or service is indexing
- Since blockchain information is not up to date the users should not use the functionality until its fully index and synced up
- Users should be guided for next steps

### Additional information

Current approach
Indexin status

eniolam1000752:
### What was the problem?

This PR resolves #5524

### How was it solved?

- [x] Introduce page loader when service is indexing
- [x] Add unit tests

### How was it tested?
- Ensure you run a local instance of core and service
- Restart both instance of core and service while you ensure the app is opened and it network is switched to the local network.
- When service and core is up and running you should see the indexing loading status show up and update as the progress proceeds.


src/modules/common/components/pageLoader/PageLoader.js
```diff
@@ -15,6 +15,11 @@ function PageLoader() {
           ))}
         </div>
       </figure>
+      {progress && (
+        <div>
+          Loading <span>{progress}%</span>
```
priojeetpriyom, line 20: Not my forte but just sharing an feedback from users perspective.

It is a bit confusing as an user what the progress percentage refers to, is it the app getting it's views ready, or waiting for some meta information from server, or the server is getting ready to serve the app.  It could be more clear if we mention something which points to the fact that blockchain indexing is in progress. Ex: `Indexing Blockchain. 50% Done..`

priojeetpriyom (dismissed):
The progress detection logic seems good to me. 👍

src/modules/common/components/pageLoader/PageLoader.js
eniolam1000752: @ManuGowda @Ade-kola what's your thought on this.

src/modules/common/components/pageLoader/PageLoader.js
```diff
@@ -15,6 +15,11 @@ function PageLoader() {
           ))}
         </div>
       </figure>
+      {progress && (
```
oskarleonard: Probably need some check for 0 here as well? Get strange layout when 0 
image

e2e/scripts/run-service.sh
```diff
@@ -12,7 +12,7 @@ if [[ ! -z $LISK_SERVICE_FILE_PATH ]]
 
     echo LISK_APP_WS=ws://host.docker.internal:7887 >>${LISK_SERVICE_FILE_PATH}/.env
```
oskarleonard, line 13: NRTTL
I wasnt able to get the auto update to work. It didnt refresh the page after index update nor when indexing was complete.
@ikem-legend could you test this refresh part and see if it works for you? Might be something strange with my service setup 🤔

ikem-legend: @oskarleonard The refresh part worked for me

src/modules/common/components/pageLoader/PageLoader.js
ikem-legend: 0.01 is the minimum indexing percentage I've experienced so far and it mitigates this issue else your concern is quite valid

ikem-legend (dismissed):

ikem-legend (approved):

oskarleonard (approved):

priojeetpriyom (approved):

PR #5537, opened 2023-12-07
Remove testnet from default network

### LINKED ISSUE #5536: Remove testnet from default network
### Description

Lisk Desktop wallet is used by both basic and advanced users, to avoid confusion with many network options, its better to default to Mainnet and allow advance users to add any custom network including testnet

### Acceptance Criteria

- Make Mainnet as the default network
- Remove Testnet from default network, but ensure users can add testnet explicitly and manage them.

ManuGowda:
### What was the problem?

This PR resolves #5536

### How was it solved?

:recycle: Update default networks

### How was it tested?

- Check network drop and by default you should see Mainnet
- Add testnet and other network


eniolam1000752 (dismissed):
I refactored the fallback network to `mainnet` since we are now removing `Testnet` from the list of available networks.

ikem-legend (dismissed):

oskarleonard (dismissed):

src/modules/network/components/DialogAddNetwork/DialogAddNetwork.js
```diff
@@ -127,9 +127,7 @@ const DialogAddNetwork = () => {
             <Input
               size="l"
               label="Service URL"
-              placeholder={t(
-                'Enter service URL, e.g. https://service.lisk.com or http://localhost:9901'
-              )}
+              placeholder={t('Enter service URL, e.g. https://testnet-service.lisk.com')}
```
ikem-legend, line 130: Since testnet has been added back to the list ot networks, wouldn't that imply that the placeholders should have custom node values rather than testnet values?

oskarleonard (approved):

eniolam1000752 (approved):

src/modules/network/components/DialogAddNetwork/DialogAddNetwork.js
eniolam1000752: Since Its just an example(because of `e.g`) @ikem-legend I don't think its a problem

PR #5545, opened 2023-12-08
Fix commission history view for all validators

### LINKED ISSUE #5544: Commission change history only visible for logged in validator
### Expected behavior
The commission change history should be visible for all validators

### Actual behavior
Commission change history is only visible for logged in validators
Screenshot 2023-12-08 at 16 31 06
Screenshot 2023-12-08 at 16 31 39

### Steps to reproduce
- Open the validators page
- Select a validator to view their details

All

ikem-legend:
### What was the problem?

This PR resolves #5544

### How was it solved?

- Updating the logic for the commission history view

### How was it tested?

- Unit tests
- Manually


src/modules/pos/validator/components/ValidatorProfile/DetailsView.js
```diff
@@ -59,9 +59,9 @@ const DetailsView = ({ data, isMyProfile }) => {
       onEdit: !isMyProfile
         ? undefined
         : () => addSearchParamsToUrl(history, { modal: 'changeCommission' }),
-      onView: !isMyProfile
-        ? undefined
-        : () => addSearchParamsToUrl(history, { modal: 'commissionHistory' }),
+      onView: commission
+        ? () => addSearchParamsToUrl(history, { modal: 'commissionHistory' })
+        : undefined,
```
ManuGowda: This wont work.. all validators will have 100% commission to start with.

We should not show the icon when it doesn't have history. So best it to handle them here with 

```
onView: commission < 10000
        ? () => addSearchParamsToUrl(history, { modal: 'commissionHistory' })
        : undefined,
```

ManuGowda (dismissed):

src/modules/pos/validator/components/ValidatorProfile/DetailsView.js
ikem-legend: Fixed

src/modules/pos/validator/components/ValidatorProfile/DetailsView.js
```diff
@@ -59,9 +59,10 @@ const DetailsView = ({ data, isMyProfile }) => {
       onEdit: !isMyProfile
         ? undefined
         : () => addSearchParamsToUrl(history, { modal: 'changeCommission' }),
-      onView: !isMyProfile
-        ? undefined
-        : () => addSearchParamsToUrl(history, { modal: 'commissionHistory' }),
+      onView:
+        commission < 10000
```
eniolam1000752: would shouldn't use magic number here. We can define 10000 in a well defined constant and then re-consum here. It kind of helps readability.

eniolam1000752 (changes requested):

src/modules/pos/validator/components/ValidatorProfile/DetailsView.js
ManuGowda: @ikem-legend you can use this `MAX_VALIDATOR_COMMISSION` src/modules/pos/validator/consts/validators.js

ManuGowda (approved):

eniolam1000752 (approved):

PR #5549, opened 2023-12-13
Fix wallet connect signing on different chain

### LINKED ISSUE #5508: Nothing happens when a wallet connect transaction signing connection is invoked on a wrong current chain.
### Expected behavior
As a user if I get a wallet connect transaction signing request from a chain I am not currently on, the application should be able to guide me to switch to the right application if I manage it. If I don't manage it, possibly, it should give me the option to add the application and switch to it so I can continue with the transaction signing request.

### Actual behavior
When I get a wallet connect transaction signing request and I am on a different chain, I am unable to proceed with the signing request.

v3.0.0

ikem-legend:
### What was the problem?

This PR resolves #5508

### How was it solved?

- Implemented error when attempting to sign transaction on different chain

### How was it tested?

- Initiate a wallet connect connection
- Switch to a different application or network
- Initiate a transaction


src/locales/en/common.json
```diff
@@ -330,7 +330,7 @@
   "Invalid amount": "Invalid amount",
```
oskarleonard, line 330: NRTTL
When i switch back to correct application i still get this message

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
```diff
@@ -139,6 +142,13 @@ const RequestSummary = ({ nextStep, history, message }) => {
           validator.validator.validate(schema, transactionObj.params);
           setTransaction(transactionObj);
         }
+        if (currentApplication.chainID !== sendingChainID) {
```
oskarleonard: Feels like this is the first type of error we should check for, when we get a request. Not after doing validations etc. 

It could be moved to top execution for this components lifecycle, which is in the useEffect above this one. 

```
  useEffect(() => {
    const event = events.find((e) => e.name === EVENTS.SESSION_REQUEST);
    const hasRequest = event?.meta?.params?.request?.params;

    if (hasRequest) {
      const eventChainId = event?.meta?.params?.chainId?.replace('lisk:', ''); // Something like this?

      if (currentApplication.chainID !== eventChainId) {
        setErrorMessage(
          t('Please switch application/network to selected value during connection initialization')
        );
      } else {
        setRequest({
          id: event.meta.id,
          ...event.meta.params,
        });
      }
    }
  }, []);
```

And then make sure we dont show emptyState if we have an error message 🤔

eniolam1000752 (changes requested):
image

I think if it would be nice if we can direct users to which network/application they should switch to

ikem-legend:
> image
> I think if it would be nice if we can direct users to which network/application they should switch to

While we can know the application the user initiated the transaction with, without changing the current WalletConnect setup, we have no means of knowing the network. Some applications also have the same name in different networks, only differentiator is chainID. So in such cases, if we tried to display the application to be switched to, the user would see an error message like `Please switch application to lisk_mainchain` and that would be confusing if they're on `lisk_mainchain` but on a different network

src/locales/en/common.json
ikem-legend: You need to be sure that the WalletConnect configuration was set up with the right token details else you'd always get such errors

oskarleonard: Aha 👍

oskarleonard (dismissed):

src/modules/blockchainApplication/connection/components/RequestSummary/index.js
ikem-legend: Fixed

oskarleonard (approved):

eniolam1000752 (approved):

PR #5560, opened 2023-12-22
Change commission summary modal width

### LINKED ISSUE #5512: Commission change modal to narrow
### Expected behavior
The window should be wide enough to show whole publickey
### Actual behavior
commsion modal

### Steps to reproduce
Try changing commission with a multisig account
Lisk Desktop 3.0.0-rc.5, Windows

oskarleonard:
### What was the problem?

This PR resolves #5512

### How was it solved?

Dont use small modal on Transaction Summary step

### How was it tested?

1. Go to changeCommission modal with a multi sig account
2. Expected: Should not cut pubkey

### DIFF
--- a/src/modules/pos/validator/hooks/useCurrentCommissionPercentage.js
+++ b/src/modules/pos/validator/hooks/useCurrentCommissionPercentage.js
@@ -8,7 +8,7 @@ export const useCurrentCommissionPercentage = (address) => {
   const config = { params: { address: address ?? currentAccount?.metadata?.address } };
   const { data, ...others } = useValidators({ config });
   const currentCommission = useMemo(
-    () => convertCommissionToPercentage(data?.data[0].commission),
+    () => convertCommissionToPercentage(data?.data[0]?.commission),
     [data]
   );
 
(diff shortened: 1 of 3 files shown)


ikem-legend (approved):

src/modules/pos/validator/hooks/useCurrentCommissionPercentage.js
```diff
@@ -8,7 +8,7 @@ export const useCurrentCommissionPercentage = (address) => {
   const config = { params: { address: address ?? currentAccount?.metadata?.address } };
   const { data, ...others } = useValidators({ config });
   const currentCommission = useMemo(
-    () => convertCommissionToPercentage(data?.data[0].commission),
+    () => convertCommissionToPercentage(data?.data[0]?.commission),
```
Balanced02, line 11: is data the full query response? if so, it'll be better to have data?.data in our dependency array and not just data, as this will watch all other methods/props in the response which also includes other methods. This might not help the memo at all and it's better just having it on a variable itself...

Approving though, not a blocker, you can decide if you want to switch

Balanced02 (approved):

src/modules/pos/validator/hooks/useCurrentCommissionPercentage.js
oskarleonard: Yes, would be better to just have `commission` probably. This change i made is just to prevent a crash on page refresh, want to limit changes to only bugs at this stage.

PR #5577, opened 2024-03-25
Fix Mapbox token compromise

### LINKED ISSUE #5576: Mapbox token compromised
### Expected behavior
The map implementation on the Network page should work as expected.

### Actual behavior
The map implementation doesn't display as expected because the API token has been compromised.
Screenshot 2024-03-25 at 14 42 03

### Steps to reproduce
- Visit the Network page
- View the map in the Connected Peers section

All

ikem-legend:
### What was the problem?

This PR resolves #5576

### How was it solved?

- Updated Mapbox access token

### How was it tested?

- Visually


clemente-xyz (dismissed):

src/modules/network/components/map/index.js
```diff
@@ -52,7 +52,7 @@ const getAttributionLinks = () => {
 
 const getTiles = () =>
   L.tileLayer(
-    'https://api.mapbox.com/styles/v1/{id}/tiles/{z}/{x}/{y}?access_token=pk.eyJ1IjoibWFudWxpZ2h0Y3VydmUiLCJhIjoiY2xodWVjYW52MGM0cDNobzYzZGRkOGU0dyJ9.aOMEZq0mbD6C-DgwU_ib_A',
+    'https://api.mapbox.com/styles/v1/{id}/tiles/{z}/{x}/{y}?access_token=pk.eyJ1IjoibWFudWxpZ2h0Y3VydmUiLCJhIjoiY2x1Nm8ybTVmMG9zMDJqbnZzNWoyZzc0bSJ9.4upFBJwPjObUVkzNwChDEw',
```
shuse2: I think this needs to be at least an environment variable which is injected in the built time

shuse2 (changes requested):

src/modules/network/components/map/index.js
ikem-legend: @shuse2 Had considered that but for the most part, the environment variables used on Desktop are mostly in the webpack files or they are used but have a fallback value.

ikem-legend: Fixed

shuse2 (dismissed):

clemente-xyz:
looks a bit weird the `env` configs for gitignoring them. I think `.env.*` files should be ignored and have a `.env.example` public.

clemente-xyz (approved):

shuse2 (approved):

PR #5585, opened 2024-05-19
Fix deployment

shuse2:
### What was the problem?

Update deployment script referencing https://github.com/electron-userland/electron-builder/issues/7683#issuecomment-1724537409

### How it was tested?

Update branch to trigger the deployment and passed https://github.com/LiskHQ/lisk-desktop/actions/runs/9143940437


.github/workflows/electron-build.yml
```diff
@@ -47,17 +47,16 @@ jobs:
         run: |
           aws s3 sync s3://lisk-desktop-files/fonts ./setup/react/assets/fonts/
 
+      - name: Update package.json
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          jq '.build.mac.notarize.teamId = "${{ secrets.APPLE_TEAM_ID }}"' package.json > package_temp.json
+          mv package_temp.json package.json
```
ikem-legend, line 54: Why this temporary file move?

shuse2: It's basically overwriting the package.json from temp

ikem-legend (dismissed):

PR #5597, opened 2024-07-30
Add final mainnet balance

### LINKED ISSUE #5596: Add final balance and history download
### Description

* On the Wallet page for the mainnet final balance of the account should be shown using 
`https://legacy.lisk.com/accounts/<LEGACY_ACCOUNT_ADDRESS>.json`
* Reqest / Send / view all tokens button should be removed
* Transactions / Events should be removed
* Add `Download History` button which download the CSV for all the history for the account

### Motivation

We want to give final information for the account on L1 and access to the whole history of L1 Lisk

### Additional Information
https://docs.lisk.com/lisk-l1/legacy-chain-service

ikem-legend:
### What was the problem?

This PR resolves #5596

### How was it solved?

- Add final balance
- Remove Request / Send / view all tokens button
- Remove Transactions / Events tabs
- Add Download Account History button

### How was it tested?

- Unit test


shuse2 (changes requested):
* Token claim on the testnet is not working, (also good to check mainnet too)
```
Invalid signature request initiated from another application/network.
Please switch application/network to selected value during connection initialization
```
* For mainnet, if we connect with the address without balance, it shows error. (on mainnet or testnet). Can we show `0` balance instead?
* For testnet, it shows network error. Can we show `0` or `unknown`?
* Can we hide the history button on testnet?

shuse2:
When i changed to testnet, i had attached error.
Screenshot 2024-08-01 at 09 23 58
Also, `wallet` page is not shown anywhere, i think we should bring it back?

shuse2 (dismissed):

clemente-xyz (approved):

shuse2 (approved):

PR #5600, opened 2024-08-12
Fix private key length bug

### LINKED ISSUE #5598: Desktop Wallet not working
### Expected behavior
Desktop wallet shows latest blockchain transactions and the latest blocks on mainnet.

### Actual behavior
However, only transactions from May 21, 2024 and prior are displayed. No blocks are displayed since may 21, 2024 (no new blocks are displayed).

### Steps to reproduce
Install latest version of Desktop wallet.

I have installed 3.0.3 on mac os and 3.0.3/3.0.2 on windows both with same results above.

-----

Maybe I am missing something here, initial impressions seems like the LISK L1 Mainnet simply isn't working any more ... I am trying to transfer LISK L1 tokens from Exodus to the LISK desktop wallet so that I can then convert to L2 tokens via wallet connect.

ikem-legend:
### What was the problem?

This PR resolves #5598

### How was it solved?

- Updated private key length

### How was it tested?

- Manually


shuse2 (dismissed):

clemente-xyz (dismissed):

shuse2 (approved):

Jenkinsfile
clemente-xyz, line 1: why dont we just remove all this stuff? will be saved anyway on the git history

clemente-xyz (approved):

Jenkinsfile
ikem-legend: It's commented because there are times in the past we've had to disable e2e tests so commenting it makes it easier to reeenable.

PR #5608, opened 2024-08-15
Implement mainnet changes to fix final balance display

### LINKED ISSUE #5607: Improve final balance display
### Description

Since the final balance is for only mainnet data, we should only make the API calls on mainnet. This also ensures that an account with a mainnet balance does not show it's legacy mainnet balance while on testnet.

### Acceptance Criteria

API call for legacy balance and account history should be only made on mainnet

ikem-legend:
### What was the problem?

This PR resolves #5607

### How was it solved?

- Fixed API calls on non mainnet networks
- Fixed display of view all tokens button

### How was it tested?

- Manually


shuse2 (dismissed):
Nice catch =)

shuse2 (approved):

clemente-xyz (approved):

